PR #993, opened 2026-07-10
Migrate Codex Safe Word to packaged plugin hooks

TheMostlyGreat:
## What changed

This migrates the Codex Safe Word path toward a packaged Codex plugin plus the `safeword codex-hook` CLI entrypoint, instead of installing repo-local Codex skill and hook files into each project.

Key pieces:
- Adds the packaged Codex plugin manifest, scoped skills, and hook manifest.
- Routes every Codex hook event through `bunx --bun safeword@<version> hook codex <event>`.
- Adds deterministic BDD coverage for install isolation, hook contracts, packaged release contents, migration cleanup, trust-boundary reporting, default verification behavior, and opt-in live smoke testing.
- Documents the recommended testing strategy and plugin migration architecture.
- Adds parity coverage so packaged plugin hook matchers stay aligned with the project-local Codex config.
- Refactors the Codex plugin migration step fixtures and baselines system `shellcheck` in generated Knip config when shell tooling is present.

## Why

Codex users should get Safe Word behavior from a trusted plugin and the shipped CLI, without Safe Word owning a large set of files inside each repository. This keeps project repos cleaner while preserving install, upgrade, verification, and hook behavior.

## Bun-only decision

Bun is an intentional prerequisite for this Codex integration. The plugin uses Bunx only; it does not use or fall back to `npx`. Every hook command pins the published CLI version as `bunx --bun safeword@<version>`, and release checks reject `npx`, missing pins, and malformed commands.

This gives every hook the same reviewed runtime and package-runner behavior. Supporting npm/`npx` compatibility is separate work, not an unaddressed gap in this PR.

## Validation

- `bunx cucumber-js --profile default features/test-codex-plugin-migration.feature` — 22 scenarios, 594 steps passed.
- `bun run --cwd packages/cli test tests/schema.test.ts` — passed.
- `bun run --cwd packages/cli test tests/packs/js-app-tooling-gate.test.ts` — passed.
- `bunx knip` — clean.
- `bun run lint` — passed.
- `bun audit` — no vulnerabilities.
- Push hook schema-drift gate — 2 test files, 661 tests passed.
- `SAFEWORD_RUN_CODEX_LIVE_SMOKE=1 bunx cucumber-js --profile live features/test-codex-plugin-migration.feature` — 1 scenario, 27 steps passed.

## Notes

The live Codex smoke remains opt-in through `SAFEWORD_RUN_CODEX_LIVE_SMOKE=1`; default verification dry-runs the non-manual, non-live suite so CI does not start a live Codex session by accident.


packages/cli/src/commands/codex-hook.ts
```diff
@@ -0,0 +1,548 @@
...
+        // must inject package-owned instructions rather than project-local text.
+        SAFEWORD_PACKAGED_CONTEXT_PATH:
+          relativePath === 'session-codex-start.ts'
+            ? (findPackagedSafewordInstructionsPath() ?? '')
+            : '',
+      },
+      stdio: ['pipe', 'pipe', 'pipe'],
+    });
+
+    return result.stdout ?? '';
```
TheMostlyGreat: **Silent fail-open.** If `bun` is missing or errors, `spawnSync` sets `result.error` and leaves `result.stdout` `undefined`, so this returns `''` and the caller treats it as "hook produced no output" — the gate no-ops with no signal. Suggest checking `result.error` / non-zero `result.status` and surfacing it (stderr warning, or fail-closed for the blocking `pre-tool-use` path) rather than collapsing every failure mode into empty output. A Codex user with node but no `bun` would get a plugin that appears installed but enforces nothing.

packages/cli/src/commands/codex-hook.ts
```diff
@@ -0,0 +1,548 @@
...
+      cpSync(nodePath.dirname(hookPath), temporaryHookDirectory, { recursive: true });
+      writeFileSync(
+        nodePath.join(temporaryHookDirectory, 'lib', 'owned-paths.ts'),
+        generateOwnedPathsModule(SAFEWORD_SCHEMA, packagedNamespaceRootLabel(projectDirectory)),
+        'utf8',
+      );
+      executableHookPath = nodePath.join(temporaryHookDirectory, nodePath.basename(hookPath));
+    }
+
+    const result = spawnSync('bun', [executableHookPath], {
```
TheMostlyGreat, line 361: This is the `bun` dependency the plugin's `npx --yes safeword` entrypoint doesn't guarantee is present. Worth a single up-front `bun --version` probe (cached) so the failure is one clear message instead of N silent no-ops across every hook event.

packages/cli/codex-plugin/hooks.json
```diff
@@ -0,0 +1,69 @@
+{
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "npx --yes safeword hook codex session-start",
```
TheMostlyGreat: `npx --yes safeword` is unpinned across all five events. `--yes` suppresses the install prompt, so a session can resolve/fetch a `safeword` version that doesn't match this plugin's manifest (`0.68.0`). Consider pinning `safeword@<version>` (templated to the release version) so the plugin and CLI can't skew mid-session.

packages/cli/codex-plugin/.codex-plugin/plugin.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "safeword",
+  "version": "0.68.0",
```
TheMostlyGreat, line 3: This `version` is now a **third** source of truth alongside `packages/cli/package.json` and `.claude-plugin/marketplace.json` (which CLAUDE.md already gates with a pre-commit sync check). Nothing appears to keep this one in lockstep — it'll drift silently on the next bump. Recommend extending the version-sync hook to cover this manifest too.

TheMostlyGreat (commented):
**Note:** posting as a comment-review (GitHub blocks Request-Changes on one's own PR). Treat concern #1 below as a change I'd want before merge.

Engineering review of the Codex plugin migration. The design is clean — packaged hooks resolve from the installed CLI module dir (not project-controlled), so there's no project-side command-injection surface, and CI is green. No security holes found. I'm requesting changes on **one** concern (a gate framework that can silently disable itself) plus two supply-chain/versioning concerns worth resolving before merge.

### Concerns

**1. (blocking) Hard `bun` dependency, silently swallowed → gates no-op.** The plugin invokes the CLI over `npx --yes safeword` (node ecosystem), but the internal hook dispatch shells out to `bun` (`runPackagedHook` → `spawnSync('bun', …)`). On a machine with node/npx but **no `bun`**, `spawnSync` fails, `result.stdout` is `undefined` → the function returns `''` → every gate silently no-ops. The plugin looks installed but enforces nothing. A gate framework that quietly disables itself is worse than one that's absent. Please detect a missing/failed `bun` and surface it loudly (or document `bun` as a hard prereq for the Codex plugin and fail visibly when it's absent). See inline.

**2. `npx --yes safeword` is unpinned.** Every hook event resolves `safeword` with no version pin, and `--yes` suppresses the install prompt — a session can silently pull a CLI version that doesn't match the plugin manifest (`0.68.0`). Consider pinning `safeword@<version>` so plugin and CLI can't skew. See inline.

**3. `plugin.json` `version` is now a third version to keep in lockstep.** CLAUDE.md already mandates syncing `package.json` ↔ `marketplace.json` via a pre-commit hook; this manifest adds a third and nothing appears to gate it against the CLI version, so it will drift silently. Worth extending the version-sync check. See inline.

### Nits
- `runStop`'s fall-through around `{}` vs `''` is coherent but dense — a one-line comment on why a `{}` (noop) packaged result still checks the continuation file would help the next reader.
- Cross-PR: this diff regenerates `INDEX.md`, same as #1039/#1040 — the three will serially conflict; merge one, rebase the rest.

Nice work overall — the packaged-plugin boundary and the parity coverage are the right shape. Just don't want the gates able to fail open without a peep.

packages/cli/src/commands/codex-hook.ts
TheMostlyGreat: Resolved: the packaged hook runner now retains child status and stderr, and PreToolUse visibly denies when Bun or the child hook fails. A real subprocess regression covers both exit-code denial and missing Bun.

packages/cli/src/commands/codex-hook.ts
TheMostlyGreat: Resolved by the accepted Bun-only delivery boundary: hooks invoke the pinned Bunx CLI, and missing Bun now fails PreToolUse visibly instead of silently allowing the operation. This behavior has a subprocess regression.

packages/cli/codex-plugin/hooks.json
TheMostlyGreat: Resolved: the generated hook commands now use the exact version-pinned bunx --bun safeword@<version> form. The release contract rejects npx and unpinned commands; Bun-only was the accepted product boundary.

packages/cli/codex-plugin/.codex-plugin/plugin.json
TheMostlyGreat: Resolved: the release contract now requires the CLI package, marketplace, and Codex plugin manifest versions to match, and the release runbook names all three files.

packages/cli/codex-plugin/hooks.json
```diff
@@ -0,0 +1,69 @@
+{
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bunx --bun safeword@0.68.0 hook codex session-start",
```
TheMostlyGreat, line 9: **Version-sync gap (medium, process).** This file pins `safeword@0.68.0` in 5 command strings, and `codex-plugin-version.release.test.ts` asserts each equals `safeword@<package.json version>` (runs in CI via `test:release`). But the release procedure doesn't keep this file in sync:

- The `versioning` skill (step 2 + the step-3 `git add`) lists only `package.json`, `marketplace.json`, and `codex-plugin/.codex-plugin/plugin.json` — **not** `hooks.json`.
- The pre-commit version-sync guard (`.husky/pre-commit`) checks those same three manifests, not `hooks.json`.

So a releaser following the documented steps on the next bump leaves these 5 strings stale → the release-contract test reds the release PR. It's caught (not a silent escape), but the canonical procedure is guaranteed to hit it.

**Fix:** add `packages/cli/codex-plugin/hooks.json` to the versioning skill's bump list + `git add`, and ideally extend the pre-commit `node -e` guard to assert these 5 commands carry the current version (catch at commit, not at the release gate).

packages/cli/src/commands/codex-hook.ts
```diff
@@ -0,0 +1,585 @@
...
+    )}.`,
+  );
+  return true;
+}
+
+async function runPreToolUse(): Promise<void> {
+  const rawInput = await readStdin();
+  const projectDirectory = resolveProjectDirectory();
+  const qualityResult = runPackagedHook('codex/pre-tool-quality.ts', rawInput, projectDirectory);
+  if (emitPackagedPreToolResult(qualityResult)) return;
```
TheMostlyGreat, line 480: **Confirm output precedence (low).** If the packaged `codex/pre-tool-quality.ts` emits any stdout, this returns early and the `test-definitions.md` intake gate (`maybeDenyTestDefinitionsWrite`) below is skipped. That's correct *if* pre-tool-quality only ever emits output when it is itself denying (single JSON output per hook invocation means only one can win anyway). It's a gap only if that hook can emit non-deny `additionalContext` on PreToolUse while a test-defs write should be denied — then the deny is silently yielded. Can you confirm the Codex pre-tool-quality path never emits non-deny stdout on PreToolUse?

TheMostlyGreat (commented):
Eng + quality review. Overall this is careful, well-tested work — the destructive `.codex/config.toml` rewrite is appropriately conservative, and release-contract + packed-archive tests run in CI. No defect that ships broken. One actionable process gap plus two low-severity notes below.

packages/cli/src/codex-plugin/hooks.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/** Reject a hook command that could evade the reviewed, versioned Bunx path. */
+export function assertPinnedBunxHookCommand(command: string, version: string): void {
+  if (command.includes('--dangerously-bypass-hook-trust')) {
+    throw new Error('Safe Word plugin hooks must not bypass Codex hook trust');
+  }
+  if (/\bnpx\b/u.test(command)) {
+    throw new Error('Safe Word plugin hooks must use Bunx, never npx');
+  }
+  if (!command.startsWith('bunx --bun safeword')) {
```
TheMostlyGreat, line 23: **Nit (low).** This check is fully subsumed by the next one (`startsWith('bunx --bun safeword@${version} ')`); it exists only to yield a distinct error message. Combined with the final regex `^bunx --bun safeword@\S+ hook codex [a-z-]+$`, the three checks overlap. Fine to leave — just flagging if you're already in here. (The strict npx-reject itself is correct and worth keeping under Bun-only.)

packages/cli/codex-plugin/hooks.json
TheMostlyGreat: Resolved in d04b63b75. The pre-commit guard is now a tested `scripts/check-version-sync.ts` check that also verifies every packaged Codex hook command uses `bunx --bun safeword@<package-version>`. The release runbook and its `git add` command now include `hooks.json`; the regression test proves a stale hook pin fails.

TheMostlyGreat (commented):
Second-pass review (fresh look after the version-sync fix landed). One medium upgrade-UX issue and one low orphaned-file note below. The version-sync fix, the destructive TOML backup/replace path, and the schema `removeIfEmpty` dir safety all re-checked out as solid.

packages/cli/src/schema.ts
```diff
@@ -675,6 +560,11 @@ export const SAFEWORD_SCHEMA: SafewordSchema = {
     '.safeword/.gherkin-lintrc',
     // Merged into session-auto-upgrade.ts — check + apply now run in one pass (XQ9CXA)
     '.safeword/hooks/session-update-check.ts',
+    // Codex implementation moved into the packaged Codex plugin and `safeword hook codex`.
+    // Keep cleanup file-scoped for `.agents/skills/*` because `.agents/skills` is a shared
+    // agent directory; a user-authored sibling skill must survive migration.
+    ...CODEX_SKILL_DEPRECATED_FILES,
+    ...CODEX_RUNTIME_ASSET_FILENAMES.map(file => `.safeword/hooks/codex/${file}`),
```
TheMostlyGreat: **Upgrade leaves existing Codex users with dangling hook references (medium).** This deprecates/deletes `.safeword/hooks/codex/*.ts` (and line 610 removes the dir) on **every** `upgrade`, but `.codex/config.toml` is intentionally left untouched — and `tests/commands/upgrade-reconcile.test.ts` ("retains existing Safe Word Codex hooks until explicit migration") asserts the config stays byte-identical while its `PreToolUse`/`PostToolUse`/`Stop` hooks still run `bun "…/.safeword/hooks/codex/pre-tool-quality.ts"` etc.

**Failure scenario:** an existing Codex user runs `safeword upgrade`. Their project-local Pre/Post/Stop hooks now invoke `bun <deleted-file>` on every matching tool call and every Stop → `bun` errors on the missing path. (session-start/user-prompt-submit survive since they're still in ownedFiles, so only those three event families break.) The state persists until the user runs `safeword migrate codex-plugin --remove-legacy-hooks`.

**Aggravating:** the upgrade nudge (`utils/codex.ts` `CODEX_PLUGIN_MIGRATION_NEXT_STEP`, printed from `upgrade.ts`) points at the **bare** `safeword migrate codex-plugin`, which installs the plugin but does *not* touch the config — so following the printed instruction does not restore working hooks; the `--remove-legacy-hooks` variant is required.

**Suggested fix (pick one):** (a) don't delete `.safeword/hooks/codex/*.ts` on upgrade until `.codex/config.toml` no longer references them (gate the deprecation on migration having run), or (b) change the upgrade nudge to warn that legacy Codex project hooks are now non-functional and steer users to `safeword migrate codex-plugin --remove-legacy-hooks` (not the bare command).

packages/cli/src/schema.ts
```diff
@@ -960,7 +839,6 @@ export const SAFEWORD_SCHEMA: SafewordSchema = {
     // would keep `.cursor/` alive after reset.
     '.claude/agents/safeword-retro-filer.md': { template: 'agents/safeword-retro-filer.md' },
     '.cursor/agents/safeword-retro-filer.md': { template: 'agents/safeword-retro-filer.md' },
```
TheMostlyGreat, line 840: **Orphaned file after upgrade (low).** `.codex/agents/safeword-retro-filer.toml` was removed from ownedFiles here and its template deleted, but it's not added to `deprecatedFiles`, and `.codex`/`.codex/agents` were dropped from ownedDirs. A legacy install's copy is therefore never cleaned up — it lingers permanently (harmless, but leftover; `migrate codex-plugin` only rewrites hook blocks in `config.toml`, not this agent file). Consider adding `.codex/agents/safeword-retro-filer.toml` to `deprecatedFiles`.

packages/cli/src/schema.ts
TheMostlyGreat: Resolved in 8a400af29. Upgrade no longer removes legacy Codex runtime scripts, so an unchanged project config remains runnable. The explicit migration removes only recognized Safe Word hook configuration and deliberately preserves runtime files, avoiding broken custom references or partial filesystem handoffs. The migration guidance now makes the two-step handoff explicit.

packages/cli/src/schema.ts
TheMostlyGreat: Resolved in 8a400af29. Added `.codex/agents/safeword-retro-filer.toml` to file-scoped deprecation, with regressions for ordinary-file cleanup, user siblings, directories, dangling symlinks, and parent-path races. File deletion now uses non-recursive unlink at execution time.

packages/cli/src/commands/migrate-codex-plugin.ts
```diff
@@ -0,0 +1,482 @@
...
+): void {
+  // Validate the requested handoff before installing a profile plugin. A malformed
+  // project config must leave both the project and the Codex profile unchanged.
+  const preparedLegacyHookRemoval = options.removeLegacyHooks
+    ? prepareLegacyHookRemoval(cwd)
+    : undefined;
+
+  run('bun', ['--version']);
+  run('codex', ['--version']);
+  run('codex', [
```
TheMostlyGreat: **Two-step handoff re-runs non-idempotent `codex plugin` commands (medium).** Every `migrateCodexPlugin()` call unconditionally runs `codex plugin marketplace add` (here) and `codex plugin add` (line 448) via `run()`, which throws on any non-zero exit. The documented handoff is two separate invocations — `safeword migrate codex-plugin`, then `safeword migrate codex-plugin --remove-legacy-hooks` — so the second run re-adds an already-installed marketplace + plugin *before* reaching `removePreparedLegacyHooks`.

**Failure scenario:** if the real `codex` CLI returns a non-zero status when the marketplace or plugin already exists, the second invocation throws at the re-`add` and aborts before any cleanup — the user can never complete the handoff via the documented path. Config is left intact (no data loss), so worst case is a *stuck* handoff, not corruption.

This is an untested assumption: the BDD `"the Safe Word plugin is already installed"` step drives a mock `codex` that exits 0 on re-add, and the live smoke calls `migrateCodexPlugin` only once — so idempotent re-invocation of the real CLI is never exercised. **Fix:** short-circuit the `marketplace add` / `plugin add` steps when `plugin list --json` already reports `safeword@safeword` enabled, or tolerate an already-installed non-zero status from those two `add` calls.

TheMostlyGreat (commented):
Third-pass review (fresh look after your refactor + preserve-runtime-files changes landed). The atomic backup/replace + TOCTOU guard, the fail-closed TOML surgery, the symlink-safe non-recursive reconcile removal, and check-version-sync all re-checked out as solid. One medium handoff-robustness item inline below, plus a low nit:

**[LOW] Stale comment at `packages/cli/src/schema.ts:93`.** The `textPatches` field doc still cites "`.codex/config.toml`'s hook retrofit + MCP-server retrofit" as the ordered-patch-list example, but this PR removed the `.codex/config.toml` textPatches entry entirely — the example now points at nothing. Worth updating to a live example or dropping the parenthetical.

(Note: a refactor pass I ran independently only surfaced dedup you've since done yourself in `codex-hook.ts`, so nothing to hand over there.)

packages/cli/src/commands/migrate-codex-plugin.ts
TheMostlyGreat: Resolved in 10dd7faed. Initial migration still adds the marketplace and plugin. `--remove-legacy-hooks` now skips both mutable add commands, verifies `safeword@safeword` is enabled via `plugin list --json`, then performs the existing TOCTOU-protected atomic cleanup. The regression suite covers initial `plugin add`, cleanup without either add, and config mutation during verification. This preserves the hook review trust boundary and no longer depends on duplicate-add behavior.

nbarbettini (approved):
Nice. is `bunx safeword` still required with this plugin? or just installing the plugin in your harness?

PR #1277, opened 2026-07-21
Scan all feature lanes for E008 drift

TheMostlyGreat:
## Summary

- Route E008 surface-tag scanning through the shared executable feature-directory resolver.
- Add an internal CLI adapter so the audit shell checks root, workspace, and configured `paths.features` lanes.
- Keep all shipped audit skill copies in parity and add regression coverage.

## Root cause

E008 hardcoded the root `features/` directory, so tags in workspace feature directories were never inspected.

## Validation

- Focused Vitest: 31 passing
- `bun run typecheck`
- `bun run lint:eslint`
- Parity check: 190 pairs and 8 contracts in sync
- Independent quality review: approved


.claude/skills/audit/SKILL.md
```diff
@@ -476,18 +476,25 @@ done
 
 # --- Surface drift (E008): @surface.<slug> tag referenced but undefined ---
 # Suppressed when surfaces.md is empty/absent — W008 already says "fill it".
-FEATURES_DIR="$PROJECT_DIR/features"
+# Use the CLI's shared resolver: root, workspace, and configured feature lanes
+# must stay aligned with executable Gherkin discovery.
+if [ -f "$PROJECT_DIR/packages/cli/src/cli.ts" ]; then
```
TheMostlyGreat: **Blocker 1 — silent-clean when the CLI can't be resolved. This is the bug this PR exists to fix, through a new door.**

The guard became `[ -n "$FEATURE_DIRECTORIES" ]` (line 488). But the resolver *always* returns at least `<cwd>/features`, existent or not — so that guard's only real trigger is now **"the CLI call failed"**, and both branches here swallow stderr with `2> /dev/null`. Reproduced against a fixture with genuine drift in root `features/`:

| scenario | result |
| --- | --- |
| `bunx safeword feature-directories` exits 1 (older installed CLI lacking the command) | **silent, exit 0** |
| `bunx` not on PATH at all | **silent, exit 0** |
| `bunx` works | correctly reports E008 |

Row 1 is the realistic upgrade-skew case: any project whose skill copy is newer than its CLI loses E008 entirely, and `/audit` reports clean. That's the same false-green this PR is fixing.

Compare the house rule elsewhere — `ticket-system/SKILL.md`: *"If the resolver cannot run, stop and report that the CLI is unresolvable"*, and `verify/SKILL.md:138` prints a loud `No test-plan-capable safeword CLI found. Tried …`.

**Suggested fix:** emit a `[W…]` note on resolution failure, and/or fall back to scanning `$PROJECT_DIR/features` so it degrades to the *old* behavior rather than to nothing.

---

**Blocker 2 — this contradicts the resolver ladder 360 lines above it in this same file.**

Lines 117-121 already establish the house ladder, with an explicit rationale:

```sh
#    Resolve the locally installed safeword CLI first so the check reflects the
#    repo's pinned version, not whatever the npm registry currently calls @latest.
if [ -x node_modules/.bin/safeword ]; then SW="node_modules/.bin/safeword"
elif [ -f packages/cli/src/cli.ts ]; then SW="bun packages/cli/src/cli.ts"
else SW="bunx safeword"; fi
```

This block (a) **drops the `node_modules/.bin/safeword` rung entirely** and (b) **inverts the precedence**, putting source `cli.ts` first. Demonstrated: a fixture with a working `node_modules/.bin/safeword` and a failing `bunx` reports **silence** — the local pinned CLI is never consulted. It also makes the fallback an unpinned registry fetch, which is precisely what the comment above warns against.

**Fix is ~4 lines:** reuse the existing ladder (and reuse `$SW` if it's still in scope).

planning/user-stories/1265-scan-all-feature-lanes.md
```diff
@@ -0,0 +1,11 @@
+# Issue 1265: Scan all feature lanes for surface-tag drift
```
TheMostlyGreat: **Blocker 3 — this recreates a top-level namespace that was deliberately deleted.**

`planning/` was removed in `3a1296830` — *"chore: remove stray top-level planning/ directory (#378) … those artifacts live under `.project/tickets/`, and `planning/` is an orphan namespace referenced by nothing (not SAFEWORD_SCHEMA, parity, or .gitignore)."*

This PR reintroduces the identical shape: `planning/test-definitions/` + `planning/user-stories/` with issue-numbered filenames — same structure, same orphan-namespace problem.

**Fix:** move these under `.project/tickets/{id}-{slug}/`, or drop them if the ticket already carries the content.

TheMostlyGreat (commented):
Full independent review (filed #1265). Reviewing early since it's still draft — everything below is cheap to fix now.

**Verdict: HOLD.** The core fix is correct and provably closes the gap. But it reintroduces the same *class* of false-confidence bug through a new door, which matters more than usual here because that bug class is the entire reason this PR exists.

## What I verified passes

- **No new shell word-splitting hazard.** I extracted the block and ran it under bash 5.3, bash 3.2, and `/bin/sh` against a fixture with three lanes including one containing a space (`paths.features: "my features"`). Identical correct output in all three. The new loop — `printf '%s\n' "$FEATURE_DIRECTORIES" | while IFS= read -r features_directory` — is properly quoted and does not rely on word splitting. This was my main worry given the repo's history here, and it's clean.
- **Gap genuinely closed.** Against the real tree the new block flags **all 7** undefined slugs (`safe-word-packaged-cli`, `codex-stop`, `codex-user-prompt-submit`, `codex-plugin-hooks`, `config-wiring`, `retro-model`, `retro-pipeline`); the pre-fix block emits **0** on the same tree.
- **4-way parity holds.** `.claude/`, `.agents/`, `templates/` are byte-identical (md5 `04dc421658f1ce50de41dfd4d98ce65a`), and the codex-plugin copy is the correct *generated* transformation, not a hand-edit — re-running `generateCodexPluginAssets` reproduces **25/25** assets exactly.

## Non-blocking

- **`bun "$PROJECT_DIR/packages/cli/src/cli.ts"` can execute a foreign CLI.** That path is a common monorepo layout, not a safeword fingerprint. In a non-safeword monorepo with its own `packages/cli/src/cli.ts`, that CLI gets invoked with `feature-directories`, returns garbage, and drift in root `features/` goes unreported — a strict regression vs. pre-fix for that shape. Pre-existing pattern (line 119 does it too), but it's now load-bearing for a correctness check. A `package.json` name check, or putting the `node_modules/.bin` rung first, defuses it.
- **The E008 message and Report Format legend still say "referenced in `features/`".** The whole point of this fix is that drift lives *outside* root `features/` — the message should name the lane or file.
- **The "Coverage limitation" paragraph (SKILL.md:550) wasn't updated** to note E008 now depends on a resolvable safeword CLI.
- **Test gaps.** Both new shell tests stub `bunx` with a shim hardcoding three paths, so (a) the `bun …/cli.ts` branch — the one this repo actually takes — has no shell-level coverage; (b) `reports a surface tag from a configured feature lane` writes a `.safeword/config.json` the shim ignores, so it isn't testing config handling; (c) there's no failure-path test, which is exactly where blocker 1 lives. A test asserting E008 fails *loudly* when the resolver errors would be the natural regression guard.
- **`.agents/` is fine in this PR**, but the only test touching it asserts *substring presence*, not byte parity — a drifted bash block there would still pass. Consistent with #1240.
- The resolver enumerates `packages/*/features` for every workspace package; in JS repos using feature-folder architecture that can be a source dir. Low risk (needs a literal `@surface.<slug>`), pre-existing shape.

Happy to re-review once the blockers are addressed.

.claude/skills/audit/SKILL.md
TheMostlyGreat: Addressed in 1633907a1: the domain-docs block now follows the pinned local CLI → source checkout → bunx ladder. If all resolution paths fail, it emits W009 and falls back to root `features/`, so E008 cannot silently return clean. Added a resolver-failure regression test. Please re-review.

planning/user-stories/1265-scan-all-feature-lanes.md
TheMostlyGreat: Addressed in 1633907a1: removed both top-level `planning/` artifacts from the PR. Please re-review.

PR #1310, opened 2026-07-22
Fix native-stack discovery in audit skill

TheMostlyGreat:
## Summary
- discover Go, Python, and Rust manifests recursively outside dependency and virtual-environment trees
- run Go checks per module and Python dead-code checks per app/job scope, with explicit absent-stack and missing-tool evidence
- remove the obsolete golangci-lint output-format flag and add regression coverage for nested and ignored manifests

## Verification
- bun run test tests/skills/audit-documentation-sources.test.ts
- bun run --cwd packages/cli typecheck
- bun scripts/parity-check.ts --mode=all
- root audit: nested Go dead-code and outdated checks executed successfully

### DIFF
--- a/.agents/skills/audit/SKILL.md
+++ b/.agents/skills/audit/SKILL.md
@@ -66,16 +66,77 @@ detect_package_manager() {
   node -e 'try { const pm = JSON.parse(require("fs").readFileSync("package.json", "utf8")).packageManager || ""; console.log(pm.split("@")[0] || "npm"); } catch { console.log("npm"); }'
 }
 
-has_python_project() {
-  [ -f pyproject.toml ] || [ -f requirements.txt ] || [ -f setup.py ] || [ -f setup.cfg ] || [ -f Pipfile ]
+# Native manifests may belong to leaf applications rather than the repository
+# root. Exclude dependency and virtual-environment trees: their manifests are
+# not applications the audit owns and would make the results both noisy and slow.
+find_manifest_dirs() {
+  find . \
+    -type d \( -name .git -o -name node_modules -o -name .venv -o -name venv -o -name vendor -o -name target \) -prune -o \
+    -type f \( "$@" \) -print 2> /dev/null \
+    | while IFS= read -r manifest; do dirname "$manifest"; done \
+    | LC_ALL=C sort -u
 }
 
-has_go_project() {
-  [ -f go.mod ]
+PYTHON_PROJECT_DIRS="$(find_manifest_dirs -name pyproject.toml -o -name requirements.txt -o -name setup.py -o -name setup.cfg -o -name Pipfile)"
+GO_MODULE_DIRS="$(find_manifest_dirs -name go.mod)"
+RUST_CRATE_DIRS="$(find_manifest_dirs -name Cargo.toml)"
+
+# Run Python dead-code checks once per application, not once per toolkit/package.
+# Conventional apps/<app>/... manifests collapse to apps/<app>; jobs are one lane.
+python_audit_roots() {
+  while IFS= read -r project_dir; do
+    [ -n "$project_dir" ] || continue
+    case "$project_dir" in
+      ./apps/*/*) printf '%s\n' "$(printf '%s' "$project_dir" | cut -d/ -f1-3)" ;;
+      ./jobs/*) printf '%s\n' './jobs' ;;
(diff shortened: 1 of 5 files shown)


TheMostlyGreat (commented):
Independent review. **Verdict: HOLD** — two small fixes short of SHIP. The rewrite itself is well-built and I verified the good parts empirically, so this is close.

## Verified solid (reproduced under real `bash`, not reasoned)

- **Shell correctness / word-splitting: clean.** Extracted the discovery block and ran it under real `bash` (`env -i`, faked analyzers) against a fixture with nested Go/Python/Rust modules and a **path containing a space**. All manifests found, `cd "$dir"` handled the space, every loop iteration ran. It uses `find | while IFS= read -r` + here-docs rather than `for x in $list`, so it sidesteps the word-splitting trap this repo has been bitten by. Good craft.
- **Missing-tool / absent-stack: loud, not silent.** With tools removed from PATH it emits `Manual evidence required: golangci-lint not installed …` etc.; on an empty project it emits `No Go modules found …`. No silent-clean.
- **golangci-lint `--out-format` removal is correct** and actually fixes a latent false-confidence bug: that flag was **deleted in golangci-lint v2**, so passing it now errors, and `2>&1 || true` would have swallowed that into a false clean. No-op on v1.
- **Generated Codex copy is correct.** Re-ran `generateCodexPluginAssets` against `templates/` — the checked-in `codex-plugin/skills/audit/SKILL.md` is byte-identical to the generated output, not a hand-edit.
- **Tests are real regression tests** — they `spawnSync('bash', …)` the block against fixtures, and the nested-discovery assertions fail against main.

## Blocker 1 — the prune list misses the most common venv name, so it doesn't deliver its own headline claim

The PR says it discovers manifests "outside dependency and virtual-environment trees." The prune list (templates/skills/audit/SKILL.md:74 and :105) is:

```
-type d \( -name .git -o -name node_modules -o -name .venv -o -name vendor \) -prune -o \
```

It prunes `.venv` but **not bare `venv`** — which is *the* canonical name (`python -m venv venv`). Reproduced: on a repo using `venv/`, discovery descends into `venv/lib/.../site-packages/` and matches `pyproject.toml`/`setup.py`/`requirements.txt` in **every installed dependency**, then runs `deadcode` / `pip list --outdated` / `lint-imports` inside vendored packages. That's exactly the "noisy and slow" outcome the PR's own comment says it prevents. `target/` (Rust build dir, can hold real `Cargo.toml`) has the same issue.

Safe direction (noise, not false-clean), but a real degradation on a very common layout.

**One-line fix:** `… -o -name venv -o -name vendor -o -name target -o -name __pycache__ \) -prune -o …` — and add `venv`/`target` cases to the exclusion test (it currently asserts only `.venv`/`node_modules`/`vendor`, giving false confidence the class is covered).

## Blocker 2 — `.agents/` copy left drifted

This updated 3 of the 4 canonical audit-skill copies. sha256 at head `1ef661ef5`:

| copy | sha256 | |
| --- | --- | --- |
| `.claude/skills/audit/SKILL.md` | `61e075ca…` | updated |
| `packages/cli/templates/skills/audit/SKILL.md` | `61e075ca…` | updated (identical to .claude) |
| `packages/cli/codex-plugin/skills/audit/SKILL.md` | (generated, matches) | updated |
| **`.agents/skills/audit/SKILL.md`** | **`5297f214…` = base** | **NOT updated → drifted** |

`.agents/` still carries the old flat root-only discovery. `install.ts` writes skills into both `.claude/skills` and `.agents/skills`, and the two were byte-identical at base, so this is a genuine omission. It stays green because nothing enforces `.agents/` parity (issue #1240) and the doc-string tests assert unchanged sections.

**Impact is bounded** — end users are unaffected (install sources from `templates/`), so this only means safeword's own Codex `/audit` on this repo runs stale native-discovery logic (moot here since this repo is TS-only). But it's exactly the silent-drift channel #1240 is about. **Fix:** re-sync `.agents/skills/audit/SKILL.md` from templates.

## Non-blocking

- Prune list is also under-inclusive for `build/`, `dist/`, `.tox/`, `env/` (lower frequency than venv/target).
- `find_knip_configs` uses `sort` without `-u` (harmless — `find` can't emit dup paths).
- Exclusion test coverage gap mirrors blocker 1: no `venv`/`target`/`__pycache__` case.

Both blockers are small; the underlying rewrite is sound. Happy to re-review.

PR #1385, opened 2026-07-24
Scope stop-quality review to the current user turn

TheMostlyGreat:
## Summary

- scope Claude's stop-quality review to edit tools used in the current user turn
- keep tool-result messages within an edited-work turn and retain the bounded fallback when no user-prompt boundary is recoverable
- migrate the regression coverage to the current stop-hook transcript harness

## Root cause

The previous bounded scan considered recent assistant messages without stopping at a genuine later user prompt. An explanatory follow-up could therefore inherit an earlier edit and receive an irrelevant decision-brief continuation.

## Validation

- `bun run test tests/integration/stop-hook-transcript-format.test.ts` (13 passed)
- `bun run lint`
- `bun packages/cli/src/cli.ts sync-config --check`

Fixes #1096.

### DIFF
--- a/packages/cli/templates/hooks/stop-quality.ts
+++ b/packages/cli/templates/hooks/stop-quality.ts
@@ -363,9 +363,13 @@ checkCumulativeArtifacts(ticketInfo);
 checkImplPlanArtifact(ticketInfo);
 checkArchitectureReviewGate(ticketInfo);
 
-// No edit tools used → skip the review path (a conversational response has
-// nothing to review). The done phase is the exception: fall through to its gate.
-if (!detectEditToolsUsed(lines) && currentPhase !== 'done') {
+// No edit tools used in the current user turn → skip the review path (a
+// conversational follow-up has nothing to review). If the transcript cannot
+// recover that turn boundary, retain the prior bounded scan. The done phase is
+// the exception: fall through to its gate.
+const editsInCurrentTurn = detectEditToolsUsedInCurrentUserTurn(lines);
+const editsToReview = editsInCurrentTurn ?? detectEditToolsUsed(lines);
+if (!editsToReview && currentPhase !== 'done') {
   process.exit(0);
 }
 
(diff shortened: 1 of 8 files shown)


packages/cli/templates/hooks/stop-quality.ts
```diff
@@ -420,6 +422,39 @@ function detectEditToolsUsed(transcriptLines: string[]): boolean {
...
+ */
+function detectEditToolsUsedInCurrentUserTurn(transcriptLines: string[]): boolean | undefined {
+  let checked = 0;
+  for (let i = transcriptLines.length - 1; i >= 0 && checked < MAX_MESSAGES_FOR_TOOLS; i--) {
+    try {
+      const message: TranscriptMessage = JSON.parse(transcriptLines[i]);
+      const content = message.message?.content ?? [];
+      if (
+        message.type === 'user' &&
+        content.some(item => item.type === 'text' && item.text?.trim().length)
```
TheMostlyGreat, line 439: **Both blockers live here.** `message.type === 'user' && content.some(...)`:

- **Blocker A:** no `isMeta`/prefix guard — 137 injected `type:"user"` lines in real data (skill `Base directory`, `Stop hook feedback`, `system-reminder` notices) are mistaken for the user boundary → review silently skipped when one lands mid-turn. Skip `isMeta === true` and text starting with `<system-reminder`/`<task-notification`.
- **Blocker B:** assumes `content` is an array. ~18% of real user messages store `content` as a bare **string** (the dominant shape for short follow-ups like `'do it'`) → `content.some(...)` throws → swallowed by the outer `catch` → boundary missed → over-review. Normalize string content (or treat non-empty string as a boundary) before `.some`.

TheMostlyGreat (commented):
Independent review. **Verdict: HOLD** — the approach (scope edit-detection to the current user turn) is right, parity is clean, and it doesn't regress the hard gates. But two boundary-predicate gaps, both verified against real transcripts on disk and both untested, mean the fix doesn't fire for the dominant shape of the follow-ups it targets and adds a new silent skip. Small, localized fixes — not a redesign.

## Verified clean

- **Parity pair — byte-identical** at head: `.safeword/hooks/stop-quality.ts` and `templates/hooks/stop-quality.ts` both `d6b78f7cf014861d17609603e8f183bba6ccd2ab74f9a071d4f3509c16f7d62f`. (CI green is not parity evidence here — tests run the templates copy — so this is the real proof.)
- **Lineage / #1289 clean:** head includes #1289; the typecheck gate runs after the line-372 early-exit (same coupling as main), and #1289's git-diff-based attribution is independent — they don't fight.
- Tool-results within an edited turn are correctly kept; done-precedence (AP3FGJ) preserved.

## Blocker A — injected `type:"user"` lines are misread as genuine prompts

The boundary predicate has no way to distinguish a human prompt from harness-injected user-role text. Traced against 8 recent sessions / 5,846 assistant stop-points: **137 injected lines** would be treated as boundaries — `"Base directory for this skill: …"` (28, `isMeta:true`), `"Stop hook feedback: …"` (95, `isMeta:true`), background-task `system-reminder` notices (3, `isMeta:false`). When one lands between a turn's edits and the stop, the review is **silently skipped** — the exact false-early-cutoff risk. The Decision Record only accounts for `tool_result`, not these. `isMeta` cleanly excludes 134/137; the 3 reminders need a text-prefix check.

## Blocker B — string-form `content` breaks the boundary (misses the common case)

The predicate assumes `content` is an array. In real transcripts **~18% (536/~3000) of user messages store `content` as a bare string** — and that is the **dominant shape for short typed follow-ups** (`'do it'`, `'go check the updated pr.'`, `'do an eng and quality review of pr 993'`). For a string, `content.some(...)` throws, is swallowed by the `catch`, the genuine boundary is missed, and the scan falls through to **over-review**. Verified the boundary function returns `true` on a string-content follow-up. So the targeted bug is **not fixed for string-form follow-ups** — the exact common case. It's also new fragility: the old `detectEditToolsUsed` never parsed user content at all.

## Tests

Only Test 1 (conversational follow-up) is genuinely RED-first (diverges from main). Tests 2 & 3 pass on main too (characterization, not RED-first); **Test 3 is mislabeled** — "uses the bounded legacy scan when no boundary exists" never reaches the fallback. None of the tests cover Blocker A or B, and they model tool-results as `type:"user"` while the repo's own frozen fixture shows real ones are top-level `type:"tool_result"`.

## Nit (non-blocking)

The `?? detectEditToolsUsed(lines)` fallback is **inert** — whenever the new function returns `undefined`, the legacy scan provably returns `false` over the identical window (0 rescues across the sampled transcripts). The "retain the bounded fallback" design claim is functionally illusory; drop it or document it as defensive-only.

## Remediation (small)

In `detectEditToolsUsedInCurrentUserTurn`, before treating a `type:"user"` message as the boundary: (1) skip `isMeta === true`; (2) normalize string `content` to a text block (or treat non-empty string content as a boundary); (3) skip text starting with `<system-reminder`/`<task-notification`. Add regression tests using the real envelope for a string-content follow-up (B) and an injected `isMeta` line mid-turn (A). Then re-confirm parity sha256. Happy to re-review after.

PR #1392, opened 2026-07-24
Avoid verdict briefs on explanatory follow-ups

TheMostlyGreat:
## What changed
Scopes the Stop quality-review prompt to edit activity in the current genuine user turn. User-role tool results remain part of the same work turn, while a later text prompt ends the scan. If no reliable prompt boundary is available, the existing bounded scan remains the conservative fallback.

## Why
An earlier edit could cause the verdict-brief format to fire again on an explanatory or exploratory follow-up, adding an unnecessary round trip.

## Validation
- `./packages/cli/node_modules/.bin/vitest run packages/cli/tests/integration/stop-hook-transcript-format.test.ts` — 11 passed
- Previously completed full verification: 5,390 passed, 5 skipped.

Closes #1096.

### DIFF
--- a/packages/cli/templates/hooks/stop-quality.ts
+++ b/packages/cli/templates/hooks/stop-quality.ts
@@ -363,9 +363,13 @@ checkCumulativeArtifacts(ticketInfo);
 checkImplPlanArtifact(ticketInfo);
 checkArchitectureReviewGate(ticketInfo);
 
-// No edit tools used → skip the review path (a conversational response has
-// nothing to review). The done phase is the exception: fall through to its gate.
-if (!detectEditToolsUsed(lines) && currentPhase !== 'done') {
+// No edit tools used in the current user turn → skip the review path (a
+// conversational follow-up has nothing to review). If the transcript cannot
+// recover that turn boundary, retain the prior bounded scan. The done phase is
+// the exception: fall through to its gate.
+const editsInCurrentTurn = detectEditToolsUsedInCurrentUserTurn(lines);
+const editsToReview = editsInCurrentTurn ?? detectEditToolsUsed(lines);
+if (!editsToReview && currentPhase !== 'done') {
   process.exit(0);
 }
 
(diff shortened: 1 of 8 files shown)


packages/cli/templates/hooks/stop-quality.ts
```diff
@@ -420,6 +422,39 @@ function detectEditToolsUsed(transcriptLines: string[]): boolean {
...
+ */
+function detectEditToolsUsedInCurrentUserTurn(transcriptLines: string[]): boolean | undefined {
+  let checked = 0;
+  for (let i = transcriptLines.length - 1; i >= 0 && checked < MAX_MESSAGES_FOR_TOOLS; i--) {
+    try {
+      const message: TranscriptMessage = JSON.parse(transcriptLines[i]);
+      const content = message.message?.content ?? [];
+      if (
+        message.type === 'user' &&
+        content.some(item => item.type === 'text' && item.text?.trim().length)
```
TheMostlyGreat: This predicate is unchanged from #1385, so both blockers persist here: (A) no `isMeta`/`<system-reminder`/`<task-notification` guard — injected user-role lines are treated as the boundary → silent skip; (B) assumes `content` is an array — a bare-string `content` (96.6% of real typed prompts) makes `.some` throw → boundary missed → over-review. Fix both here before this can ship.

TheMostlyGreat (commented):
Independent review — this is the re-do of #1385 (same ticket WSFBVS, same `Closes #1096`). **Verdict: HOLD.** The decisive finding: **the runtime fix is byte-identical to #1385.** `packages/cli/templates/hooks/stop-quality.ts` at this head hashes to `d6b78f7cf014861d17609603e8f183bba6ccd2ab74f9a071d4f3509c16f7d62f` — the exact hash #1385 carried; `diff` of the whole detection region is empty. The +284/−12 vs #1385 is entirely ticket / work-log / verify artifacts plus one relocated happy-path test. **The two blockers I filed on #1385 are therefore both still present**, and real-transcript data makes them worse than first estimated.

## Parity — clean ✅
`.safeword/hooks/stop-quality.ts` == `templates/hooks/stop-quality.ts` (both `d6b78f7c…`). The one thing that's right.

## Blocker A — injected `isMeta` lines misread as prompts — NOT FIXED
Grep across the file: zero hits for `isMeta`, `system-reminder`, `task-notification`. `lines` is a raw `split('\n')` with no pre-filter. Traced with the real "Base directory" shape (`type:"user"`, `isMeta:true`, array content) landing between an edit and the stop: the predicate returns `true` on the injected line → function returns `false` → and `false ?? detectEditToolsUsed(…)` does **not** fall through (`??` only catches `null`/`undefined`) → the real edit is never reached → **review silently skipped**. Prevalence: **225 of 1176** real transcripts contain `isMeta:true`; the Base-directory array shape alone appears 718+ times.

## Blocker B — string-form `content` crashes the predicate — NOT FIXED
`content` is normalized only with `?? []`, which does not convert a **string** to an array. Traced with `content: "go check the pr"` (bare string): `"…".some` throws → swallowed by the `catch` → boundary missed → scan runs into the prior turn, finds the edit, **over-fires** — the exact behavior the ticket exists to prevent, on the most common follow-up shape. Prevalence: **96.6% (2563/2652)** of genuine typed-prompt lines store `content` as a bare string.

## Tests — no blocker coverage
Exactly one test added (`stop-hook-transcript-format.test.ts`, "skips … after a genuine user follow-up"). It uses **array** content and a **non-meta** follow-up — happy path only. It validly guards the core feature (fails on base) but touches neither blocker. Also: verify.md's wiring evidence cites `stop-quality-response.test.ts`, which **does not exist at this head**, and the ticket claims 3 integration tests where 1 is present — the "real dogfood hook" wiring claim is unsubstantiated.

## Consolidation
#1385 and #1392 are both OPEN for WSFBVS with identical fix logic. **Close one** — keep whichever you'll push the real fix to.

## To pass
1. Skip `isMeta === true` messages, and text starting with `<system-reminder`/`<task-notification`, in the boundary scan (A).
2. Normalize string content before `.some`: `const items = typeof c === 'string' ? [{type:'text', text:c}] : (c ?? [])` (B).
3. Two regression tests from the **real** envelope — an `isMeta:true` line mid-turn (review must still fire) and a bare-string follow-up after a prior-turn edit (review must skip) — each failing at the current head.

Happy to re-review once the boundary function actually changes.

PR #1465, opened 2026-07-26
fix(retro): dedup by issue enumeration, not the search index (#1453)

TheMostlyGreat:
Fixes #1453.

> **Amended post-merge (2026-07-26).** One claim below was corrected in review after merge: this description originally said the listing path was "exact, index-independent, **lag-free**." The third term overclaims — removing the search index removes *index* lag and says nothing about replication lag on the listing endpoint. The text now reads "exact and search-index-independent". Other post-merge amendments are the residual-risk clarification and follow-on pointers listed below.
>
> Direct links, so this isn't a scroll through the thread:
> - [Reviewer's correction](https://github.com/ArcadeAI/safeword/pull/1465#issuecomment-5084434675) — "lag-free", and a commit miscount (nine, not eight)
> - [Author acceptance](https://github.com/ArcadeAI/safeword/pull/1465#issuecomment-5084518174)
> - [Author retraction](https://github.com/ArcadeAI/safeword/pull/1465#issuecomment-5084583618) — two claims about the token defect I withdrew: "close to a reconstructible credential" (the test used a synthetic header) and "prerequisite for #1479" (it is a readiness gate, not an architectural prerequisite)
> - [Reviewer's dependency qualifier](https://github.com/ArcadeAI/safeword/pull/1465#issuecomment-5084563702) and [author acknowledgement](https://github.com/ArcadeAI/safeword/pull/1465#issuecomment-5084639674)
>
> Follow-on work: **#1495** (stateless `ghs_` token recognition and scrubbing — P1; split out of #834 after this review) and **#1479** (durable filing boundary; its body was rewritten to carry the amendments agreed here). #834 now covers only the original credential-absence defect.

## The reported cause does not hold

#1453 concluded that GitHub's index does not match text inside HTML comments, from "hash query → 0, prose query → issue found". Those two steps differ in **two** ways, and the second is the one doing the work: MCP `search_issues` is semantic and cannot exact-match a bare token *anywhere*.

| Query | Where that text lives | `total_count` |
| --- | --- | --- |
| `ea35770e9930` | **rendered prose** of #631 (filed 2026-07-03) | 0 |
| `searchBySignature` | **rendered prose** of #631 and #581 | 0 |
| `3c72162d3d08` | HTML comment of #581 | 0 |

A natural-language query returns #631 and #581 — the same issues whose plainly visible word `searchBySignature` returns zero. The `OR` lexical-fallback escape hatch doesn't rescue it either. So step 3 returning 0 is a property of the *query*, not of where the text sits.

What **is** confirmed: `issue_read` and `list_issues` strip HTML comments from returned bodies (checked on #1428 and freshly-filed #1454/#1455 — the rendered `<sub>` attribution survives, the marker never appears), while `search_issues` payloads return them intact. Whether GitHub's lexical `/search/issues` index tokenizes comment text remains **unverified** — this environment can't answer it (`/search/*` is proxy-blocked, `GITHUB_TOKEN` holds the `proxy-injected` placeholder, #834).

## Why the fix doesn't depend on that

The durable defect is the one #1453 flagged last: **a zero is indistinguishable from success.** An unindexed marker, an index still catching up, and a genuine absence all return the same empty array, so triage cannot tell "no duplicate" from "could not tell". That holds whichever way the indexing question resolves — so this removes the search index from dedup rather than moving the key.

`searchByExactMarker` now enumerates open issues via the repo-scoped listing endpoint and string-matches locally. Exact and search-index-independent. Raw bodies are what make it work: REST serves them verbatim — it's the MCP wrappers that sanitize. No label filter, so recall matches the old query and a marker hand-copied into an unlabeled issue during a merge (the #1425 sighting) still dedups. Truncation **throws** instead of reporting no match: triage isolates it and leaves the draft spooled, which is recoverable.

Not adopted: moving the signature into rendered text. It would only help issues filed *after* the change, leaving dedup broken against the ~340 existing open `retro` issues — and reading raw bodies makes the marker's location irrelevant.

## Defects caught in review, before merge

1. **Same-batch duplicate filing.** `prepareEncounters` does no signature dedup, so one batch can carry two findings that collide — most likely on `canonicalSignature`, which hashes *only* the repro. The cached enumeration predates the first create, so the second encounter missed and filed a duplicate. The transport now keeps a created-this-run view. Notably `FakeGitHub` searches its store live, making the test double *more correct than production* — which is why this went unnoticed; that's now called out in the double.
2. **Off-by-one at exactly 3000 items.** 30 full pages exited the loop and threw "exceed 3000" when the enumeration was complete, halting all filing on a false claim. Fixed with a probe page. The original truncation test used an always-full mock and so *enshrined* the bug; the boundary now has its own test.
3. **The guides never named the load-bearing tool.** Both files a subagent actually loads said "use a search whose payload returns raw bodies" while telling it read/list strip comments, without naming one — an agent could reasonably conclude none qualifies. `search_issues` is now named in all three.

Also: `sort=created&direction=asc` (the default `desc` lets an issue filed mid-enumeration shift later pages and skip one), a falsy `pull_request` check (a `null` must not drop a real issue from the dedup view), and a genuinely-near-miss test fixture — the previous one used an unrelated hash and passed under any substring implementation.

## Live smoke

Dedup rests on one assumption nothing in CI can reach: that the REST listing returns raw bodies with comments intact. Unit tests mock `fetch`, so they pin our parsing, never GitHub's payload. `retro-dedup.live.test.ts` closes it — self-locating fixture (newest open `retro` issue carrying a marker, so no pinned number goes stale), round-trip through `searchBySignature`, plus a negative control.

Both live smokes previously claimed to "skip loudly" via a module-scope `console.warn`; they never did — vitest's default reporter drops it, so a maintainer saw only `1 skipped`, indistinguishable from verified. Vitest's purpose-built routes don't close it either (annotations print only on failure; a skipped test cannot annotate itself). A failing test is the one output every reporter always prints, so the precondition is asserted as a test. `SAFEWORD_LIVE_ALLOW_SKIP=1` acknowledges and restores the skip.

## Verification

- Full suite **5423 passed**; the 4 failures are pre-existing and reproduce on clean `HEAD` (root-user `self-report` — that's #1454 — plus the go/python/rust lint-fallback tests).
- Acceptance lane **494 scenarios passed** (a `.feature` file changed here, and `bun run lint` skips that lane — #1455).
- Both new regression tests confirmed **non-vacuous** by reverting each fix and watching them fail.
- Live-lane behavior verified three ways: no token → loud failure under the *default* reporter; acknowledged → green skip; normal lane still excludes `*.live.test.ts`.
- `lint`, `typecheck`, `format:check`, `parity:fix`, `knip`, `depcruise` (651 modules, 0 violations) all clean.

## Residual risk

The listing endpoint's comment-preserving behavior is verified from GitHub's REST docs and from `search_issues` returning markers, but not observed directly here — that's exactly what the live smoke exists to catch on first maintainer run.

Ambiguous cross-run creates are **not** covered: GitHub exposes no documented conditional-create/idempotency or strong-consistency contract for issue creation, so a create whose response is lost cannot be resolved deterministically from the client. `createdThisRun` closes same-transport visibility only. That is #1479's territory.

https://claude.ai/code/session_01Nudd5uKCsRpcQchQUggEny

### DIFF
--- a/.claude/agents/safeword-retro-filer.md
+++ b/.claude/agents/safeword-retro-filer.md
@@ -19,12 +19,18 @@ procedure, your target repo, or your tools.
    the file is missing or holds no drafts, report `retro-filer: nothing to file`
    and stop.
 2. **Dedup each draft against open issues on `ArcadeAI/safeword`** (and only
-   there). Every search uses `is:issue is:open`, then exact-checks the raw body:
-   first search the `<!-- safeword-retro-signature: ... -->` marker for the
-   draft's signature. Only if that misses, and `canonicalSignature` is present,
-   confirm the draft body contains its exact
-   `<!-- safeword-retro-canonical: <canonicalSignature> -->` marker, then search
-   and exact-check that canonical marker. A missing or mismatched body marker
+   there). First consult the sibling `.acks.jsonl`: a signature acked there is
+   already filed — comment on the recorded issue, never create.
+   **Never search for a marker or its hash.** The markers sit in HTML comments,
+   which issue read/list tools strip from the bodies they return and which no
+   available search matches as query text (#1453) — a zero from such a query
+   means "could not tell", not "not filed", and never authorizes a create.
+   Instead search by **topic** with `is:issue is:open`, using a search whose
+   payload returns raw bodies, then exact-check those bodies: first the draft's
+   `<!-- safeword-retro-signature: ... -->` marker. Only if that misses, and
+   `canonicalSignature` is present, confirm the draft body contains its exact
+   `<!-- safeword-retro-canonical: <canonicalSignature> -->` marker, then
+   exact-check that canonical marker. A missing or mismatched body marker
    disables canonical fallback. Never use a title as duplicate authority.
    - **Match** → add a one-line comment that the finding recurred, ending with
      the draft's `<!-- safeword-retro-signature: ... -->` marker on its own
--- a/.claude/agents/safeword-retro-filer.md
+++ b/.claude/agents/safeword-retro-filer.md
@@ -25,18 +25,9 @@ procedure, your target repo, or your tools.
    which issue read/list tools strip from the bodies they return and which no
    available search matches as query text (#1453) — a zero from such a query
    means "could not tell", not "not filed", and never authorizes a create.
-   Work the two reads for what each is actually good for:
-   - **`list_issues` (labels: `retro`, state: `OPEN`) is your candidate
-     universe.** Page through it to the end — it is exhaustive and
-     deterministic. Its bodies come back with HTML comments stripped, so it
-     cannot confirm a marker, but it is the only read that can tell you nothing
-     matches.
-   - **`search_issues` is your marker check.** It is the one read path whose
-     payload returns raw bodies with markers intact — but it is
-     relevance-ranked and capped, so it is **never exhaustive**. Use it to
-     confirm a candidate, never to prove absence.
-
-   Exact-check the raw bodies it returns: first the draft's
+   Use `search_issues` for the check: it is the one read path whose payload
+   returns raw bodies with markers intact. Query it by **topic**, then
+   exact-check the bodies it returns: first the draft's
    `<!-- safeword-retro-signature: ... -->` marker. Only if that misses, and
    `canonicalSignature` is present, confirm the draft body contains its exact
    `<!-- safeword-retro-canonical: <canonicalSignature> -->` marker, then
@@ -45,19 +36,27 @@ procedure, your target repo, or your tools.
    - **Marker confirmed** → add a one-line comment that the finding recurred,
      ending with the draft's `<!-- safeword-retro-signature: ... -->` marker on
      its own line. Do not open a duplicate.
-   - **No candidate anywhere in the exhaustive `list_issues` enumeration** →
-     create a new issue with the draft's `title`, `body`, and `labels`
-     **verbatim**. Never add, remove, or rephrase content — the draft is the
-     sanitized surface, and the signature marker inside the body is what future
-     dedup depends on.
-   - **A plausible candidate you could not confirm** — same
-     `**Safeword surface:**` line or substantially the same title, but no marker
-     you could verify — → comment on it rather than create. Those rendered
-     fields are weak identity (they drift between sessions, #631), so they may
-     never merge two findings on their own; they are only ever enough to make
-     you withhold a create. **An empty `search_issues` result is not absence**
-     — creating on one is the same "could not tell means not filed" mistake
-     #1453 was about, one layer up.
+   - **No marker confirmed** → create a new issue with the draft's `title`,
+     `body`, and `labels` **verbatim**. Never add, remove, or rephrase content —
+     the draft is the sanitized surface, and the signature marker inside the
+     body is what future dedup depends on.
+
+   **This path is best-effort by construction, and you should know why.** No
+   read available to you can prove a finding is absent: `search_issues` returns
+   raw bodies but is relevance-ranked and capped, and the reads that enumerate
+   exhaustively (`list_issues`, `issue_read`) strip HTML comments, so they can
+   never see a marker. A duplicate is therefore possible. File anyway — a
+   duplicate is recoverable (the reconcile sweep closes confirmed ones), while a
+   finding you decline to file is simply lost, because this path runs precisely
+   when the code-owned REST path could not authenticate (#834) and no later
+   retry exists.
+
+   What that licence does **not** cover: never merge two findings on a
+   resemblance. A matching `**Safeword surface:**` or a similar title is weak
+   identity that drifts between sessions (#631). Commenting-and-acking on that
+   basis binds the signature to that issue permanently and throws the draft body
+   away — a silent, lossy merge, and the worse error of the two. Only a
+   confirmed marker may join a draft to an existing issue.
 
 3. **Ack each post before you drain it.** After each successful post (create or
    comment), append exactly one COMPACT single-line JSON object `{"signature": "<signature>", "issue": <number>}`
--- a/.claude/agents/safeword-retro-filer.md
+++ b/.claude/agents/safeword-retro-filer.md
@@ -25,21 +25,40 @@ procedure, your target repo, or your tools.
    which issue read/list tools strip from the bodies they return and which no
    available search matches as query text (#1453) — a zero from such a query
    means "could not tell", not "not filed", and never authorizes a create.
-   Instead search by **topic** with `is:issue is:open` using GitHub MCP
-   `search_issues` — its payload returns raw bodies with the markers intact, and
-   it is the one read path that does — then exact-check those bodies: first the draft's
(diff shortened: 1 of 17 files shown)


TheMostlyGreat (commented):

**Verdict: REQUEST CHANGES**

### Findings

1. **P1 — Agent filing still creates from an incomplete topic search.**

   The shipped filer instructions say to search by an unspecified “topic,” exact-check only the returned candidates, and create on “no match.” That result set is not exhaustive: GitHub search is relevance-ranked, capped, and can report incomplete results. The MCP tool also exposes a bounded `topn`, not an exhaustive raw-body enumeration. Thus an older issue can be absent because its title/category/surface drifted—the exact case `canonicalSignature` is intended to cover—and the agent will still create a duplicate.

   Affected canonical source: `packages/cli/templates/skills/retro-filer/SKILL.md:17-34`; mirrored agent/skill/guide copies carry the same behavior.

   **Required fix:** when only topic search is available, do not authorize a create from an empty candidate set. Either provide an exhaustive raw-body listing primitive or fail closed and retain the draft. Add a real search→create behavior test where topic search omits an existing canonical match.

   Source: https://docs.github.com/en/rest/search/search

2. **P2 — Page-number pagination can skip an issue when an earlier issue closes mid-enumeration.**

   `packages/cli/src/retro/github-rest.ts:136-153` sorts ascending, which prevents newly-created issues from shifting already-read pages. It does not handle removals: if an issue closes after page 1, later items shift backward and one item can fall across the page boundary unseen. If that skipped issue carries the requested marker, dedup returns “no match” and files the silent duplicate this PR is intended to prevent.

   **Required fix:** use a stable snapshot/cursor strategy, or repeat the enumeration and fail closed unless the observed item set is stable. Add a mutation-between-pages regression test.

3. **P3 — The stated 3,000-item cap actually accepts 3,001–3,099 items.**

   `packages/cli/src/retro/github-rest.ts:132-160` appends and accepts a short probe page. The probe correctly distinguishes exactly 3,000 from a real tail, but any 1–99 items on page 31 are returned instead of tripping the advertised cap.

   **Suggested fix:** treat any non-empty probe page as “exceeds 3,000,” without appending it, and add a 3,001-item test.

4. **P3 — Deterministic truncation is retried for every encounter.**

   `packages/cli/src/retro/github-rest.ts:164-173` clears the cached promise for every failure, including terminal truncation. Each later encounter can therefore repeat all 31 requests before failing again.

   **Suggested fix:** retry transient HTTP failures, but cache terminal truncation for the transport lifetime.

### Quality review

- **Currency:** ✓ Current GitHub documentation confirms raw Markdown bodies, issue/PR filtering, sorting, and pagination behavior.
- **Sources:** ✓ Load-bearing API claims were checked against current official documentation.
- **Correct:** ❌ The agent path and pagination mutation race can still file duplicates.
- **Elegant:** ⚠ The code-owned REST approach is thoughtful; the agent fallback remains structurally ambiguous.
- **No-bloat:** ✓ Most added complexity is justified by fail-closed behavior.
- **Wiring:** ⚠ `retro-dedup.live.test.ts` is a strong real-API wiring test for the REST path, but the changed agent path has only static prompt/phrasing tests.

### Engineering review

- **Security:** ✓ No new security vulnerability found.
- **Performance:** ⚠ Terminal truncation can amplify into repeated 31-request enumerations.
- **Correctness:** ❌ Two duplicate-filing paths remain.
- **Maintainability:** ⚠ The probe behavior contradicts the documented cap.
- **Positive observations:** exact local marker matching, promise sharing, `createdThisRun`, fuzzy near-miss coverage, synchronized templates, and the live raw-body round trip are all strong.

### Validation performed

- All four GitHub CI checks passed.
- 53 targeted unit tests passed.
- 9 shipped-prompt contract tests passed.
- Both live GitHub smoke files passed against the real API (3 + 2 tests).
- `git diff --check` passed.

**Next:** fix P1 and P2, add their regression coverage, then re-review.

TheMostlyGreat (commented):
Follow-up analysis on P1:

### Recommended near-term design: bounded GitHub-native index

Use 16 pre-created labels, `retro-key-0` through `retro-key-f`, keyed by the first hex character of each dedup identity. A retro issue receives the bucket for its canonical key and, during migration, the bucket for its legacy signature (at most two labels, using the same 16-label set).

The filer then:

1. searches open issues in the relevant bucket;
2. receives their raw bodies;
3. performs the existing full hidden-marker equality check;
4. creates only when there is no exact match.

The labels are only a candidate index; the complete marker remains the source of identity.

Current repository measurements make this comfortably bounded: there are 334 open retro issues, while the largest current first-nibble bucket contains 30 issues. A `topn:100` label-qualified lookup therefore has substantial headroom and avoids treating relevance-ranked global search as exhaustive. I also verified that the installed GitHub connector honors the label qualifier while returning raw bodies containing the hidden markers.

Required safeguards:

- pre-create the 16 labels;
- backfill existing canonical and legacy markers;
- verify the expected bucket labels after issue creation;
- repair labels after issue-body edits or copied markers;
- fail closed and alert if a bucket reaches the connector's result cap.

This resolves the specific P1 false-absence problem without introducing another service. It does **not** provide atomic uniqueness for simultaneous filers, nor safe recovery when GitHub creates an issue but the response is lost.

### Long-term correctness boundary

I filed #1479 for a dedicated idempotent filing operation backed by durable key claims. That design is the correctness ceiling: repeated and concurrent requests converge on one issue, ambiguous creates can be recovered, and temporary outages retain the draft. It is also materially more expensive because it introduces a service/store, authentication, migration, and operations.

Recommendation: use the bounded label index as the practical P1 follow-up; treat #1479 as the long-term architecture. The current at-least-once behavior should remain an explicit policy choice rather than being described as complete deduplication.

TheMostlyGreat:
**author** — both record corrections accepted; my closing summary above is wrong on both points.

**Nine commits, not eight.** Verified: `389bffa`, `5a7d1a7`, `6dad56e`, `a58f066`, `44ca711`, `87f970f`, `7ca092b`, `de73574`, `2b32858`. I dropped `87f970f` — which is the commit where I reverted my own incorrect fix to the stability check. Miscounting by omitting my worst commit is worth stating rather than quietly amending.

**"Lag-free" overclaims.** The PR body says the listing path is "exact, index-independent, lag-free." The first two hold; the third doesn't. Removing the search index removes *index* lag and says nothing about replication lag on the listing endpoint — a distinction I raised myself hours later on #1479 and should not have blurred here. `createdThisRun` closes same-transport visibility only; ambiguous cross-run creates remain #1479 territory, as you say.

On the adjacent `ghs_` finding: I reproduced both halves against the real functions and it's worse than "incompletely scrubbed" — the payload **and** signature survive verbatim, only the header segment is redacted, and JWT headers are low-entropy enough that the remainder is close to a reconstructible credential. It's also a **prerequisite for #1479** rather than merely adjacent, since that ticket's server-held GitHub App credential *is* a stateless `ghs_` token. Detail and a test-assertion caution are on #834.

Agreed the merge conclusion is unaffected: this remains the honest interim, the bucket index stays deferred, #1479 remains the destination.

---

PR #1471, opened 2026-07-26
Keep zombie cleanup inside the current project

### LINKED ISSUE #1451: Cleanup zombies targets unrelated projects that share the detected port
**Category:** bug
**Safeword surface:** `packages/cli/templates/scripts/cleanup-zombies.sh`

**What happened**

The cleanup preview selected a server owned by a different project because it was listening on Safeword's auto-detected port.

**Why it is friction**

Confirming the documented cleanup command would terminate an unrelated project's process, contradicting the script's multi-project safety guarantee.

**Repro**

Run [path] in a Safeword project while another project owns the auto-detected dev port, then inspect the preview candidate.

<sub>Auto-filed by `safeword retro` from a session transcript; sanitized at egress (no customer data).</sub>

TheMostlyGreat:
## What changed

- Apply one fail-closed working-directory ownership rule to both port and command-pattern candidates.
- Parse `lsof` field output safely with NUL delimiters, including paths containing newlines.
- Revalidate and signal each PID individually to narrow process-identity races.
- PATH-resolve `kill` through an explicit test seam and count only successful signals.
- Explain when foreign port owners were skipped instead of reporting the project as “already clean.”
- Fail loudly when required tools or `pgrep` discovery fail, while preserving summaries for earlier completed work.
- Validate explicit ports as decimal values in the 1–65535 range and derive the conventional test port only when it remains valid.
- Pass patterns after `pgrep`'s option terminator and avoid duplicate built-in/custom pattern sweeps.
- Add deterministic integration coverage for unrelated projects, descendants, unknown ownership, tricky pathnames, pattern candidates, failed signals, multi-PID behavior, port boundaries, and leading-dash patterns.
- Update the cleanup guide, README, website reference, generated Codex skill, and dogfood mirrors; raw machine-wide kill examples are no longer presented as project-safe.

## Why

Port and pattern discovery are machine-wide. Previously, a foreign process could be treated as project-owned when its argv mentioned this project, and filtered foreign port owners produced a misleading “already clean” result. Cleanup now uses the process working directory as the single ownership boundary and leaves anything unverified alone.

## Impact

Developers can run zombie cleanup without terminating another project that shares a port or whose command line references the current project. If a foreign process still occupies a detected port, the command explains that recovery condition. Failed signals are no longer counted as successful kills. Broad foreign pattern matches remain silently ineligible to avoid machine-wide skip noise.

## Decision evidence

A `$safeword:figure-it-out` pass compared a narrow polish, a unified ownership rule, and a split follow-up. The unified rule won because destructive cleanup should not change its definition of “this project” based on how it discovered the PID. The main tradeoff is that a legitimate child that changes working directory will be skipped; fail-closed behavior is intentional.

A fresh `$safeword:quality-review` and full `$safeword:refactor` pass resolved every valid finding, including port arithmetic, process-label fallback, duplicate pattern discovery, test-harness argument fidelity, generated plugin parity, and stale verification evidence. Risky Bash generalizations and framework-detection restructuring were deliberately deferred because they add complexity or fall outside issue #1451.

## Verification

- 369 test files passed: 5,513 tests passed; 5 skipped (5,518 total).
- 494/494 active Cucumber scenarios and 15,313/15,313 active steps passed; 3 scenarios and 4 steps skipped.
- Focused cleanup suite: 51/51 passed.
- Build, ESLint, Gherkin lint, TypeScript, Bash syntax, ShellCheck, formatting, config sync, mirror checks, generated Codex catalogue, and targeted Markdown checks passed.
- Changed production script, integration test, and guide duplication scan: 0 clones.
- Audit passed with warnings only for pre-existing/out-of-scope repository conditions: hidden-worktree scan noise, unavailable Python audit tools, and two dev-only patch updates.

Fixes #1451

### DIFF
--- a/.claude/skills/cleanup-zombies/SKILL.md
+++ b/.claude/skills/cleanup-zombies/SKILL.md
@@ -28,9 +28,9 @@ If the preview looks correct, confirm the kill with `--yes`:
 ## What It Does
 
 1. **Auto-detects framework** - Finds port from vite.config.ts, next.config.js, etc. (checks root, `packages/*/`, `apps/*/` for monorepos)
-2. **Kills by port** - Dev server port AND test port (port + 1000)
-3. **Kills test processes** - Playwright, Chromium, Electron (scoped to this project)
-4. **Multi-project safe** - Only kills processes matching this project's directory
+2. **Checks project-owned port processes** - Inspects the dev port and in-range test port (port + 1000), reporting owners it skips when project ownership cannot be verified
+3. **Checks project-owned test processes** - Inspects Playwright, Chromium, and Electron matches whose working directory is inside this project
+4. **Revalidates before signaling** - Only kills processes whose current working directory still belongs to this project
 
 ## Manual Override
 
--- a/.cursor/commands/cleanup-zombies.md
+++ b/.cursor/commands/cleanup-zombies.md
@@ -24,9 +24,9 @@ If the preview looks correct, confirm the kill with `--yes`:
 ## What It Does
 
 1. **Auto-detects framework** - Finds port from vite.config.ts, next.config.js, etc. (checks root, `packages/*/`, `apps/*/` for monorepos)
-2. **Kills by port** - Dev server port AND test port (port + 1000)
-3. **Kills test processes** - Playwright, Chromium, Electron (scoped to this project)
-4. **Multi-project safe** - Only kills processes matching this project's directory
+2. **Checks project-owned port processes** - Inspects the dev port and in-range test port (port + 1000), reporting owners it skips when project ownership cannot be verified
+3. **Checks project-owned test processes** - Inspects Playwright, Chromium, and Electron matches whose working directory is inside this project
+4. **Revalidates before signaling** - Only kills processes whose current working directory still belongs to this project
 
 ## Manual Override
 
--- a/.project/tickets/1451-cleanup-zombies-project-safety/ticket.md
+++ b/.project/tickets/1451-cleanup-zombies-project-safety/ticket.md
@@ -1,0 +1,44 @@
+---
+id: "1451"
+slug: cleanup-zombies-project-safety
+type: task
+phase: done
+status: done
+created: 2026-07-25
+last_modified: 2026-07-26
+external_issue: https://github.com/ArcadeAI/safeword/issues/1451
+---
+
+# Task: Keep zombie cleanup inside the current project
+
+**Type:** Bug
+
+**Scope:** Restrict processes found through auto-detected dev and test ports to
+the current project before previewing or killing them.
+
+**Out of Scope:** Changing framework detection, port conventions, pattern-based
+cleanup, or the explicit consent flow.
+
+**Done When:**
+
+- [x] A process from another project is omitted from the cleanup preview when it
+      owns this project's auto-detected port.
+- [x] A process owned by the current project remains eligible for cleanup by its
+      auto-detected port.
+- [x] Port ownership checks deny by default when project ownership cannot be
+      established.
+
+**Tests:**
+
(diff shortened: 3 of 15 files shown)


TheMostlyGreat (commented):
## Reviewer: APPROVE

Verified live rather than by reading. Two real listeners — one in a project dir on the dev port, one in an unrelated dir on the test port. `main` lists both (`Found 2 process(es) that would be killed`, including the foreign PID); this branch lists only the project-owned one. The bug in #1451 is genuinely closed on the port path.

The `lsof -Fn0` parse is also correct against the real binary, not just the mock: `od -c` of live `lsof -a -p PID -d cwd -Fn0` (lsof 4.91, darwin) gives exactly `p<pid>\0 \n f c w d \0 n<path>\0 \n`, and the test stub emits that byte sequence verbatim. The set-terminating NL only ever prepends to the `f` field, so scanning for the first `n*` chunk is sound — and the NUL delimiting is what makes the newline-in-cwd case safe. `man lsof` documents the contract (§OUTPUT: with the NUL field terminator, each set still ends with a NL). Script parses and runs under `/bin/bash` 3.2.57 (`local arr=()`, `+=`, process substitution all fine). Template and `.safeword/` mirror are byte-identical.

No critical issues. Four suggestions, none blocking:

### 1. The pattern path still has the exact hole the port path just closed

`cleanup_pattern` matches on argv (`cleanup-zombies.sh:232`), so ownership there still means "the command line mentions this project," not "the process lives in this project."

Repro'd on this branch: a process with cwd in an unrelated project and argv containing `playwright <this-project>/report` is listed as a kill candidate when the script runs from this project. That is precisely the scenario the new test _"an unrelated command argument referencing this project is excluded"_ asserts against on the port path — so within a single run the script now applies two different definitions of ownership.

Pre-existing, and #1451 is port-specific, so this isn't a reason to hold the PR. But the guide's "Scopes all pattern matching to current project directory" reads stronger than what that line delivers. Worth a follow-up issue to run the same `process_belongs_to_project` filter over `pgrep` results.

### 2. `printf … | xargs -n 1 kill -9` deserves a comment saying why

`cleanup-zombies.sh:218`. For a single known PID this looks like an obvious simplification target — but collapsing it to `kill -9 "$pid"` would hit the bash **builtin** and bypass the `PATH`-injected `kill` mock, silently blinding the kill-log assertions in the new tests. One line ("xargs so the PATH-resolved `kill` is used — test seam") protects it from a future cleanup pass.

Related nit: `KILLED_COUNT` increments even when the signal fails (errors swallowed), so "Killed N" can overcount a process that exited between revalidation and signal. Cosmetic.

### 3. "already clean" overstates the outcome when candidates were filtered out

`cleanup-zombies.sh:282`. When a foreign process owns the port, the user's actual symptom — port in use — survives, and the script reports _"No zombie processes found - already clean!"_. The deny-by-default decision is invisible at exactly the moment it matters.

Something like `skipped N process(es) on port X owned by other projects` would explain why the port is still busy. Current wording is pinned by the new tests, so this is a deliberate-choice question rather than a defect — flagging it because this PR is what creates the divergence between "clean" and "nothing we're willing to touch."

### 4. Two small ones in the same neighborhood as #1

- `$project_dir` is interpolated into `pgrep`'s regex unescaped (`:232`). `.` over-matches harmlessly, but a `+` or `(` in a project path makes `pgrep` error into `|| true` — the cleanup silently does nothing.
- The `/private/var` → `/var` alias (`:136`) doesn't cover `/private/tmp`, also a macOS temp root. This doesn't affect the cwd path (both `pwd -P` and lsof report resolved paths — confirmed live), only the argv path, so low value — noting it because the comment above it reads as a general macOS-temp claim.

Issue #904, opened 2026-07-07
`git commit --amend` invalidates a just-recorded phase_anchors SHA


TheMostlyGreat:
**Category:** gap
**Safeword surface:** `packages/cli/templates/hooks/lib/phase-provenance.ts`

**What happened**

Commit, anchor a phase advance to that SHA, then amend the commit (e.g. via lint-staged or a human edit) — the anchor SHA is gone, and because amending changes the patch, the ledger's patch-id-based rebase recovery also fails to find it.

**Why it is friction**

An entirely honest workflow (commit → amend, a routine git action) produces a false 'anchor unreachable' verdict later, with no recovery path built in.

**Repro**

Advance a ticket's phase, anchor it to the current commit SHA, run `git commit --amend`, then attempt to resolve that anchor via the ledger's git-based ShaResolver — resolution fails.

<sub>Auto-filed by `safeword retro` from a session transcript; sanitized at egress (no customer data).</sub>

PR #1477, opened 2026-07-26
Prevent phase anchors from accepting foreign artifacts

TheMostlyGreat:
## Summary

- bind ticket-scoped phase anchors to the owning ticket instead of accepting any same-kind artifact
- bind feature-source anchors to the ticket slug and an executable/default/configured feature lane
- require ownership scope on both public detectors and remove the redundant fail-open scoped wrappers
- read project and feature roots from the same staged/HEAD tree being validated, including repository-root configurations
- reject Git index-stage aliases, traversal, duplicate separators, and other noncanonical anchor paths
- add all 37 ledger behaviors to executable Gherkin coverage, including the ten ownership/configuration cases previously covered only below the acceptance lane
- consolidate the boundary scope model, shared workspace-root constant, health config reads, and test fixtures

## Why

PR #964 made issue #904 amend-safe by replacing commit-SHA anchors with artifact paths. The remaining implementation accepted artifact kind and shape without fully binding the path to its ticket/configuration authority, allowing foreign-artifact substitution and staged/worktree configuration mismatches. This follow-up hardens that substrate without changing the warn-and-record boundary policy.

Follow-up to #904 and #964; it does not re-close the already closed issue.

## Validation

- CI on the remediation head: Dogfood parity, lint, Node 22, and Node 24 all pass
- focused Vitest: 4 files, 76 tests passed
- Cucumber: 505 passed, 3 skipped; 15,645 executed steps passed
- issue-focused Cucumber: 40 scenarios and 1,210 steps passed
- build, ESLint, Gherkin lint, `tsc --noEmit`, diff hygiene, and dogfood parity: clean
- audit: 0 change-scoped errors; dependency-cruiser 0 violations across 662 modules / 2,166 dependencies; Knip clean; dependencies current; clone count improved from 490 to 455 at the same scope
- independent quality review and post-refactor review: PASS

## Local environment note

- Full local Vitest: 5,492 passed, 5 skipped, with one failure in untouched upstream Astro preset files: `astro/no-omitted-end-tags` is `"error"` under the locally installed eslint-plugin-astro 3.0.1 while `astro.test.ts` expects it to be absent. The preset, test, manifests, and lockfile are unchanged from `origin/main`, and both Node 22 and Node 24 CI pass on this remediation head, so this does not reproduce in the clean branch environment.
- The push boundary emits warn-only ledger reachability advisories for historical HGYGND RED/GREEN/REFACTOR SHAs that were squash-merged before this follow-up. New hardening evidence references resolve on this branch.

### DIFF
--- a/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
+++ b/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
@@ -139,7 +139,7 @@ partitions after real CLI probes found two anchor-substitution gaps.
 
 ### Scenario: A ticket cannot reuse another ticket's feature source
 
-- [ ] RED
+- [x] RED b612e91
 - [ ] GREEN
 - [ ] REFACTOR
 
@@ -151,7 +151,7 @@ partitions after real CLI probes found two anchor-substitution gaps.
 
 ### Scenario: OS-native ticket paths are normalized to the anchor grammar
 
-- [ ] RED
+- [x] RED b612e91
 - [ ] GREEN
 - [ ] REFACTOR
 
--- a/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
+++ b/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
@@ -71,9 +71,9 @@ partitions after real CLI probes found two anchor-substitution gaps.
 
 ### Scenario: A rebased commit does not disturb a recorded anchor
 
-- [x] RED 4b546cd
-- [x] GREEN ca68343
-- [x] REFACTOR f78c62e
+- [x] RED 2cd0ef5
+- [x] GREEN 356c4ee
+- [x] REFACTOR a611f69
 
 ### Scenario: A shallow clone's anchor check passes with no unreachable-history hedging
 
@@ -133,63 +133,63 @@ partitions after real CLI probes found two anchor-substitution gaps.
 
 ### Scenario: A ticket cannot reuse another ticket's same-kind artifact
 
-- [x] RED 4b546cd
-- [x] GREEN ca68343
-- [x] REFACTOR 6751bcf
+- [x] RED 2cd0ef5
+- [x] GREEN 356c4ee
+- [x] REFACTOR a0e95de
 
 ### Scenario: A ticket cannot reuse another ticket's feature source
 
-- [x] RED b612e91
-- [x] GREEN 5a3b22d
-- [x] REFACTOR 93bf3fe
+- [x] RED b5ed264
+- [x] GREEN bdcb3cb
+- [x] REFACTOR da20049
 
 ### Scenario: A Git index-stage prefix cannot alias a different artifact path
 
-- [x] RED 4b546cd
-- [x] GREEN ca68343
-- [x] REFACTOR 6751bcf
+- [x] RED 2cd0ef5
+- [x] GREEN 356c4ee
+- [x] REFACTOR a0e95de
 
 ### Scenario: OS-native ticket paths are normalized to the anchor grammar
 
-- [x] RED b612e91
-- [x] GREEN 5a3b22d
-- [x] REFACTOR 93bf3fe
+- [x] RED b5ed264
+- [x] GREEN bdcb3cb
+- [x] REFACTOR da20049
 
 ### Scenario: Canonical feature ownership is independent of unstaged worktree state
 
-- [x] RED 713d352
-- [x] GREEN 511ea43
-- [x] REFACTOR 93bf3fe
+- [x] RED ae77c9d
+- [x] GREEN 6a179bc
+- [x] REFACTOR da20049
 
 ### Scenario: Feature anchors must live in an executable or configured feature lane
 
-- [x] RED 8c8c4ca
-- [x] GREEN 8eee17d
-- [x] REFACTOR 93bf3fe
+- [x] RED 4f1475d
+- [x] GREEN eb7ef1e
+- [x] REFACTOR da20049
 
 ### Scenario: Ticket discovery uses the staged project-root configuration
 
-- [x] RED caa410f
-- [x] GREEN 430a493
-- [x] REFACTOR 93bf3fe
+- [x] RED dfec6ce
+- [x] GREEN 27ba49d
+- [x] REFACTOR da20049
 
 ### Scenario: Configured project roots normalize dot and separator segments
 
-- [x] RED 32ead6b
-- [x] GREEN 492b4cd
-- [x] REFACTOR 93bf3fe
+- [x] RED ccc5783
+- [x] GREEN 4dd7121
+- [x] REFACTOR da20049
 
 ### Scenario: Repository-root ticket and feature lanes remain enforceable
 
-- [x] RED 23278c9
-- [x] GREEN 1186488
-- [x] REFACTOR 93bf3fe
+- [x] RED 1f9fce8
+- [x] GREEN fcb6d78
+- [x] REFACTOR da20049
 
 ### Scenario: A configured project root outside the repository warns instead of failing open
 
-- [x] RED 23278c9
-- [x] GREEN 1186488
-- [x] REFACTOR 93bf3fe
+- [x] RED 1f9fce8
+- [x] GREEN fcb6d78
+- [x] REFACTOR da20049
 
 ### Scenario: A backward phase move is not flagged
 
--- a/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
+++ b/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
@@ -181,13 +181,13 @@ partitions after real CLI probes found two anchor-substitution gaps.
 
 ### Scenario: Repository-root ticket and feature lanes remain enforceable
 
-- [ ] RED
+- [x] RED 23278c9
 - [ ] GREEN
 - [ ] REFACTOR
 
 ### Scenario: A configured project root outside the repository warns instead of failing open
 
-- [ ] RED
+- [x] RED 23278c9
 - [ ] GREEN
 - [ ] REFACTOR
 
--- a/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
+++ b/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
@@ -169,7 +169,7 @@ partitions after real CLI probes found two anchor-substitution gaps.
 
 ### Scenario: Ticket discovery uses the staged project-root configuration
 
-- [ ] RED
+- [x] RED caa410f
 - [ ] GREEN
 - [ ] REFACTOR
 
--- a/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
+++ b/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
@@ -131,6 +131,44 @@ partitions after real CLI probes found two anchor-substitution gaps.
 - [x] GREEN 73b9759
 - [x] REFACTOR 5bd56ea
 
+### Scenario: A backward phase move is not flagged
+
+- [x] RED e90eb07
+- [x] GREEN 73b9759
+- [x] REFACTOR 5bd56ea
+
+### Scenario: Re-declaring the same phase is not flagged
+
+- [x] RED e90eb07
+- [x] GREEN 73b9759
+- [x] REFACTOR 5bd56ea
+
+### Scenario: A non-feature ticket advancing is not flagged
+
+- [x] RED e90eb07
+- [x] GREEN 73b9759
+- [x] REFACTOR 5bd56ea
+
+### Scenario: A ticket becoming a feature past intake is not flagged
+
+- [x] RED e90eb07
+- [x] GREEN 73b9759
+- [x] REFACTOR 5bd56ea
+
+### Scenario: An edit that leaves the phase unchanged is not flagged
+
+- [x] RED e90eb07
+- [x] GREEN 73b9759
+- [x] REFACTOR 5bd56ea
+
+### Scenario: The at-rest advisory nudges a missing anchor with the path grammar
+
+- [x] RED e90eb07
+- [x] GREEN 73b9759
+- [x] REFACTOR 5bd56ea
+
+## Rule: Ownership and configured lanes are resolved from canonical staged paths
+
 ### Scenario: A ticket cannot reuse another ticket's same-kind artifact
 
 - [x] RED 2cd0ef5
@@ -190,42 +228,6 @@ partitions after real CLI probes found two anchor-substitution gaps.
 - [x] RED 1f9fce8
 - [x] GREEN fcb6d78
 - [x] REFACTOR da20049
-
-### Scenario: A backward phase move is not flagged
-
-- [x] RED e90eb07
-- [x] GREEN 73b9759
-- [x] REFACTOR 5bd56ea
-
-### Scenario: Re-declaring the same phase is not flagged
-
-- [x] RED e90eb07
-- [x] GREEN 73b9759
-- [x] REFACTOR 5bd56ea
-
-### Scenario: A non-feature ticket advancing is not flagged
-
-- [x] RED e90eb07
-- [x] GREEN 73b9759
-- [x] REFACTOR 5bd56ea
-
-### Scenario: A ticket becoming a feature past intake is not flagged
-
-- [x] RED e90eb07
-- [x] GREEN 73b9759
-- [x] REFACTOR 5bd56ea
-
-### Scenario: An edit that leaves the phase unchanged is not flagged
-
-- [x] RED e90eb07
-- [x] GREEN 73b9759
-- [x] REFACTOR 5bd56ea
-
-### Scenario: The at-rest advisory nudges a missing anchor with the path grammar
-
-- [x] RED e90eb07
-- [x] GREEN 73b9759
-- [x] REFACTOR 5bd56ea
 
 ## Rule: Legacy SHA anchors neither warn at rest nor block new work
 
--- a/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
+++ b/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
@@ -167,6 +167,12 @@ partitions after real CLI probes found two anchor-substitution gaps.
 - [ ] GREEN
 - [ ] REFACTOR
 
+### Scenario: Ticket discovery uses the staged project-root configuration
+
+- [ ] RED
+- [ ] GREEN
+- [ ] REFACTOR
+
 ### Scenario: A backward phase move is not flagged
 
 - [x] RED e90eb07
--- a/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
+++ b/.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
@@ -12,10 +12,10 @@ green.
 
 Deliberate coverage notes (from the independent scenario review, applied
 2026-07-08): per-kind shape partitions for spec.md / feature source / ledger /
-verify.md are unit-lane only (dimensions.md test-layers note); the post-rebase
-history partition is subsumed by the squash scenario (a strictly harsher
-rewrite); done-gate ledger isolation is carried by the existing regression
-suites.
+verify.md are unit-lane only (dimensions.md test-layers note); done-gate ledger
+isolation is carried by the existing regression suites. The 2026-07-25 quality
+review added explicit rebase coverage and adversarial ticket/path binding
+partitions after real CLI probes found two anchor-substitution gaps.
 
 ## Rule: A forward advance anchors the entered phase to the exited phase's artifact
 
@@ -68,6 +68,12 @@ suites.
 - [x] RED 73b9759
 - [x] GREEN ea8bcf6
 - [x] REFACTOR 5bd56ea
+
+### Scenario: A rebased commit does not disturb a recorded anchor
+
+- [ ] RED
+- [ ] GREEN
+- [ ] REFACTOR
 
 ### Scenario: A shallow clone's anchor check passes with no unreachable-history hedging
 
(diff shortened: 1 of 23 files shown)


TheMostlyGreat (commented):
Reviewer: Quality + engineering review.

Verification done this session: targeted suite 145/145 green locally (6 files); CI green on all four checks; dogfood/template `phase-provenance.ts` byte-identical; the `:0:path` index-stage hole reproduced in a scratch repo (`git show :0:a.md` -> content, exit 0), so that guard fixes a real bypass; and I mutation-tested the new ownership branch (disabled `isOwnedArtifact`, rebuilt, reran) to confirm the new boundary tests actually kill it.

What's sound: reading config from the validated tree rather than disk is the right call for a check whose premise is "verify the tree being shipped"; keying `byTicket` on `ticketPath` correctly disambiguates identical folder names under both namespace roots; and the new `boundary.test.ts` cases are genuine wiring coverage (real CLI subprocess, real git, real staging — `unlinkSync` after `git add` to prove staged-vs-worktree).

Two things I'd fix before merge, both design/spec hygiene rather than defects in what ships (warn-only boundary = no user-facing breakage risk either way):

1. The ownership check is fail-open by default, and the acceptance lane exercises the fail-open path — see the inline comment on `phase-provenance.ts`.
2. `test-definitions.md` declares 10 ledger scenarios with no Gherkin counterpart, and they are exactly this PR's new behaviors — see the inline comment on `test-definitions.md`.

The rest are non-blocking notes, inline.

packages/cli/templates/hooks/lib/phase-provenance.ts
```diff
@@ -531,6 +558,31 @@ function validateAnchor(
...
+  const anchorSegments = anchor.split('/');
+  const isConfiguredOrDefaultFeature =
+    scope?.featureRoots.some(root => root === '' || anchor.startsWith(`${root}/`)) ?? false;
+  const isWorkspaceFeature =
+    anchorSegments.length >= 4 &&
+    (scope?.workspaceRoots.includes(anchorSegments[0] ?? '') ?? false) &&
+    anchorSegments[1] !== '' &&
+    anchorSegments[2] === 'features';
+  const isOwnedArtifact =
+    scope === undefined ||
```
TheMostlyGreat: Reviewer (blocking): this leaves the ownership check **fail-open by default**, and the acceptance lane exercises the fail-open path.

`scope` stays optional on `detectUnanchoredPhaseTransition` (:475) and `detectUnanchoredPhaseState` (:613), and when it's omitted `scope === undefined ||` short-circuits `isOwnedArtifact` to `true`. Both unscoped functions remain exported from the **shipped template lib**, so any new caller reaching for the obvious-sounding name silently gets no ownership binding. The `detectScoped*` pair are pure argument-reorder shims whose only job is making one parameter required — 4 wrapper functions across the 2 mirrored copies.

Worse: `steps/artifact-content-phase-anchors.steps.ts:115` imports the **unscoped** functions, so all 27 Gherkin scenarios for this feature validate the permissive variant rather than what production runs.

Suggested fix: make `scope` required on the two originals, delete the four wrappers, and update the call sites (~40 in `tests/hooks/phase-anchor.test.ts` plus the steps file). Smaller surface, no fail-open, and the acceptance lane tests the real thing.

.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
```diff
@@ -125,6 +131,66 @@ suites.
 - [x] GREEN 73b9759
 - [x] REFACTOR 5bd56ea
 
+### Scenario: A ticket cannot reuse another ticket's same-kind artifact
```
TheMostlyGreat, line 172: Reviewer (blocking): ledger/feature drift — 10 of these 37 scenario headings have no scenario in `features/artifact-content-phase-anchors.feature` (which has 27), and they are precisely this PR's new behaviors:

```
A ticket cannot reuse another ticket's same-kind artifact
A ticket cannot reuse another ticket's feature source
A Git index-stage prefix cannot alias a different artifact path
OS-native ticket paths are normalized to the anchor grammar
Canonical feature ownership is independent of unstaged worktree state
Feature anchors must live in an executable or configured feature lane
Ticket discovery uses the staged project-root configuration
Configured project roots normalize dot and separator segments
Repository-root ticket and feature lanes remain enforceable
A configured project root outside the repository warns instead of failing open
```

`verify.md` records "All 37 scenarios marked complete" against a 27-scenario feature. Because HGYGND is `phase: done`, the ticket-scoped uncovered/orphan coverage gate never fires — so the drift is invisible to the gate that exists to catch it.

Either add the Gherkin scenarios (preferred — they're the acceptance contract for the PR's headline behavior, and this PR already showed the bar is reachable by adding "A rebased commit does not disturb a recorded anchor"), or extend the coverage-notes paragraph above to declare these partitions unit-lane-only, the way the existing per-kind shape partitions are declared.

packages/cli/tests/commands/boundary.test.ts
```diff
@@ -91,6 +92,319 @@ describe('safeword boundary (slice 1: engine core)', () => {
...
+          phase: 'implement',
+          anchors: [`implement: ${foreignTicket}/impl-plan.md`],
+        }),
+      );
+      git(dir, 'add -A');
+
+      const result = await runCli(['boundary', '--at', 'commit'], { cwd: dir });
+
+      expect(result.exitCode).toBe(0);
+      expect(`${result.stdout}\n${result.stderr}`).toMatch(/ticket.*artifact|artifact.*ticket/i);
```
TheMostlyGreat: Reviewer (note): loose assertion. I mutation-tested this (disabled the `isOwnedArtifact` branch, rebuilt, reran) and it **does** fail, so it isn't vacuous today — but it discriminates by luck: the competing `[ledger] ... R/G/R ledger is missing` warning that fires in the mutant simply happens not to contain the word "ticket". Assert on the distinctive substring `outside this ticket`, or on the `[phase-anchor]` check tag the way the negative tests in this PR already do.

packages/cli/tests/commands/boundary.test.ts
```diff
@@ -91,6 +92,319 @@ describe('safeword boundary (slice 1: engine core)', () => {
...
+          phase: 'scenario-gate',
+          anchors: [`scenario-gate: ${foreignFeature}`],
+        }),
+      );
+      git(dir, 'add -A');
+
+      const result = await runCli(['boundary', '--at', 'commit'], { cwd: dir });
+
+      expect(result.exitCode).toBe(0);
+      expect(`${result.stdout}\n${result.stderr}`).toMatch(/ticket/i);
```
TheMostlyGreat: Reviewer (note): same as :122 — `toMatch(/ticket/i)` is the weakest form of this assertion. It survives mutation only because the competing ledger warning doesn't say "ticket". Prefer `outside this ticket` or `[phase-anchor]`.

steps/boundary-reconciliation-gate.steps.ts
```diff
@@ -464,6 +464,9 @@ const ticket = (phase, anchors) => ['---','id: BGT','type: feature','phase: '+ph
 const anchor = 'implement: .project/tickets/BGT001-fixture/impl-plan.md';
 const verdicts = reconcileChange([{
   ticketFolder: 'BGT001-fixture',
+  ticketPath: '.project/tickets/BGT001-fixture',
+  featureRoots: ['features'],
+  workspaceRoots: ['packages', 'apps', 'libs', 'modules'],
```
TheMostlyGreat: Reviewer (note): this hand-copies `WORKSPACE_ROOTS`. `packages/cli/src/utils/workspaces.ts:19` documents that constant as shared specifically "so the scanners that walk them — workspace-member, BDD feature-source, and cucumber-harness detection — cannot drift (a root added in one place but not another silently breaks detection or discovery)." A literal copy in a steps fixture is exactly that drift, and it's silent when a root is added.

The fixture body is an evaluated string so it can't `import` directly — but the array can be interpolated from the real constant when the snippet is built.

packages/cli/src/boundary/engine.ts
```diff
@@ -50,6 +51,12 @@ export interface LegalityStep {
 export interface TicketChange {
   /** Ticket folder name, e.g. `BND001-clean`. */
   ticketFolder: string;
+  /** Repo-relative path of this ticket's folder, for ticket-local anchors. */
+  ticketPath: string;
```
TheMostlyGreat: Reviewer (note): these three fields are exactly `PhaseAnchorScope`'s three fields, and `reconcileTicket` (:316) reassembles them into a scope object on every call. Carrying `anchorScope: PhaseAnchorScope` instead would drop the reassembly and shrink the fixture copy in `steps/boundary-reconciliation-gate.steps.ts`.

Related: `ticketFolder` is now derivable as `basenameOf(ticketPath)` — duplicated state that can disagree.

packages/cli/src/utils/configured-paths.ts
```diff
@@ -239,6 +252,30 @@ export function resolveTicketsDirectory(cwd: string): string {
...
+  if (configuredProjectRoot === undefined) {
+    return [
+      nodePath.posix.join(NAMESPACE_ROOT_DEFAULT, 'tickets'),
+      nodePath.posix.join(NAMESPACE_ROOT_LEGACY, 'tickets'),
+    ];
+  }
+  const projectRoot = toRepoDirectory(cwd, configuredProjectRoot);
+  if (projectRoot === undefined) {
+    throw new Error(
+      `Configured project root "${configuredProjectRoot}" is outside the repository.`,
```
TheMostlyGreat: Reviewer (note): throw-as-control-flow in a module whose header documents the opposite contract ("Empty-string or non-string values are treated as unset (defensive)") — every other reader here returns `undefined` for a bad value.

Consequence: `boundary()`'s catch converts this into the generic `boundary check could not complete: ...` and disables **every** check for that repo — ledger, legality, artifact shape — not just anchors. Returning `[]` and warning at the call site keeps the others alive.

No practical loss today (an out-of-repo project root has no in-repo tickets to check), so this is style rather than a bug — but the blast radius is wider than the failure warrants.

packages/cli/src/health.ts
```diff
@@ -351,7 +356,11 @@ function phaseAnchorAdvisoryForTicket(
 ): string | undefined {
   const content = readFileSafe(nodePath.join(ticketsRoot, ticketId, 'ticket.md'));
   if (content === undefined || !isInProgress(content)) return undefined;
-  const verdict = detectUnanchoredPhaseState(content, relpath =>
+  const ticketDirectory = nodePath.join(ticketsRoot, ticketId);
+  const ticketPath = toRepoRelativePath(cwd, ticketDirectory);
+  const configuredFeatures = readConfiguredPath(cwd, 'features');
```
TheMostlyGreat: Reviewer (note): this sits inside `phaseAnchorAdvisoryForTicket`, so `.safeword/config.json` is re-read and re-parsed once per in-progress ticket. Hoist the `readConfiguredPath` / `createPhaseAnchorScope`-invariant part to the caller that iterates tickets.

.project/tickets/HGYGND-artifact-content-phase-anchors/verify.md
```diff
@@ -27,3 +27,33 @@ fixture; cli-test and steps lanes each share theirs); outdated: knip 6.24.0→6.
...
 both updated in this ticket).
+
+## Issue #904 hardening verification — 2026-07-25
+
+Evidence from the repository-generated test plan after the complete audit,
+independent quality review, and refactor ledger.
+
+## Verify Checklist
+
+**Test Suite:** ✓ 5492/5492 tests pass (366 files; 5 skipped)
```
TheMostlyGreat, line 39: Reviewer (note): this records 5492/5492 passing, while the PR description lists a post-rebase `astro.test.ts` failure under "Baseline limitations after rebasing onto current `main`". CI (node 22.22.3 and node 24, both green) doesn't reproduce that failure.

Worth reconciling: either this section is a pre-rebase run and should say so, or the local astro failure was stale-`dist` noise and the PR description should drop it. As written the two documents disagree about whether the branch is green.

packages/cli/templates/hooks/lib/phase-provenance.ts
TheMostlyGreat: Reviewer: adding the mechanics I left out, so this is actionable without re-deriving them.

**Both copies move together.** `packages/cli/src/schema.ts:696` binds `.safeword/hooks/lib/phase-provenance.ts` -> `templates/hooks/lib/phase-provenance.ts`. They're byte-identical today (verified with `diff -q`); editing only the template turns the *Dogfood parity* CI job red.

**Exact call sites needing a `scope` argument** (counted on this branch):

| Location | Calls |
| --- | --- |
| `packages/cli/tests/hooks/phase-anchor.test.ts` -> `detectUnanchoredPhaseTransition(` | 27 |
| `packages/cli/tests/hooks/phase-anchor.test.ts` -> `detectUnanchoredPhaseState(` | 10 |
| `steps/artifact-content-phase-anchors.steps.ts` (:115 import, :122, :123) | 3 |

The 37 unit call sites can share one module-level fixture scope — every existing case anchors inside `TICKET_DIR` or `features/<slug>.feature`, so a single `{ ticketPath: <fixture ticket dir>, featureRoots: ['features'], workspaceRoots: [...WORKSPACE_ROOTS] }` satisfies all of them.

**The steps file needs the same treatment as the `WORKSPACE_ROOTS` note.** That snippet is an evaluated string, so it can't `import` the scope type or constant — the literal has to be interpolated in when the snippet is built, same fix shape as `steps/boundary-reconciliation-gate.steps.ts:469`. Worth doing both in one pass so the two fixtures don't drift apart from `workspaces.ts:19` independently.

With the parameter required, `detectScopedUnanchoredPhaseTransition` / `detectScopedUnanchoredPhaseState` delete cleanly in both copies — their only callers are `engine.ts:162` and `health.ts:363`.

.project/tickets/HGYGND-artifact-content-phase-anchors/test-definitions.md
TheMostlyGreat: Reviewer: two mechanics I left out, whichever option you take.

**Where the scenarios go, if you add them.** Seven of the ten are rejection cases and belong under the existing `Rule: A forward advance without a real artifact behind it is detectable as unanchored` (feature file :94), tagged `@artifact-content-phase-anchors.SM1.R3` to match its siblings. The three configuration ones — staged project-root discovery, dot/separator normalization, repository-root lanes — aren't about a missing artifact and read as a new Rule ("anchor ownership is resolved from the tree's own configuration") with its own R-tag, since spec.md already grew the matching language in this PR.

**The count line needs updating either way.** `verify.md:42` reads `**Scenarios:** All 37 scenarios marked complete (112/112 R/G/R cells)` — that 37 is the ledger heading count, not the feature's 27. Adding the scenarios makes it true; declaring them unit-lane-only means the line should say how many are executable versus unit-lane, the way the coverage-notes paragraph already distinguishes the per-kind shape partitions.

packages/cli/src/health.ts
TheMostlyGreat: Reviewer: naming the caller I was vague about — `findCoverageDiagnostics` (`health.ts:331`). Its `for (const ticketId of listTicketIds(ticketsRoot))` loop at :334 calls this once per ticket, so `readConfiguredPath(cwd, 'features')` re-reads and re-parses `.safeword/config.json` once per ticket in the repo. Read it once above the loop and pass the value down.

features/artifact-content-phase-anchors.feature
```diff
@@ -83,6 +91,68 @@ Feature: Artifact-content phase anchors — a phase advance is evidenced by the
       When the boundary command runs at the commit boundary
       Then it exits zero and warns that the anchored artifact is missing from the staged tree
 
+  Rule: Ownership and configured lanes are resolved from canonical staged paths
+
+    @artifact-content-phase-anchors.SM1.R2
```
TheMostlyGreat: Reviewer (note): this Rule reuses R2's tag, and `spec.md` has no rule for it.

These ten scenarios sit under `Rule: Ownership and configured lanes are resolved from canonical staged paths` (:94) but carry `@artifact-content-phase-anchors.SM1.R2`, whose `spec.md:241` definition is *"An anchored advance verifies from the tree alone, under any history"* — a different claim. `spec.md` defines exactly R1–R5; the feature now has six Rule blocks mapped onto five R-tags.

Nothing catches this. `findFeatureLineageIssues` (`packages/cli/src/utils/gherkin-feature.ts:100-116`) enforces only *exactly one lineage tag per scenario* — it never checks that the tag resolves to a rule in `spec.md`, and never checks that a `Rule:` heading agrees with the tag beneath it. Same blind-spot class as the ledger/feature drift from the first pass, one level up: the behavior is now executable, but its lineage points at someone else's rule.

Either add `SM1.R6` to `spec.md` with this Rule's wording and retag the ten scenarios, or fold them into the existing R2 block and drop the new heading. I'd lean R6 — the ownership claim genuinely isn't the tree-only claim.

Your call on which; I'm not blocking on it.

packages/cli/src/utils/workspaces.ts
```diff
@@ -9,14 +9,14 @@ import nodePath from 'node:path';
 
 import { exists, readJson } from './fs.js';
 
+export { WORKSPACE_ROOTS } from './workspace-roots.js';
```
TheMostlyGreat, line 12: Reviewer (note): the docstring got orphaned by the move.

`WORKSPACE_ROOTS` now lives in `workspace-roots.ts`, but its doc comment stayed here at :14-19 and sits above `interface PackageJson` — so it currently documents the wrong declaration.

The bigger loss is at the destination. `workspace-roots.ts:1-6` reads *"Kept dependency-free so acceptance fixtures can import the production constant"* — true, but that's the **extraction** rationale. The **sharing** rationale is the one that just got deleted:

> Shared so the scanners that walk them — workspace-member, BDD feature-source, and cucumber-harness detection — cannot drift (a root added in one place but not another silently breaks detection or discovery).

That sentence is exactly why the hand-copied literal was worth fixing in the first place. Move it to `workspace-roots.ts` alongside the new paragraph and delete the stranded copy here.

packages/cli/src/utils/configured-paths.ts
```diff
@@ -239,6 +252,28 @@ export function resolveTicketsDirectory(cwd: string): string {
...
 
+/**
+ * Repo-relative ticket directories for a caller-supplied project-root config.
+ * With no explicit root, preserve namespace precedence: `.project` wins over
+ * legacy `.safeword-project`, and a fresh project defaults to `.project`.
+ */
+export function ticketDirectoriesForConfiguredRoot(
+  cwd: string,
+  configuredProjectRoot?: string,
+  rootExists: (repoRelativeRoot: string) => boolean = root => isDirectory(nodePath.join(cwd, root)),
```
TheMostlyGreat: Reviewer (note): this default is dead, and it defaults to the wrong semantics.

`rootExists` falls back to `isDirectory(...)` — **worktree** state — but the sole caller (`boundary.ts:227`) always passes the tree probe, and no test exercises the 2-arg form. So the fallback is unreachable today, and if someone later calls this with two arguments they'd silently get worktree semantics from a function whose whole contract is *resolve from the tree being validated*. That's precisely the staged-vs-worktree confusion this PR set out to close.

Make the parameter required. The one caller already supplies it, so it costs nothing and removes a trap.

.project/tickets/HGYGND-artifact-content-phase-anchors/verify.md
```diff
@@ -27,3 +27,86 @@ fixture; cli-test and steps lanes each share theirs); outdated: knip 6.24.0→6.
...
+  and repository-path suites.
+- Full Cucumber: 505/508 scenarios pass, 3 skip; 15,645/15,645 executed steps
+  pass. The issue-focused lane is 40/40 scenarios and 1,210/1,210 steps.
+- Build succeeds; ESLint, Gherkin lint, `tsc --noEmit`, diff hygiene, and
+  template/dogfood parity are clean.
+- Full local Vitest: 5,492 pass, 5 skip, with one failure in the untouched
+  upstream Astro preset expectation (`astro/no-omitted-end-tags` is `"error"`
+  under the installed eslint-plugin-astro 3.0.1 while the test expects it to be
+  absent). The Astro preset, test, manifests, and lockfile are unchanged from
+  `origin/main`; Node 22 and Node 24 CI passed on remediation head `a36e0d027`.
```
TheMostlyGreat: Reviewer (note): stale SHA — seven commits landed after `a36e0d027`. CI **is** green on the current head (all four checks), so this is just a refresh, not a re-run.

The rest of this section resolves my earlier note cleanly: relabelling the original block "pre-rebase baseline" and recording the Astro discrepancy with the unchanged-from-`origin/main` evidence is exactly the reconciliation I was asking for.

TheMostlyGreat (commented):
Reviewer: re-review of the remediation head `f31af0748`. **All nine threads verified resolved — approving.** Four documentation/lineage notes below, none blocking.

### What I re-verified rather than took on faith

| Check | Result |
| --- | --- |
| Targeted suite (6 files, from `packages/cli`) | **150/150 pass** (was 145) |
| Mutation test — disable `isOwnedArtifact`, rebuild, rerun | **kills 3 tests** (was 2) — the strengthened `/outside this ticket/i` assertions caught one more failure mode |
| `diff -q` on the two `phase-provenance.ts` copies | byte-identical |
| Ledger vs feature reconciliation | **37 = 37**, zero ledger headings without a Gherkin scenario |
| CI on current head | green on all four checks |
| Delta dependency/lockfile churn | zero files |

The fail-open path is gone by construction, not by convention: `scope` is required on both public detectors, the four wrappers are deleted across both mirrors, and — the part that mattered most — both `steps/` files now import the real `WORKSPACE_ROOTS` and pass a real scope, so the BDD lane finally exercises the production path instead of the permissive one.

Two fixes came out better than what I suggested: extracting `workspace-roots.ts` as a dependency-free module beats interpolating a literal, and `TicketChange.anchorScope` cleanly removed the derived-state duplication.

### Worth an explicit callout — not a defect

The namespace-precedence fix is new scope beyond my review. A repo with **both** `.project/` and `.safeword-project/` present now gets boundary checks on `.project/` tickets only — legacy-root tickets go silently unchecked mid-migration. That matches `resolveNamespaceRoot` (`configured-paths.ts:235-247`) exactly and four new tests pin both directions at both tiers, so it's the right call. Flagging it because it changes *which* tickets get policed, and that deserves to be a decision rather than a side effect.

### Also checked, no action needed

The `.ts` import specifier in both steps files is fine — root `tsconfig.json` includes only `packages/*/src/**/*` and `lint` typechecks `packages/cli` alone, so tsc never sees `steps/`. Don't "fix" it to `.js`.

### Not verified

I did not run the full suite, so I can't independently confirm or refute the Astro preset failure. `verify.md` now documents it honestly with the unchanged-from-`origin/main` evidence, which was the ask.

PR #1478, opened 2026-07-26
Keep override regressions fast for maintainers

TheMostlyGreat:
## Summary

- route every fixture upgrade through an install-disabled `runFixtureUpgradeWithoutInstall` helper that throws on reconciliation failure
- reuse the repository JavaScript toolchain only in TypeScript override fixtures, preflighting local ESLint and Prettier; Python continues to resolve Ruff from PATH
- reject hook spawn failures, non-zero exits, and Bun/Ruff infrastructure warnings before evaluating lint assertions
- replace the brittle source-grep boundary test with focused subprocess-boundary contracts and centralize repeated TypeScript fixture setup

## Root cause

`override-survival.test.ts` verifies customer configuration preservation and real generated-hook behavior, but every fresh fixture let `safeword upgrade` perform unrelated package and language-skill installation. Negative lint assertions could also appear green when the upgrade or a language tool failed, because an unchanged file and an absent rule name looked like success.

## Impact

- target file in the original full-suite comparison: 68.27s → 22.52s (45.75s / 67.0% faster)
- independent original full-suite comparison: 449.27s → 397.95s (51.32s / 11.4% faster)
- all ten historical TypeScript and Python override examples remain intact and still run the generated post-tool lint hook
- upgrade-time dependency installation remains covered by `integration/add-language.test.ts`
- failed upgrades and Bun/Ruff infrastructure failures now fail loudly instead of allowing false greens

## Validation

- focused override integration: 10/10 pass
- focused final quality-review set: 18/18 pass
- full local Vitest on current `main`: 5,503 passed, 5 skipped, 0 failed across 371 files
- Gherkin: 497 scenarios (494 passed, 3 skipped), 15,317 steps (15,313 passed, 4 skipped)
- ESLint, Gherkin lint, TypeScript checks, build, formatting, audit, and independent quality review pass
- verification record: `.project/tickets/5KHSQB-keep-override-regressions-fast/verify.md`

## Scope

This PR targets `main` directly and is caught up through `86a883dff`. It does not change Vitest scheduling, production upgrade behavior, lint-hook behavior, or the historical override coverage matrix.


packages/cli/tests/integration/override-survival.test.ts
```diff
@@ -45,15 +59,19 @@ async function runUpgradeAndAssertFileUnchanged(
   relativePath: string,
 ): Promise<void> {
   const before = readTestFile(projectDirectory, relativePath);
-  await runCli(['upgrade'], { cwd: projectDirectory });
+  await runCli(['upgrade'], { cwd: projectDirectory, env: SKIP_INSTALL_ENV });
```
TheMostlyGreat: **Reviewer** — 🔴 the symlink turns a future perf regression into repo corruption

This is the one I'd fix before merge.

If anyone drops `SKIP_INSTALL_ENV` from a `runCli` in this file, `upgrade` reaches `installDependencies`. `detectPackageManager` finds no lockfile in the temp fixture and doesn't walk up, and the CLI subprocess is spawned with `process.execPath` (node, not bun) — so it returns `npm`. That runs `npm install --save-dev …` with cwd set to the fixture, where `node_modules` resolves straight through the new symlink into the developer's real checkout.

Before this PR that mistake cost 45 seconds. Now it can mutate the tree.

(Reasoned from `src/utils/install.ts:44-52` + `tests/helpers.ts:405` — deliberately **not** executed, since executing it is the damage.)

Fix: move the skip out of the call sites into a helper, so there is no call site left to forget.

```ts
// helpers.ts
export async function runFixtureUpgrade(cwd: string) {
  return runCli(['upgrade'], { cwd, env: SKIP_INSTALL_ENV });
}
```

That makes the invariant structural instead of conventional — and it retires the grep test in the other file.

packages/cli/tests/override-survival-install-boundary.test.ts
```diff
@@ -0,0 +1,19 @@
...
+import { readRepoFile } from './helpers';
+
+const OVERRIDE_SUITE = 'packages/cli/tests/integration/override-survival.test.ts';
+
+describe('override-survival install boundary (5KHSQB)', () => {
+  it('reuses repository tooling, disables upgrade installs, and rejects launcher failures', () => {
+    const source = readRepoFile(OVERRIDE_SUITE);
+
+    expect(source.match(/linkRepoToolchain\(projectDirectory\);/g)).toHaveLength(4);
+    expect(source).toContain(
```
TheMostlyGreat: **Reviewer** — 🟠 suggest deleting this file

It asserts on another test file's source text, which guards the wrong layer and breaks for the wrong reasons:

- `toHaveLength(4)` goes red the moment someone adds a legitimate fifth fixture — in an unrelated file, with a message that doesn't explain why.
- `toContain("await runCli(['upgrade'], { cwd: projectDirectory, env: SKIP_INSTALL_ENV });")` breaks on an options-key reorder or a prettier rewrap. False reds teach people to edit the assertion, which is how a guard dies.
- It's a new pattern here. The repo's other four `readRepoFile` tests assert on *shipped content* — `SAFEWORD.md`, skills, `codex-plugin/hooks.json` — files whose text is the product. Test source isn't.
- Half of what it guards is already enforced at runtime (I proved the `bunx failed` guard fires — see the review body); the other half is better enforced by construction via the `runFixtureUpgrade` helper.

If a guard must stay, assert behavior rather than source text.

packages/cli/tests/integration/override-survival.test.ts
```diff
@@ -267,6 +287,7 @@ select = ["E", "F"]
       );
       initGitRepo(projectDirectory);
       await setupOrThrow(projectDirectory);
+      linkRepoToolchain(projectDirectory);
```
TheMostlyGreat: **Reviewer** — 🟠 dead weight: the Python fixtures don't need the toolchain link

In my mutation run (symlink disabled), all six Python scenarios still passed — only the four TypeScript ones failed. Ruff resolves from PATH, never from `node_modules`.

So two of the four links buy nothing, and they hand Python fixtures a writable handle into the repo tree, compounding the hazard flagged on the upgrade call. Same applies to line 386.

`toHaveLength(4)` in the boundary test is what currently locks them in — both go away together.

packages/cli/tests/integration/override-survival.test.ts
```diff
@@ -22,13 +22,27 @@ import {
...
   TIMEOUT_BUN_INSTALL,
   writeTestFile,
 } from '../helpers';
 
+function linkRepoToolchain(projectDirectory: string): void {
+  const repoNodeModules = nodePath.join(repoRoot, 'node_modules');
+  if (!existsSync(repoNodeModules)) {
+    throw new Error(`Repository toolchain is missing: ${repoNodeModules}`);
+  }
+  symlinkSync(
```
TheMostlyGreat, line 52: **Reviewer** — 🟡 unguarded `EEXIST`

`symlinkSync` throws `EEXIST` if `node_modules` already exists (verified locally). Setup can't create it today — both deps and the skills pull are skipped — but if that ever changes, the failure surfaces inside `beforeAll` as a raw fs error with no hint about what happened.

One line:

```ts
const target = nodePath.join(projectDirectory, 'node_modules');
if (existsSync(target)) return;
```

packages/cli/tests/integration/override-survival.test.ts
```diff
@@ -45,15 +59,19 @@ async function runUpgradeAndAssertFileUnchanged(
...
 }
 
 function runHookAndGetOutput(projectDirectory: string, violationRelativePath: string): string {
   const violationPath = nodePath.join(projectDirectory, violationRelativePath);
   const hookResult = runLintHook(projectDirectory, violationPath);
-  return `${hookResult.stdout ?? ''}${hookResult.stderr ?? ''}`;
+  const hookOutput = `${hookResult.stdout ?? ''}${hookResult.stderr ?? ''}`;
+  expect(hookResult.error).toBeUndefined();
+  expect(hookResult.status).toBe(0);
+  expect(hookOutput).not.toContain('bunx failed');
```
TheMostlyGreat: **Reviewer** — ✅ confirmed this guard is real

Worth recording, since a negative assertion guarding against vacuous negative assertions deserves scrutiny.

I disabled `linkRepoToolchain` and re-ran the suite. All four TypeScript scenarios failed here with:

```
AssertionError: expected 'bunx failed: Oops! Something went wro…' not to contain 'bunx failed'
```

The string is constructed at runtime by `.safeword/hooks/lib/lint.ts:260` (`` warnings.push(`${command[0]} failed: ${stderr.split('\n')[0]}`) ``), so `'bunx failed'` is a genuine literal in the failure path, not a hopeful match. Scenario 1.4 — which has only a negative assertion and no positive companion — was caught too.

No change requested. Noting it so the next reader doesn't have to redo the mutation.

TheMostlyGreat (commented):
**Reviewer** — quality + engineering review

**Verdict: request changes** — nothing is broken today. The suite is green, the speedup is real, and the guards are genuine. The ask is one consolidated cleanup that removes a latent hazard and a brittle file. Inline comments carry the specifics.

---

### What I verified (not assumed)

**The guard is load-bearing, not decorative.** I neutered `linkRepoToolchain` and re-ran: all 4 TypeScript scenarios went red with `bunx failed: Oops! Something went wrong! :(` — including Scenario 1.4, which has only a negative assertion. The `not.toContain('bunx failed')` string is constructed at runtime by `.safeword/hooks/lib/lint.ts:260` (`` `${command[0]} failed: …` ``), so it genuinely catches the vacuous-pass mode this PR was written to close.

**`SKIP_INSTALL_ENV` on `upgrade` does not weaken the override-preservation assertion.** The env var gates exactly two things: `installDependencies` (`src/utils/install.ts:65`) and the skills pull. `reconcile`, `syncAndRecordPackageJsonSafewordVersion`, and `maybeAutoPatchOrNudge` — every path that could touch `eslint.config.mjs` — still run. The `skipPackageChecks` flag at `src/commands/upgrade.ts:369` only reaches `selfVerify`'s health report, which by its own doc comment cannot change exit status. The "file unchanged" assertions are as strong as before.

**Cleanup won't eat the repo's `node_modules`.** `rmSync({recursive: true})` unlinks the symlink rather than deleting through it — canary file survived on Node v24.16.0.

**Local run green:** 6 files, 40 tests, 37.70s. ESLint and `tsc --noEmit` clean on both changed files. Worktree restored clean after the mutation.

---

### Review dimensions

| | |
|---|---|
| Currency | ✅ No dependency or version claims in the change |
| Sources | ✅ Every load-bearing claim traced to code read or a command run |
| Correct | ✅ Guards are real; install-skip doesn't weaken assertions |
| Elegant | ⚠️ Boundary test greps another test file's source |
| No-bloat | ⚠️ Two of four toolchain links are dead weight, cemented by a count assertion |
| Wiring | ✅ Fixtures still drive real collaborators — real `safeword upgrade` subprocess, real generated hook, real eslint/ruff. Only the process boundary (dependency install) is stubbed, via env. Correct seam. |

---

### Notes (non-blocking)

- **Name the lane you handed off to.** These ten examples now lint against the *repo's* eslint and plugin versions, not what safeword installs for a customer. Right trade for this suite — but the file header still reads as if it proves end-to-end setup. Out-of-Scope says install lives in its own lanes; naming that lane in the header comment keeps the next reader from assuming coverage that moved.
- **CI is ubuntu-only** (all nine workflows `runs-on: ubuntu-latest`), so the `win32` junction branch is untested. It's correct per Node's docs — `type` is Windows-only and junction targets are auto-normalized to absolute — so leave it, just don't count it as covered.
- **Perf figures are self-reported.** I didn't reproduce the in-suite 68.27→22.52s measurement. The mechanism is sound (ten upgrades × one install each), so the direction holds; the exact numbers are unverified.
- Base branch is two commits behind `main`; GitHub reports `MERGEABLE`.

---

**Next:** add `runFixtureUpgrade` to `helpers.ts` and route all upgrade calls through it, delete `override-survival-install-boundary.test.ts`, drop the two Python `linkRepoToolchain` calls. Re-run `tests/integration/override-survival.test.ts` — that's the whole re-review.

**Sources for the Windows/junction note:** [File system — Node.js docs](https://nodejs.org/api/fs.html), [fs.symlink — Node API reference](https://docs.deno.com/api/node/fs/~/symlink). Everything else above was verified by running it against this branch this session (mutation run, `rmSync` symlink probe, ESLint/`tsc`), not by fetching a doc.

packages/cli/tests/integration/override-survival.test.ts
```diff
@@ -5,30 +5,63 @@
  * customer-owned config file is (a) honored by the LLM hook lint run and
  * (b) not mutated by `safeword upgrade`.
  *
- * See `.safeword-project/tickets/137-customer-override-survival/test-definitions.md`
+ * Dependency installation is covered by `conditional-setup.slow.test.ts`.
```
TheMostlyGreat: **Reviewer** — 🟡 non-blocking: this points at the one lane CI doesn't run

The handoff is the right idea, but `conditional-setup.slow.test.ts` is an awkward target for two reasons:

1. **It's excluded from the default run** (its own header says so — it needs `test:slow` / `vitest.slow.config.ts`), and CI only invokes `test`, `test:bdd`, and `test:release`. So a reader who follows this pointer lands on coverage that never executes in CI.
2. **It covers install during `setup`**, whereas what this PR removed is install during `upgrade`.

The coverage you actually handed off to does exist and does run in CI — [`tests/integration/add-language.test.ts:43`](packages/cli/tests/integration/add-language.test.ts#L43) sets `INSTALL_DEPENDENCIES_ENV`, then upgrades at `:75`. Naming that one instead would make the pointer true and useful.

(Verified this session: workflow greps + file reads.)

packages/cli/tests/helpers.ts
```diff
@@ -994,6 +1016,27 @@ export function runLintHook(
...
+export function assertLintHookSucceeded(result: SpawnSyncReturns<string>): string {
+  if (result.error) {
+    throw new Error(`Lint hook failed to start: ${result.error.message}`);
+  }
+  if (result.status !== 0) {
+    throw new Error(`Lint hook exited with status ${String(result.status)}`);
+  }
+
+  const output = `${result.stdout ?? ''}${result.stderr ?? ''}`;
+  if (/\b(?:bunx|ruff) failed:/iu.test(output)) {
```
TheMostlyGreat, line 1036: **Reviewer** — 🟡 non-blocking: the guard is allowlist-shaped

I probed this regex against the real runtime strings. It matches exactly the two tools this suite drives, and nothing else:

```
true   "bunx failed: Oops! Something went wrong! :("
true   "ruff failed: invalid config"
false  "golangci-lint failed: x"
false  "clippy failed: x"
```

Correct for today's scope — but if a Go or Rust override scenario ever lands in this suite, the guard silently won't cover it, and the failure mode is a green test that asserted nothing. Worth a comment naming the coupling to the `` `${command[0]} failed:` `` construction at `.safeword/hooks/lib/lint.ts:260`, so the next person adding a language knows this line is part of the contract.

Separately — the `.bin/prettier` preflight above is the right call *precisely because* the hook can't report a Prettier failure: `runPrettier` swallows it with `.nothrow().quiet()`, so no `prettier failed:` warning is ever produced. Link-time preflight is the only place that failure can surface. Nice.

packages/cli/tests/helpers.ts
```diff
@@ -629,6 +629,28 @@ function buildSetupFailureError(
...
 
+/**
+ * Run a fixture upgrade with package and skill installation disabled.
+ *
+ * Fixtures that link repository tooling into `node_modules` must never invoke
+ * an install-capable upgrade: a package manager could otherwise write through
+ * the link into the developer checkout. Keep that invariant in this helper
+ * instead of repeating environment options at call sites.
+ */
+export async function runFixtureUpgradeWithoutInstall(
```
TheMostlyGreat, line 640: **Reviewer** — ✅ this closes the hazard properly

Exactly the shape I was hoping for. The invariant is now structural rather than conventional: there's no call site that can forget the env, because there's no env at any call site. The doc comment naming the *why* — a package manager writing through the link into the developer checkout — is what keeps someone from "simplifying" it back out in six months.

Matching the existing `setupOrThrow` injectable-runner convention rather than inventing a new mocking style is the right call too.

packages/cli/tests/fixture-upgrade.test.ts
```diff
@@ -0,0 +1,46 @@
...
+        args: ['upgrade'],
+        options: {
+          cwd: '/fake/project',
+          env: { SAFEWORD_SKIP_INSTALL: '1' },
+        },
+      },
+    ]);
+  });
+
+  it('rejects a failed upgrade instead of allowing an unchanged fixture to false-green', async () => {
```
TheMostlyGreat, line 33: **Reviewer** — ✅ this is a false-green I missed

Worth calling out explicitly, because it was hiding inside an assertion I'd just reviewed and certified.

The original did `await runCli(['upgrade'], …)` and discarded the result. A crashed upgrade leaves the config file untouched — so `expect(after).toBe(before)` passes, and the scenario reports success having verified nothing. That's a third false-green path alongside the two named in the PR description, and I read straight past it last pass while confirming the install-skip didn't weaken the same assertion.

Throwing on non-zero exit is the right fix, and testing it against a synthetic exit 23 rather than trying to crash a real upgrade keeps it fast and deterministic.

packages/cli/tests/integration/override-survival.test.ts
```diff
@@ -5,30 +5,63 @@
...
   TIMEOUT_BUN_INSTALL,
   writeTestFile,
 } from '../helpers';
 
+function linkRepoToolchain(projectDirectory: string): void {
+  const repoNodeModules = nodePath.join(repoRoot, 'node_modules');
+  if (!existsSync(repoNodeModules)) {
+    throw new Error(`Repository toolchain is missing: ${repoNodeModules}`);
+  }
+  for (const executable of ['eslint', 'prettier']) {
```
TheMostlyGreat, line 43: **Reviewer** — ✅ preflight is a better guard than I suggested

I'd only asked for the `EEXIST` fix (now at line 51). Checking that `.bin/eslint` and `.bin/prettier` actually exist before linking is the stronger version: it fails at link time with a precise message, rather than surfacing later as a confusing lint-hook infrastructure error three scenarios deep.

This is also what makes the Prettier path safe at all — see the note on `helpers.ts:1034`.

TheMostlyGreat (commented):
**Reviewer** — second pass, re-review at `c0675510e`

**Verdict: approve.** All five findings from the [first review](https://github.com/ArcadeAI/safeword/pull/1478#pullrequestreview-4782981401) are addressed, and two of the fixes go beyond what I asked for. No critical issues. Two non-blocking notes inline.

(Posted as a comment rather than a formal approval — GitHub won't accept a self-approval on your own PR.)

---

### Finding disposition

| First pass | Now |
|---|---|
| 🔴 Symlink + dropped env → writes into the real checkout | **Fixed structurally.** `runFixtureUpgradeWithoutInstall` hard-codes the env; no call site left to forget. Doc comment names the exact hazard. |
| 🟠 Delete the source-text grep test | **Deleted.** Replaced by `fixture-upgrade.test.ts` + `lint-hook-result.test.ts` — behavioral, injected runners. |
| 🟠 Python fixtures don't need the toolchain link | **Fixed.** `setupTypeScriptFixture` is TypeScript-only; Python blocks link nothing. |
| 🟡 Unguarded `EEXIST` | **Fixed.** `if (existsSync(fixtureNodeModules)) return;` |
| 🟡 Name the handoff lane | **Done**, with one caveat — see the inline note on the header. |

---

### Two fixes I didn't ask for, and should have

**The original ignored `upgrade`'s exit code entirely.** A crashed upgrade leaves the file unchanged, so `expect(after).toBe(before)` passed vacuously — a third false-green path, sitting inside the assertion I had just certified as "as strong as before." Now it throws, with unit coverage for exit 23. Good catch; I missed it.

**The guard now covers Ruff.** The old `'bunx failed'` string never protected the six Python scenarios at all. I probed `/\b(?:bunx|ruff) failed:/iu` directly against both real runtime strings — both match.

---

### Verified by running it, not by reading the diff

- **Mutation re-run** — disabled the symlink; all 4 TypeScript scenarios failed with `Lint hook reported an infrastructure failure`. The guard survived the refactor, and now fails with a readable message instead of an assertion diff.
- **47/47 pass** across the three affected files (36.47s vitest; the 6:42 wall clock was the machine-global build lock queueing behind another worktree, not the suite).
- **ESLint and `tsc --noEmit` both exit 0.**
- Both new doc claims check out: `conditional-setup.slow.test.ts` and `.project/tickets/completed/137-customer-override-survival/test-definitions.md` both exist.
- Worktree restored clean after the mutation.

### Review dimensions

| | |
|---|---|
| Currency | ✅ No dependency changes, no new API surface |
| Sources | ✅ Both new doc references traced to real files |
| Correct | ✅ Guards still fire — proved by mutation |
| Elegant | ✅ Grep test gone; invariants live in named helpers with contract tests |
| No-bloat | ✅ Python fixtures no longer link a toolchain they never used |
| Wiring | ✅ Integration suite still drives real collaborators; contract tests inject only the subprocess boundary — correct seam |

**Provenance:** every claim above was verified by execution this session — mutation run, test run, ESLint/`tsc`, regex probe, file and workflow reads. Nothing rests on training data.

---

**Next:** merge. Both inline notes are one-line comment edits that can't affect behavior — fold them in first if convenient, but they don't warrant another review round.

PR #1497, opened 2026-07-26
Accept and fully redact stateless GitHub installation tokens

TheMostlyGreat:
Closes #1495

## What changed

- accept GitHub's stateless JWT-like `ghs_` installation-token format while keeping whole-value validation anchored
- redact the complete stateless token before the classic `ghs_` rule can partially consume it
- cover the exact 35/36-character boundary, leading and trailing anchor rejection, representative ~520-character `ghs_APPID_JWT`, ordering mutations, `-`/`_` tails, classic regression, and REST transport hand-off

## Root cause

The validator allowed only alphanumeric characters after GitHub token prefixes, so dots and additional underscores caused valid stateless installation tokens to fall through. The egress scrubber's classic `\w` matcher stopped at the first dot, leaving the payload and signature in output.

## Impact

Retro can use either installation-token format during GitHub's staged rollout, and stateless credentials are fully removed before public issue filing.

## Validation

- safeword verify: 5,494 tests passed, 5 skipped
- Gherkin: 494 scenarios passed, 3 skipped; 15,313 steps passed, 4 skipped
- focused retro tests: 82/82 passed
- explicit mutation checks killed removal of `$`, loosening `{36,}`, deletion of the stateless scrubber rule, and moving it after the classic rule
- build, TypeScript, ESLint, Prettier, Markdownlint, and `git diff --check` passed
- safeword audit: no scoped errors; config sync, dependency boundaries, Knip, domain docs, and dependency freshness clean
- independent quality review: PASS, no findings or suggestions

## Release evidence

Live GitHub App rollout validation remains a release activity: force `X-GitHub-Stateless-S2S-Token: enabled`, force `disabled`, then remove the override, without logging credentials.


packages/cli/src/retro/github-rest.test.ts
```diff
@@ -527,9 +534,28 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', ()
...
+
+    expect((calls[0]?.init.headers as Record<string, string>).Authorization).toBe(
+      `Bearer ${shaped}`,
+    );
+  });
+
   it.each([
     ['a proxy placeholder', 'proxy-injected'],
     ['an unknown prefix', `gha_${'a'.repeat(32)}`],
+    ['a stateless token embedded in prose', `prefix-ghs_${'a'.repeat(40)}.${'b'.repeat(40)}`],
```
TheMostlyGreat, line 559: Reviewer: the trailing `$` anchor is unpinned. I mutated `^ghs_[\w.-]{36,}$` -> `^ghs_[\w.-]{36,}` and the whole suite still passes (38/38).

#1495's AC says the validator stays anchored (`^...$`), but this fixture only exercises the *leading* anchor — the junk is on the front. It matters concretely: `resolveGitHubToken` doesn't trim, so without `$` a `GITHUB_TOKEN` of `ghs_<36+ valid chars>` followed by whitespace and junk is accepted and interpolated straight into `Bearer ${token}`.

One more fixture in this `it.each` closes it:

```ts
['a stateless token with trailing junk', `ghs_${'a'.repeat(40)}.${'b'.repeat(40)} extra`],
```

packages/cli/src/retro/github-rest.test.ts
```diff
@@ -507,6 +507,11 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', ()
   // a literal that looks like a real credential.
   const envToken = `ghp_${'a'.repeat(32)}`;
   const ghToken = `ghp_${'b'.repeat(32)}`;
+  // GitHub's published matcher requires at least 36 allowed characters after
+  // `ghs_`; pin that exact boundary separately from the representative ~520-char
+  // `ghs_APPID_JWT` rollout shape.
+  const minimumStatelessToken = `ghs_${'a'.repeat(10)}_${'b'.repeat(10)}.${'c'.repeat(6)}.${'d'.repeat(7)}`;
```
TheMostlyGreat, line 513: Reviewer: the 36-char minimum is pinned only on the accept side — loosening the validator to `{20,}` also survives the suite (38/38 pass).

Low severity: a looser validator just means a malformed token gets a 401 instead of falling back to `gh`. But the PR body advertises "exact 36-character minimum", and right now only one side of the boundary is proven. A 35-char rejection fixture in the `rejects %s and falls back to gh` block would pin both.

packages/cli/src/retro/github-rest.ts
```diff
@@ -59,6 +59,7 @@ function ghAuthToken(): string | undefined {
  */
 function looksLikeGitHubToken(value: string): boolean {
   return (
+    /^ghs_[\w.-]{36,}$/.test(value) ||
```
TheMostlyGreat, line 64: Reviewer: nit — this line carries no comment, unlike its `egress.ts` twin, and the docblock just above still describes only `ghp_`/`gho_`/`ghu_`/`ghs_`/`ghr_` with no mention of the stateless format.

`36` is a GitHub-published constant, not a chosen one. Without the citation, someone will eventually "simplify" this back to `[A-Za-z0-9]{20,}` and silently reintroduce #1495. Suggest:

```ts
// GitHub stateless installation tokens — GitHub's published matcher (changelog 2026-05-15)
/^ghs_[\w.-]{36,}$/.test(value) ||
```

packages/cli/src/retro/egress.test.ts
```diff
@@ -149,6 +149,51 @@ describe('scrubSecrets — formats from review #543 round 2', () => {
...
+  const token = `ghs_${'A'.repeat(40)}.${'B'.repeat(40)}.${'C'.repeat(40)}`;
+  const minimumToken = `ghs_${'A'.repeat(10)}_${'B'.repeat(10)}.${'C'.repeat(6)}.${'D'.repeat(7)}`;
+
+  function scrubWith(patterns: readonly RegExp[]): string {
+    let out = `token is ${token}`;
+    for (const pattern of patterns) out = out.replaceAll(pattern, () => redacted);
+    return out;
+  }
+
+  it('pins the four load-bearing rule-ordering mutations', () => {
```
TheMostlyGreat: Reviewer: this test pins nothing about production. It re-declares local copies of the three regexes and never calls `scrubSecrets`, so it would pass with `SECRET_PATTERNS` emptied entirely.

I verified that empirically: across five production mutations (delete the stateless rule, move it after the classic rule, add a trailing `\b`, drop `-` from the char class, `{36,}`->`{60,}`) this test never failed once. Every kill came from the `it.each` below — that's the real ordering guard.

It's fine as executable documentation of *why* the ordering matters, but the name claims to be the guard, which invites a future reader to delete the `it.each` as redundant. Two suggestions: rename to something like `documents why stateless-before-classic ordering matters`, and note that these local copies drift silently if `SECRET_PATTERNS` changes.

docs/user-stories/1495-stateless-github-installation-tokens.md
```diff
@@ -0,0 +1,40 @@
+# User stories: accept and redact stateless GitHub installation tokens
```
TheMostlyGreat: Reviewer: `docs/user-stories/` and `docs/test-definitions/` are both new directories — `git log --diff-filter=A` shows no prior history for either, and each holds exactly this one file.

The canonical homes in this repo are `<namespace-root>/tickets/{id}-{slug}/spec.md` + `test-definitions.md` (SAFEWORD.md:48, :177) and root `features/*.feature`; 454 tickets live under `.project/tickets/`. Nothing under `docs/` is visible to the phase gate, coverage/lineage, or the done-flip guard — so this is unenforced prose that will drift, and a parallel artifact tree alongside the real one.

This is also a bug fix, which the bdd flow explicitly excludes from the spec/test-definitions chain.

Suggest deleting both: #1495's body, the PR description, and the load-bearing code comments already carry the rationale, and the tests are the durable spec. If you do want them durable, they belong in a ticket dir where the gates can see them.

TheMostlyGreat (commented):
Reviewer: engineering + quality review at `100fa541a`.

**Production logic is sound.** Both regexes are byte-equivalent to GitHub's published matcher `ghs_[A-Za-z0-9\.\-_]{36,}` (`[\w.-]` = `[A-Za-z0-9_.\-]`) — verified against the [2026-05-15 changelog](https://github.blog/changelog/2026-05-15-github-app-installation-tokens-per-request-override-header/). Fixture lengths check out: the minimum fixtures are exactly 36 chars after the prefix, the representative one is exactly 520. Single char class, no nested quantifiers, so no ReDoS surface.

**Mutation-tested the egress side** against the real `SECRET_PATTERNS` — every one is guarded:

| Mutation | Result |
| --- | --- |
| Delete stateless rule from `SECRET_PATTERNS` | 3 tests fail |
| Move it *after* the classic rule | 3 tests fail |
| Add a trailing `\b` to it | 1 test fails |
| Drop `-` from its char class | 1 test fails |
| `{36,}` -> `{60,}` | 1 test fails |

So the "deliberately no trailing `\b`" comment is genuinely load-bearing and genuinely pinned.

**Also probed the layered path,** since the tests only cover the sync floor: fed `sanitizeTextDeep` a realistic token whose JWT header actually begins `eyJ`, to check whether secretlint could match-and-replace the `ghs_...` prefix first and strand `.payload.signature` for a floor that can no longer see a `ghs_` prefix — the same failure shape this PR fixes, one layer up. It doesn't happen: output is `token is [redacted]`, no residue.

Inline: one test gap I'd close before merge (the trailing `$` anchor is unpinned), one placement question on the two new `docs/` trees, and three nits.

packages/cli/src/retro/github-rest.test.ts
```diff
@@ -527,9 +534,28 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', ()
     expect(ghConsulted).toBe(false);
   });
 
+  it.each([
+    ['classic opaque', `ghs_${'a'.repeat(40)}`],
+    ['stateless', representativeStatelessToken],
+  ])('passes the selected %s installation token to the REST transport', async (_label, shaped) => {
```
TheMostlyGreat: Reviewer: positive note, so this doesn't get churned while addressing the other comments — this is a real wiring test and it's the right shape.

It composes the actual `resolveGitHubToken` -> `createRestTransport` chain with real collaborators and mocks only `fetch` (the process boundary), then asserts the exact `Bearer <token>` header that goes over the wire. That's what catches a validator/transport seam a fully-mocked suite would miss — e.g. a token accepted by `looksLikeGitHubToken` but mangled before it reaches the header. The `gh` fallback throwing on call is a nice touch: it proves selection came from the env var rather than coincidentally matching.

Keep it as-is.

packages/cli/src/retro/github-rest.test.ts
TheMostlyGreat: Addressed in 3c3dd4925. Added a trailing-prose rejection fixture and confirmed it kills removal of the trailing `$` anchor.

packages/cli/src/retro/github-rest.test.ts
TheMostlyGreat: Addressed in 3c3dd4925. Added a 35-character dotted stateless fixture so the `{36,}` lower boundary is pinned without conflicting with the legacy opaque `ghs_` format.

packages/cli/src/retro/github-rest.ts
TheMostlyGreat: Addressed in 3c3dd4925. The validator docblock now cites GitHub’s published `{36,}` matcher and changelog.

packages/cli/src/retro/egress.test.ts
TheMostlyGreat: Addressed in 3c3dd4925. Renamed this as executable documentation and added a drift warning; the production `scrubSecrets` table remains the actual guard.

docs/user-stories/1495-stateless-github-installation-tokens.md
TheMostlyGreat: Addressed in 3c3dd4925. Deleted both parallel planning artifacts; the final PR contains only the canonical code and test changes for this bug fix.

packages/cli/src/retro/github-rest.test.ts
TheMostlyGreat: Thanks. Kept this wiring test unchanged; the final independent review also confirmed that it uses real collaborators and mocks only the fetch boundary.

PR #1521, opened 2026-07-27
Keep lint-hook edits free of Safeword upgrades

TheMostlyGreat:
## Summary

- remove the edit-time lint hook's duplicate language-pack repair path
- fall back to existing tool behavior when generated configs are absent, without running Safeword upgrade, staging, or commit commands
- preserve generated config arguments when configs exist across TypeScript, Python, Go, Rust, and SQL
- document the missing-config behavior and keep the shipped template and dogfood hook aligned

## Root cause

The lint hook treated a missing language-pack config as a repair trigger. That allowed an ordinary post-edit lint pass to run `safeword upgrade` and then stage and commit repository changes, duplicating the session-start upgrade owner and introducing network and git side effects into the edit path.

## Validation

- 5,515 Vitest tests passed; 5 skipped
- 494/497 Gherkin scenarios passed; 3 expected skips
- build, lint, and TypeScript typecheck passed
- real subprocess matrix covers configured and missing-config behavior for both hook copies across all five language branches
- final independent quality review approved with no correctness, security, scope, test, or documentation findings

Closes #1511

### DIFF
--- a/packages/cli/templates/hooks/lib/lint.ts
+++ b/packages/cli/templates/hooks/lib/lint.ts
@@ -4,9 +4,9 @@
 // Uses explicit --config flags pointing to .safeword/ configs for LLM enforcement.
 // This allows stricter rules for LLMs while humans use their normal project configs.
 //
-// Missing language-pack configs degrade to each linter's defaults. Upgrades are
-// owned by the session-start hook, so missing configs never trigger upgrade,
-// staging, or commit side effects from linting.
+// Missing language-pack configs never trigger upgrade, staging, or commit side
+// effects from linting. Version upgrades run at session start; pack repair is
+// manual, so fallback linting reports the missing config once per session.
 
 import { existsSync, readFileSync, realpathSync } from 'node:fs';
 import nodePath from 'node:path';
@@ -218,6 +218,23 @@ function configArgs(configPath: string): string[] {
   return hasConfig(configPath) ? ['--config', configPath] : [];
 }
 
+/** Warn once per session when fallback linting cannot enforce Safeword rules. */
+function warnMissingSafewordConfig(
+  packName: string,
+  tool: string,
+  configPath: string,
+  warnings: string[],
+): void {
+  if (hasConfig(configPath)) return;
+  const warningKey = `pack:${packName}`;
+  if (toolWarnings.has(warningKey)) return;
+  toolWarnings.add(warningKey);
+  warnings.push(
+    `${packName} Safeword config is missing — linting with ${tool} defaults, not Safeword rules. ` +
+      'Run `safeword setup` or `safeword upgrade` to install it.',
+  );
+}
+
 /** Run prettier with safeword config if available */
 async function runPrettier(file: string): Promise<void> {
   // A non-Prettier formatter owns this repo — defer to it, don't restyle (V7GGJZ).
@@ -275,6 +292,7 @@ export async function lintFile(file: string, _projectDir: string): Promise<LintR
       };
     }
     if (host) return runHostToolchain(host);
+    warnMissingSafewordConfig('TypeScript', 'ESLint', SAFEWORD_ESLINT, warnings);
     const configArguments = configArgs(SAFEWORD_ESLINT);
     await $`bunx eslint ${configArguments} --fix ${normalizedFile}`.nothrow().quiet();
     await runPrettier(normalizedFile);
@@ -297,6 +315,7 @@ export async function lintFile(file: string, _projectDir: string): Promise<LintR
     ) {
       return { warnings };
     }
+    warnMissingSafewordConfig('Python', 'Ruff', SAFEWORD_RUFF, warnings);
     const configArguments = configArgs(SAFEWORD_RUFF);
     await $`ruff check ${configArguments} --fix ${normalizedFile}`.nothrow().quiet();
     await $`ruff format ${configArguments} ${normalizedFile}`.nothrow().quiet();
@@ -336,6 +355,7 @@ export async function lintFile(file: string, _projectDir: string): Promise<LintR
       }
       toolWarnings.add('golangci-lint-v2-ok');
     }
+    warnMissingSafewordConfig('Go', 'golangci-lint', SAFEWORD_GOLANGCI, warnings);
     const configArguments = configArgs(SAFEWORD_GOLANGCI);
     await $`golangci-lint run ${configArguments} --fix ${normalizedFile}`.nothrow().quiet();
     await $`golangci-lint fmt ${configArguments} ${normalizedFile}`.nothrow().quiet();
@@ -349,6 +369,7 @@ export async function lintFile(file: string, _projectDir: string): Promise<LintR
   // Rust files - clippy for linting (package-level), rustfmt for formatting (file-level)
   if (RUST_EXTENSIONS.has(extension)) {
     const hasRustConfig = hasConfig(SAFEWORD_RUSTFMT);
+    warnMissingSafewordConfig('Rust', 'rustfmt', SAFEWORD_RUSTFMT, warnings);
 
     // Run clippy with package targeting for workspaces
     const packageName = detectRustPackage(normalizedFile);
--- a/packages/cli/templates/hooks/lib/lint.ts
+++ b/packages/cli/templates/hooks/lib/lint.ts
@@ -4,7 +4,9 @@
 // Uses explicit --config flags pointing to .safeword/ configs for LLM enforcement.
 // This allows stricter rules for LLMs while humans use their normal project configs.
 //
-// Auto-upgrades safeword if a language pack is missing.
+// Missing language-pack configs degrade to each linter's defaults. Upgrades are
+// owned by the session-start hook, so missing configs never trigger upgrade,
+// staging, or commit side effects from linting.
 
 import { existsSync, readFileSync, realpathSync } from 'node:fs';
 import nodePath from 'node:path';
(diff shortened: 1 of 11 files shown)


packages/cli/templates/hooks/lib/lint.ts
```diff
@@ -4,7 +4,9 @@
 // Uses explicit --config flags pointing to .safeword/ configs for LLM enforcement.
 // This allows stricter rules for LLMs while humans use their normal project configs.
 //
-// Auto-upgrades safeword if a language pack is missing.
+// Missing language-pack configs degrade to each linter's defaults. Upgrades are
+// owned by the session-start hook, so missing configs never trigger upgrade,
```
TheMostlyGreat: **Finding 1 — the load-bearing claim here is only half true.**

"Owned by the session-start hook" holds for *version* upgrades. It does not hold for *pack repair*, and this PR drops that rather than relocating it:

- `session-auto-upgrade` returns `{ kind: 'skipped', reason: 'latest is not newer' }` (`auto-upgrade.ts:317`) whenever the installed version is current. A missing `.safeword/ruff.toml` on an up-to-date install never reaches it.
- `session-lint-check.ts` inspects the **host's** eslint/prettier config and `package.json`. It never looks at `.safeword/` pack configs.

Concrete case: a TypeScript-only repo later gains Python. `.safeword/ruff.toml` never existed, no newer release is pending. Ruff runs with stock defaults forever, and nothing says so — I ran the real hook in a stubbed temp project and got `{"warnings":[]}`.

Issue #1511 proposed dropping repair **and** having the hook report a missing pack. This PR did the dropping half only. The machinery for the other half is already in this file — `toolWarnings` gives once-per-session dedupe, `warnings` reaches the agent via `additionalContext`:

```ts
if (!hasConfig(configPath) && !toolWarnings.has(`pack:${packName}`)) {
  toolWarnings.add(`pack:${packName}`);
  warnings.push(
    `${packName} safeword config missing — linting with ${tool} defaults, not safeword rules. ` +
      `Run \`bunx safeword upgrade\` to install the pack.`,
  );
}
```

Not a blocker — this PR still strictly improves on what it replaces. But either add the warning, or reword this comment so it doesn't imply repair has an owner.

packages/website/src/content/docs/reference/configuration.mdx
```diff
@@ -90,6 +90,8 @@ Safeword auto-detects your stack and configures linting accordingly:
 
 Detection is stored in `.safeword/config.json`. To force a refresh, delete that file and run `safeword setup` again.
 
+Edit-time linting uses the generated language-pack config when it exists. For JavaScript/TypeScript, Python, Go, and Rust, a missing config falls back to the tool's defaults. SQLFluff runs only when its generated config exists. In either case, the lint hook does not repair the pack, run `safeword upgrade`, stage files, or create a commit. Safeword updates remain the responsibility of the session-start auto-upgrade.
```
TheMostlyGreat: **Finding 1, docs half.** Same issue as the header comment in `lint.ts`: "Safeword updates remain the responsibility of the session-start auto-upgrade" reads to a user as though a missing pack will eventually get fixed for them. It won't — session-start only fires when a newer version exists, so on a current install a missing pack stays missing.

The honest version is that repair is manual. Something like: *"A missing pack config is not repaired automatically — run `safeword upgrade` or `safeword setup` to install it."*

If you take the warning-on-missing-pack route in `lint.ts`, this paragraph becomes true as written and only needs the manual-repair pointer added.

packages/cli/tests/hooks/lint-config-fallback.test.ts
```diff
@@ -0,0 +1,226 @@
...
+    stubCommands: ['rustfmt'],
+    expectedInvocations: [/^rustfmt --config-path .*\/\.safeword\/rustfmt\.toml .*source\.rs$/],
+  },
+  {
+    languageName: 'SQL',
+    extension: 'sql',
+    configRelativePath: undefined,
+    forbidConfigArgument: true,
+    stubCommands: [],
+    expectedInvocations: [],
```
TheMostlyGreat, line 140: **Finding 2 (nit) — this case asserts nothing positive.**

`expectedInvocations: []` flows into `expect.arrayContaining([])` at line 208, which passes against any invocation list whatsoever. The negative assertions below still do real work, so the case isn't worthless — but the positive half is vacuous.

For SQL with no host plugin declared and no sqlfluff config, the correct behavior is *zero* subprocesses (early return). Asserting that exactly would catch a regression where prettier starts getting invoked on `.sql` files in non-SQL projects:

```ts
if (expectedInvocations.length === 0) expect(invocations).toEqual([]);
```

packages/cli/tests/hooks/lint-no-auto-upgrade.test.ts
```diff
@@ -0,0 +1,226 @@
...
+        cwd: REPO_ROOT,
+        env: {
+          ...process.env,
+          CLAUDE_PROJECT_DIR: directory,
+          PATH: `${stubBin}${path.delimiter}${process.env.PATH ?? ''}`,
+        },
+        timeout: 30_000,
+      });
+
+      expect(JSON.parse(stdout.trim())).toEqual({ warnings: [] });
```
TheMostlyGreat: **Coupling worth knowing about, not a defect.**

This pins `warnings: []` across all ten missing-config cases, which means the *silence* described in finding 1 is now enforced by tests. If the missing-pack warning lands later as a follow-up, it has to touch this line — that's expected, not a regression. Flagging so it doesn't read as one six weeks from now.

packages/cli/tests/hooks/lint-config-fallback.test.ts
```diff
@@ -0,0 +1,226 @@
+import { execFile } from 'node:child_process';
```
TheMostlyGreat, line 1: **Finding 3 (nit) — filename names the deleted thing.**

`lint-no-auto-upgrade.test.ts` is named for `ensurePackInstalled`, which this PR removes from the tree. Once it's gone, "no-auto-upgrade" stops being a searchable anchor and the file reads as a guard against a ghost. `lint-config-fallback.test.ts` names the contract actually being held.

Low value — only worth it if you're touching the file for finding 2 anyway.

Separately, credit where due: this suite is the strongest part of the PR. It runs the real `lint.ts` in a real `bun` subprocess and stubs only the process boundary (PATH), across both shipped copies. That's genuine wiring coverage, not a mock theatre — it's what let me verify the behavior claims independently instead of trusting the description.

packages/cli/tests/integration/invisible-extension.test.ts
```diff
@@ -340,9 +340,9 @@ formatters:
         expect(lintHook).toContain('.safeword/ruff.toml');
         expect(lintHook).toContain('.safeword/.golangci.yml');
 
-        // Should use --config flag when safeword configs exist
+        // The generated hook carries explicit config support. Runtime selection
+        // with and without these files is covered by lint-no-auto-upgrade.test.ts.
```
TheMostlyGreat: **Finding 4 — this trade is correct, no action.**

Dropping `expect(lintHook).toContain('hasEslint')` removes a string-match on an internal identifier, which was always brittle — it would break on any rename that changed nothing semantically. Pointing at the runtime matrix instead is the right call, and that matrix is strictly stronger coverage than the string match ever was.

Noting it only so the deletion doesn't look like lost coverage in review.

TheMostlyGreat (commented):
Line-anchored version of the findings from my two comments above, so they show up in Files changed next to the code they're about. Same four items, same verdict: approve, nothing blocking.

One mechanical note — this is your own PR, so GitHub won't let me file this as an APPROVE event. It's a COMMENT review; read the verdict as approve.

packages/cli/templates/hooks/lib/lint.ts
TheMostlyGreat: Addressed in d41027f51. Missing TypeScript, Python, Go, and Rust configs now produce a once-per-session warning that fallback defaults—not Safeword rules—are active. The header now distinguishes session-start version upgrades from manual pack repair.

packages/cli/tests/hooks/lint-config-fallback.test.ts
TheMostlyGreat: Addressed in d41027f51. Missing-config SQL now asserts the invocation list is exactly empty instead of relying on vacuous `arrayContaining([])`.

packages/website/src/content/docs/reference/configuration.mdx
TheMostlyGreat: Addressed in d41027f51. The docs now state that pack repair is manual and point to `safeword setup` / `safeword upgrade`; session start is described only as the owner of version upgrades.

packages/cli/tests/hooks/lint-no-auto-upgrade.test.ts
TheMostlyGreat: Addressed alongside finding 1 in d41027f51. Missing-config cases now expect the warning on the first edit and an empty warning list on the second edit in the same process, explicitly pinning once-per-session deduplication.

packages/cli/tests/hooks/lint-config-fallback.test.ts
TheMostlyGreat: Addressed in d41027f51. Renamed the suite to `lint-config-fallback.test.ts` and updated its cross-reference.

packages/cli/tests/integration/invisible-extension.test.ts
TheMostlyGreat: Agreed—no code change needed. The runtime matrix remains the behavioral coverage for this path.

Issue #1479, opened 2026-07-26
Guarantee retry-safe retro filing without duplicate issues


TheMostlyGreat:
## Problem

Retro filing currently asks each agent/runtime to search GitHub, infer whether a marker already exists, create an issue when it appears absent, and acknowledge its local spool. That cannot provide a strong invariant across:

- relevance-ranked or capped searches;
- hidden-marker indexing behavior;
- simultaneous filers;
- ambiguous create responses and network retries;
- ephemeral runtimes and per-session spools.

PR #1465 makes the immediate policy choice explicit, but neither filing anyway nor suppressing on uncertainty can guarantee both delivery and uniqueness. Downstream duplicate reconciliation also occurs after notifications and other issue-created side effects have already fired.

GitHub exposes no documented conditional-create/idempotency or strong-consistency contract for issue creation. That absence — not any specific observed replication behavior — is what makes an ambiguous create impossible to resolve deterministically from the client.

## User value

A retro finding can be retried from any supported harness without being lost or filed twice **for the same request**. See Non-goals: this does not eliminate duplicates arising from semantic identity drift.

## Proposed direction

Introduce one dedicated idempotent filing boundary, exposed to every harness as a single operation such as:

```text
fileRetroDraft({
  requestId,          // persisted once, before the first delivery attempt
  repository,
  canonicalKey,       // versioned semantic-dedupe evidence, NOT the request token
  legacySignature,    // versioned semantic-dedupe evidence, NOT the request token
  title,
  body,
  labels
}) -> { issueNumber, disposition }
```

The boundary owns durable request state and GitHub creation. It records enough state to recover safely when the GitHub response is ambiguous.

**Request identity vs. semantic identity are separate concerns.** `requestId` is the idempotency token: generated once when the draft is first persisted and reused verbatim for every retry. `canonicalKey` and `legacySignature` are versioned semantic-dedupe *evidence* — useful for matching, but demonstrably unstable (see Non-goals), so they must not serve as the request token.

### Identity namespace

Uniqueness is scoped by **tenant / installation / repository + `requestId`** — a stable identity that is **independent of transport**.

It must **not** be scoped by the harness actor (Claude, Cursor, Codex). A draft persisted in one harness and retried through another is the *same* request; a harness-scoped namespace would place the retry in a different bucket and file a second issue while appearing to satisfy every invariant below. The authenticated harness principal is for **authorization and audit only**, never for dedupe.

[...]

PR #1522, opened 2026-07-27
Deliver retry-safe retro findings across every harness

TheMostlyGreat:
author —

Refs #1479.

Implements the smallest bounded end-to-end relay slice as one unit because the client spool, shared request identity, authenticated HTTP boundary, encrypted SQLite lifecycle, raw-REST reconciliation, operator recovery, and real-collaborator wiring are one safety property: splitting them would leave a path that can accept a finding without proving durable, duplicate-safe recovery.

The slice provides:

- one persisted UUIDv4 request identity across Claude, Codex, and Cursor, including all six installed local/cloud surfaces;
- an atomic durable source reservation so concurrent or corrupt local state cannot silently re-identify a finding;
- a 750 ms aggregate client budget and one absolute retry deadline, with visible local/server dead letters and explicit operator recovery under the original marker;
- independently rotatable harness/operator credentials constrained to one tenant, installation, and repository with exact role authorization;
- encrypted SQLite WAL lifecycle state, key rotation, bounded retries, 30-day terminal payload compaction, and indefinite non-reusable identity tombstones;
- raw GitHub REST bodies as the only request-marker authority; sanitized MCP reads never participate in duplicate decisions;
- complete ascending pagination, serialized reconciliation/recovery, typed GitHub outcomes, rate/resource limits, durable alerts, and production Docker/Railway volume wiring.

Production routing remains fail-closed. #1474 and #1481 are closed and present on main, but semantic cross-request uniqueness stays disabled until fresh post-fix collision measurements and evidence review satisfy the checked-in readiness manifest. This PR does not supersede #834. #1495 is a readiness gate only if its client credential helpers are reused; this slice does not reuse them.

Verification:

- relay: 138 passed, 1 ordinary-lane environment skip; CI-mode production Docker qualification passed 6/6;
- CLI aggregate: 5,622 passed and 5 skipped; eight subprocess timeouts in three legacy integration files passed 19/19 on immediate isolated rerun;
- BDD: 623 scenarios (620 passed, 3 skipped), 19,861 steps (19,857 passed, 4 skipped);
- release gate 22/22; non-git physical install proof 1/1; schema-drift push gate 693/693;
- build, full lint, typecheck, format, dependency validation, diff hygiene, and production audit passed; no vulnerabilities found;
- fresh independent quality review and final delta review: APPROVE, no critical issues.

## Deployment automation (#1748)

This branch now includes `c4b5ca7f7`, which adds an auditable GitHub Actions Railway deployment path for the private relay. It runs only for relay build inputs on `main` or manual dispatch, serializes production deployments, uses read-only GitHub permissions, targets the existing project/environment/service through repository variables, and validates configuration before `railway up --ci`. The project-scoped `RAILWAY_TOKEN` repository secret is intentionally required and remains the only gate before a real production deployment.

The three non-secret target variables are configured. Runtime credentials and the Railway token are not committed or logged.


TheMostlyGreat (commented):
Reviewer: quality + engineering review at `d7b6249`. Draft, so read this as in-progress rather than a regression.

**Headline: CI is red, and the failing test is the one that proves this PR's central claim.** The body says *"relay 110/110"*; actual is **109/110**, failing identically on Node 22 and Node 24 in [run 30245439363](https://github.com/ArcadeAI/safeword/actions/runs/30245439363). Root cause and one-line fix are inline on `cli-wiring.integration.test.ts`.

Everything else here is a suggestion. None of the four block.

## Confirmed sound

Most of this PR is solid, and I want that on the record before the findings:

- **Crypto** — `payload.ts` is textbook AEAD: AES-256-GCM, random 12-byte nonce, AAD binding tenant/installation/repo/requestId/payload-hash, key length checked on encrypt. `identity.ts` uses length-prefixed field encoding, closing the concatenation-collision hole naive `join()` hashing leaves open.
- **GitHub App auth** — RS256; `iat: now-60` for clock drift; `exp: now+9min`; the token request is scoped to a single repo with `issues: write` only, and the returned permissions are *verified* before caching. Timeout via unref'd `AbortController`, in-flight dedup, 5-minute refresh margin. Errors carry the status, never the token.
- **Credentials** — HMAC-SHA256 with pepper plus `timingSafeEqual` over equal-length digests. The 256-bit hex secret makes a fast HMAC the right call over a slow KDF.
- **HTTP surface** — 256 KB body cap with a content-length precheck, plus 401/403/429 paths.
- **The idempotency coupling holds.** `payloadHash` includes `retryDeadlineAt` and `submit` 409s on a hash mismatch, so a client recomputing the deadline per attempt would break idempotency. `createRelayRequest` computes it once, persists the whole request to `.safeword/retro-drafts/relay/<id>.json`, and retries re-read that file. Stable across retries, as designed.
- **Readiness is genuinely fail-closed** — `CHECKED_IN_RELAY_READINESS = { enabled: false }`, and every validation branch returns `{ enabled: false }`. That claim in the body checks out.

One thing I chased and cleared: I expected the v1/v2 migration backfill `strftime('%Y-%m-%dT%H:%M:%fZ', accepted_at, '+24 hours')` to choke on the `Z` suffix that `toISOString()` emits. It doesn't — SQLite 3.53.0 returns `2026-07-28T14:23:17.477Z` correctly. Backfill is clean.

## Also worth doing

Correct the "relay 110/110" line in the verification ledger once CI is actually green — it currently reads as verified when it isn't.

packages/retro-relay/tests/cli-wiring.integration.test.ts
```diff
@@ -0,0 +1,492 @@
...
+  sessionId: string,
+): string[] {
+  const capture = path.join(project, `${sessionId}.json`);
+  const executable = path.join(project, 'capture-retro-args.mjs');
+  writeFileSync(
+    executable,
+    `#!/usr/bin/env node\nimport { writeFileSync } from 'node:fs';\nwriteFileSync(process.env.SAFEWORD_CAPTURE_PATH, JSON.stringify(process.argv.slice(2)));\n`,
+    { mode: 0o755 },
+  );
+  const bunExecutable = path.join(process.env.BUN_INSTALL ?? '/missing-bun-install', 'bin', 'bun');
```
TheMostlyGreat: Reviewer: **this is the CI failure.** Deterministic on both Node 22 and Node 24.

No workflow sets `BUN_INSTALL` — `oven-sh/setup-bun@v2` puts bun on `PATH` without exporting it — so CI spawns the literal `/missing-bun-install/bin/bun`. I reproduced the exact signature:

```
status=null | stderr=undefined | error=ENOENT
```

which is precisely the `AssertionError: expected null to be +0` two lines down. `BUN_INSTALL` is also unset in a plain non-interactive shell on macOS; it's set by the bun installer in an interactive profile, which is why this passes locally and nowhere else. Grep confirms the variable appears nowhere else in the repo (the `TIMEOUT_BUN_INSTALL` hits are an unrelated constant), so this is a one-off host assumption rather than an established pattern.

Bare-name PATH resolution works — I confirmed `spawnSync('bun', ['--version'])` returns status 0, bun 1.3.14:

```ts
const result = spawnSync('bun', [hook], { ... });
```

What makes this worth fixing before anything else: this test is the one that proves *"six fresh-runtime wiring"* and *"routes all six installed surfaces"* — the PR's headline claim is the single thing CI has never verified.

packages/retro-relay/tests/cli-wiring.integration.test.ts
```diff
@@ -0,0 +1,492 @@
...
+    encoding: 'utf8',
+    env: {
+      ...process.env,
+      CLAUDE_PROJECT_DIR: project,
+      SAFEWORD_CAPTURE_PATH: capture,
+      SAFEWORD_RETRO_EXTRACT_CMD: executable,
+    },
+    input: JSON.stringify({ session_id: sessionId, transcript_path: transcript }),
+  });
+  expect(result.status, result.stderr).toBe(0);
```
TheMostlyGreat: Reviewer: this assertion conceals its own cause. On ENOENT `result.stderr` is `undefined`, so the custom message is empty and the failure reads `expected null to be +0` with no hint at what went wrong — which is why diagnosing the line above took a CI log dig rather than a glance.

For a test whose whole job is spawning subprocesses, assert the spawn succeeded first:

```ts
expect(result.error, String(result.error)).toBeUndefined();
expect(result.status, result.stderr ?? '<no stderr>').toBe(0);
```

packages/retro-relay/src/service.ts
```diff
@@ -0,0 +1,287 @@
...
+    'legacySignature',
+    'repository',
+    'requestId',
+    'retryDeadlineAt',
+    'title',
+  ];
+  if (
+    keys.join('\0') !== expectedKeys.join('\0') ||
+    !validText(candidate.requestId, 36) ||
+    !/^[\da-f]{8}-[\da-f]{4}-4[\da-f]{3}-[89ab][\da-f]{3}-[\da-f]{12}$/iu.test(
```
TheMostlyGreat: Reviewer: `requestId` is matched case-insensitively (`/iu`), but it is part of the identity key — the `retro_requests` primary key is `(tenant_id, installation_id, repository, request_id)` over the raw string.

Note the asymmetry: `repository` **is** normalized (`normalizeRepo` lowercases it), `requestId` is not. So `A1B2...` and `a1b2...` are the same UUID to a human and two distinct rows to the store — which means a duplicate issue, the exact outcome this whole slice exists to prevent.

Low likelihood in practice, since the client generates the id once and persists it. But it's the identity key and the inconsistency looks unintended. Either lowercase it when building the scope, or drop the `i` flag so the case is pinned at the boundary.

packages/retro-relay/src/store.ts
```diff
@@ -0,0 +1,781 @@
...
+  }
+  const migrate = database.transaction(() => {
+    database.exec(`
+      ALTER TABLE retro_requests ADD COLUMN dead_lettered_at TEXT;
+      ALTER TABLE retro_requests ADD COLUMN tombstoned_at TEXT;
+      ALTER TABLE retro_requests ADD COLUMN payload_compacted_at TEXT;
+      ALTER TABLE retro_requests ADD COLUMN next_attempt_at TEXT;
+      ALTER TABLE retro_requests ADD COLUMN attempt_count INTEGER NOT NULL DEFAULT 0;
+      ALTER TABLE retro_requests ADD COLUMN dispatch_started_at TEXT;
+      ALTER TABLE retro_requests ADD COLUMN retry_deadline_at TEXT;
```
TheMostlyGreat: Reviewer: migrated and freshly-created schemas diverge here, and `validateVersionThree` can't see it.

`createVersionThree` declares `retry_deadline_at TEXT NOT NULL` (store.ts:180). Both migrations add it via `ALTER TABLE ADD COLUMN`, which SQLite cannot make `NOT NULL` without a default — so a migrated database has a nullable column where a fresh one has a constrained one. `validateVersionThree` compares column *names* only, so both pass identically.

Harmless today: the backfill fills every existing row, and every insert path supplies the value. The reason it's worth closing anyway is the failure mode if it ever were NULL — `claim()` evaluates

```sql
julianday(?) < julianday(retry_deadline_at)
```

which yields NULL against a NULL deadline, so the `UPDATE` matches nothing. The request is then **never claimable and never dead-lettered** — silently stuck, invisible to the alerting. I verified `julianday(x) < julianday(NULL)` returns `null`.

Cheapest fix is a table rebuild in the migration, or have `validateVersionThree` assert the constraint via `PRAGMA table_info` `notnull` rather than just the name set.

packages/retro-relay/src/service.ts
```diff
@@ -0,0 +1,287 @@
...
+    }
+    try {
+      const issueNumber = await this.#github.createIssue({
+        repository: scope.repository,
+        title: durableRequest.title,
+        body: `${durableRequest.body}\n\n${record.requestMarker}`,
+        labels: durableRequest.labels,
+        installationToken,
+      });
+      this.faults.afterGitHubCreate?.();
```
TheMostlyGreat: Reviewer: test fault-injection is shipping on the production class. `RelayService.faults` is a public field holding a mutable `RelayFaults` (service.ts:15, :108), and this line calls into it on the live dispatch path.

The need is legitimate — simulating a crash between the GitHub create and the receipt commit is exactly how you prove the ambiguous-reconciliation path, and I'd want that test. It's the *surface* that's the issue: a mutable public hook on the production service means anything holding the instance can install a callback into the dispatch path.

Constructor-only injection (drop it to a `#private` field) or a test-only subclass keeps the capability without the reachable seam.

packages/retro-relay/package.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "scripts": {
+    "build": "tsup",
+    "spike": "bun scripts/railway-spike.ts",
+    "start": "node dist/main.js",
+    "test": "vitest run",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "better-sqlite3": "13.0.1"
```
TheMostlyGreat: Reviewer: `better-sqlite3` is a native dependency in a Bun-first repo. `bun install --frozen-lockfile` fails here on darwin/arm64 with a node-gyp error — no matching prebuild — so the new package can't be built or tested locally on an Apple-silicon machine without extra toolchain setup. CI is Linux and installs fine, which is why this wouldn't show up in the pipeline.

Worth checking whether `node:sqlite` removes the native dep outright: it's built into every Node version in this package's `engines` range, and the usage here (sync API, `.transaction()`, `pragma()`) maps closely.

Flagging rather than insisting — better-sqlite3's ergonomics are genuinely better, and "contributors need a build toolchain" may be an acceptable trade. But it should be a decision rather than a default.

packages/retro-relay/tests/cli-wiring.integration.test.ts
TheMostlyGreat: author — Fixed in 795559b97. The wiring test now resolves `bun` through PATH, and the regression passes with `BUN_INSTALL` explicitly unset.

packages/retro-relay/tests/cli-wiring.integration.test.ts
TheMostlyGreat: author — Fixed in 795559b97. The test asserts `result.error` first, then reports stderr (or an explicit fallback) for a nonzero status.

packages/retro-relay/src/service.ts
TheMostlyGreat: author — Fixed in 795559b97. The HTTP boundary now accepts only canonical lowercase UUIDv4 request identities; an uppercase-identity regression proves rejection occurs before GitHub dispatch.

packages/retro-relay/src/store.ts
TheMostlyGreat: author — Fixed in 795559b97. The v1/v2 migrations rebuild the request table with `retry_deadline_at TEXT NOT NULL`, preserve FK-linked audit/outbox rows transactionally, validate the constraint, and assert an empty `foreign_key_check`.

packages/retro-relay/src/service.ts
TheMostlyGreat: author — Fixed in 795559b97. Fault hooks are constructor-only, cloned into a private field, and no mutable fault surface is returned by the server; the crash/recovery tests inject one-shot faults at construction.

packages/retro-relay/package.json
TheMostlyGreat: author — Fixed in 795559b97. The relay now uses the qualified Node runtime’s built-in `node:sqlite`; better-sqlite3, its types, and the Docker compiler toolchain are removed. A built-dist smoke test and the runtime image both pass.

packages/retro-relay/src/sqlite.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  immediateTransaction<Result>(operation: () => Result): Result {
+    this.#database.exec('BEGIN IMMEDIATE;');
+    try {
+      const result = operation();
+      this.#database.exec('COMMIT;');
+      return result;
+    } catch (error) {
+      this.#database.exec('ROLLBACK;');
```
TheMostlyGreat: Reviewer: this unguarded `ROLLBACK` swallows the error you actually need.

SQLite auto-rolls-back on `SQLITE_FULL`, `SQLITE_IOERR`, and `SQLITE_NOMEM`. When it already has, this `exec('ROLLBACK;')` throws and **replaces** the original error. Reproduced against `node:sqlite` with a transaction that's no longer active:

```
surfaced error => cannot rollback - no transaction is active
```

The real error is gone. `better-sqlite3`'s `.transaction()` guarded this by checking `inTransaction` before rolling back; the swap in #6 dropped that guard, so this is a regression the driver change introduced rather than anything pre-existing.

It bites hardest exactly where you least want it: a disk-full on the Railway volume mid-migration would surface as a confusing rollback error instead of `SQLITE_FULL`. Same class as comment #2 last round — an error path concealing its own cause.

```ts
} catch (error) {
  try { this.#database.exec('ROLLBACK;'); } catch { /* already rolled back */ }
  throw error;
}
```

packages/retro-relay/src/sqlite.ts
```diff
@@ -0,0 +1,96 @@
...
+import type {
+  DatabaseSync as NodeDatabaseSync,
+  SQLInputValue,
+  StatementResultingChanges,
+  StatementSync,
+} from 'node:sqlite';
+
+// tsup strips `node:` from static built-in imports, but SQLite has no bare
+// `sqlite` alias. Keeping the specifier in require() preserves the runtime API.
+const { DatabaseSync } = createRequire(import.meta.url)('node:sqlite') as {
```
TheMostlyGreat, line 11: Reviewer: question rather than a finding — `tsup.config.ts` gained `platform: 'node'` in this batch, which should make esbuild preserve `node:` prefixes on built-in imports. That's precisely the problem this `createRequire` hack (and the comment above it) exists to work around.

Is it still needed, or can this become a plain static import now? I couldn't test it here — tsup won't install in my worktree — but `runtime-qualification.test.ts` loads the built `dist/index.js` on the real runtime, so it would catch a regression immediately if you tried it and the prefix got stripped after all.

Worth 30 seconds to check; if it works it removes a workaround and its explanatory comment.

packages/retro-relay/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@safeword/retro-relay",
+  "version": "0.1.0",
+  "private": true,
+  "type": "module",
+  "engines": {
+    "node": "^22.23.1 || ^24.18.0 || >=26.3.0"
```
TheMostlyGreat: Reviewer: note, not a blocker — `node:sqlite` is not stable yet. Node's docs give it **"Stability: 1.2 - Release candidate"**; it came out from behind `--experimental-sqlite` in 22.13.0/23.4.0 but stayed *experimental* until 25.7.0.

Empirically, on this machine:

- Node 22.22.2 -> `ExperimentalWarning: SQLite is an experimental feature and might change at any time` on import
- Node 24.16.0 -> silent

Since this range starts at `^22.23.1`, a chunk of the supported matrix writes that warning to stderr on every relay boot. Worth a line in the README so nobody debugging a production start is surprised by it — and worth checking nothing downstream parses relay stderr.

To be clear I'd still take this trade over the native dependency, and `sqlite.ts` is the correct mitigation: the API surface you depend on is ~40 lines behind one seam, so an RC-stage breaking change is a contained edit rather than a scattered one. Just worth knowing the durability layer sits on a release candidate.

TheMostlyGreat (commented):
Reviewer: second pass at `6118cde`. **All six comments addressed, CI now 4/4 green** (was 109/110 red). One new finding introduced by the driver swap, plus a question and a note — inline.

| # | Finding | Fix |
| --- | --- | --- |
| 1 | `BUN_INSTALL` -> ENOENT in CI | `spawnSync('bun', [hook])`, PATH resolution |
| 2 | Assertion concealed its cause | `expect(result.error).toBeUndefined()` before the status check |
| 3 | `requestId` case-insensitive on the identity key | `/iu` -> `/u` |
| 4 | Migrated vs fresh schema drift | Full table rebuild **and** a `notnull` assertion |
| 5 | Mutable public `faults` | `#faults` private, spread-copied at construction |
| 6 | `better-sqlite3` native dep | Removed — `dependencies: {}`, `node:sqlite` adapter |

Two went further than asked. **#4** does both remedies — the rebuild copies through a constrained `retro_requests_v3` under `defer_foreign_keys`, *and* `validateVersionThree` now fails closed on a missing `NOT NULL`. **#5**'s split is better than what I suggested: faults are passed *in* through options and divided by ownership (`afterGitHubCreate` to the service, `afterReceiptCommit` to the transport) rather than handed back as a mutable handle.

**#6 paid off in the Dockerfile** — the whole native-build stanza is gone (apt-get `g++ make python3`, `npm rebuild --build-from-source`, the prebuild deletion, the purge). The runtime image now carries no build toolchain and no `node_modules` at all, just the copied `dist`.

## Worth naming

`runtime-qualification.test.ts` is the right de-risking for a driver swap: it spawns `process.execPath` to import the **built** `dist/index.js`, then opens a real store, asserts WAL, migrates, closes, and reopens. That covers the `node:` stripping workaround, the adapter, and the migration on the actual runtime in one test — and it already applies the `result.error`-before-`result.status` pattern from comment #2, proactively.

**Verdict: APPROVE** once the rollback guard lands. Nothing else blocks; still a draft.

packages/retro-relay/src/service.ts
```diff
@@ -0,0 +1,286 @@
...
+    });
+    if (!scan.complete) {
+      this.#store.recordReconciliation(receiptId, principal.subject, 'incomplete', 0);
+      throw new RelayError(503, 'raw reconciliation is incomplete or non-unique', {
+        receiptId,
+        state: 'ambiguous',
+        disposition: 'incomplete',
+      });
+    }
+    if (scan.issueNumbers.length !== 1) {
```
TheMostlyGreat: **Critical — zero-match `ambiguous` is terminal, and the spec's manual recovery path doesn't exist.**

When `scan.complete` is true and `issueNumbers.length === 0`, GitHub has *proven* nothing was created. This branch records disposition `zero`, throws 503, and leaves the record `ambiguous`. The routes in `http-server.ts` are submit / reconcile / status / operations / health — there is no operation that can create with the original reserved marker.

#1479 promises one: *"Manual zero-match recovery may create only with the original reserved marker and requestId."*

So a GitHub 5xx or a timeout where **nothing was created** strands that finding permanently, and a complete scan proving non-creation doesn't help. `submit` re-entry returns 503 on the same `ambiguous` record, and no fresh identity may be minted by design. That's the toil path from the spec review, now a dead end rather than merely expensive.

The conservatism itself is defensible — GitHub documents no read-your-writes guarantee, so a fresh zero-match could in principle be an in-flight create. But that argues for a bounded re-scan before recovery, not for having no exit at all. Suggest a `reconcile` variant restricted to the `operate` role that re-creates using the original `requestMarker` after N consecutive complete zero-match scans spanning more than a stated propagation window. Re-creating with the *same* marker keeps it safe: if the original ever appears, the next scan sees two markers and quarantines as `multiple` rather than silently duplicating.

packages/retro-relay/src/payload.ts
```diff
@@ -0,0 +1,56 @@
+import { createCipheriv, createDecipheriv, randomBytes } from 'node:crypto';
+
+import type { FileRetroDraftRequest, RequestScope } from './types.js';
+
+export interface PayloadEnvelope {
```
TheMostlyGreat, line 5: **Critical — no key version, so rotating `RELAY_PAYLOAD_KEY` destroys queued findings.**

`PayloadEnvelope` carries `{nonce, ciphertext, tag}` and the AAD in `associatedData` carries no key reference. The key arrives as an env var on a PaaS, so it *will* rotate.

When it does, every un-filed envelope becomes undecryptable. `#processClaimed` catches the `decryptPayload` failure in its preparation `try` and calls `markRetryable`, so those records churn against the new key until the retry deadline and then dead-letter. Rotation silently destroys queued findings, and the failure looks like a transient GitHub problem rather than a key problem.

Suggest adding a `keyId` to the envelope and to the AAD, and accepting a keyring rather than a single `payloadKey` — decrypt by the envelope's `keyId`, encrypt with the current one. That also makes rotation a normal operation instead of a migration.

The construction otherwise looks right: 12-byte random nonce per envelope, GCM tag stored separately, and the AAD binding scope + payload hash so an envelope can't be replayed under a different identity.

packages/cli/src/retro/relay-delivery.ts
```diff
@@ -0,0 +1,440 @@
...
+    fetch: typeof fetch;
+    now: () => number;
+    relayUrl: string;
+  },
+): Promise<{ accepted: number; deadLettered: number; retryable: number }> {
+  const initial = await listRelayRequests(projectDirectory);
+  const processed = new Set<string>();
+  let accepted = 0;
+  const initialDeadLetters = await listRelayDeadLetters(projectDirectory);
+  let deadLettered = initialDeadLetters.length;
```
TheMostlyGreat: **`agentFilingNeeded` becomes permanently sticky.**

`deadLettered` is seeded with the count of *pre-existing* dead letters before the loop increments it for newly dead-lettered drafts. In `retro.ts` the return feeds `agentFilingNeeded: delivery.retryable > 0 || delivery.deadLettered > 0`, and dead-letter files are never removed — `recoverRelaySpool` only clears primaries and claims whose request is acknowledged.

So a single old dead letter forces `agentFilingNeeded: true` on every subsequent run forever, and the reported count conflates the standing backlog with this run's outcome. The signal that should mean "this run needs agent fallback" degrades into "this project once had a failure."

Suggest returning the two separately — `deadLettered` for newly dead-lettered this run, and a `deadLetterBacklog` for the standing count — and keying `agentFilingNeeded` on the former.

packages/cli/src/retro/relay-delivery.ts
```diff
@@ -0,0 +1,440 @@
...
+  let retryable = 0;
+  for (const request of initial) {
```
TheMostlyGreat: **No aggregate latency budget — the per-request deadline doesn't bound the session.**

`deadlineMs` defaults to a tight 750ms per request, but this loop runs over every spooled draft with no ceiling. If the relay black-holes rather than refusing connections, a batch of N drafts costs N × 750ms, and this sits in the session's path via the Stop hook.

`claimRelayRequest` compounds it: each iteration re-runs `recoverRelaySpool` and re-lists the directory, so the filesystem work is O(n²) in spooled drafts.

The #1479 review asked for a hard latency budget so filing can never delay a session. The per-request half is here; the aggregate half isn't. Suggest an overall deadline checked at the top of each iteration — break out and leave the remainder spooled once the budget is spent, since the drafts are durable and the next run picks them up.

packages/retro-relay/src/github.ts
```diff
@@ -0,0 +1,233 @@
...
+
+export class GitHubCreateError extends Error {
+  readonly outcome: 'ambiguous' | 'rejected' | 'retryable';
+  readonly status: number;
+
+  constructor(status: number) {
+    super(`GitHub create failed with ${status}`);
+    this.status = status;
+    if ([400, 404, 410].includes(status)) this.outcome = 'rejected';
+    else if ([401, 403, 422, 429].includes(status)) this.outcome = 'retryable';
```
TheMostlyGreat: **Minor — 422 and 403 conflate fail-fast cases with retryable ones.**

GitHub documents 422 on create-issue as "Validation failed, **or** the endpoint has been spammed." Only the second half is retryable. A genuine validation failure now retries against an identical payload until the 24-hour deadline and then dead-letters, instead of failing fast as `rejected` with a usable error.

403 has the same shape: secondary rate limiting (retryable, and the documented risk here since retro files in batches) versus a real permission denial (never going to succeed).

Not blocking — the outcome is a delayed dead letter rather than a wrong one. But splitting on the response body's message would turn a 24-hour silence into an immediate, actionable failure.

TheMostlyGreat (commented):
Reviewer: **verdict — REQUEST CHANGES** (submitted as a comment review; GitHub blocks a formal request-changes on your own PR).

`/quality-review` + engineering pass on the relay slice. CI is green (4/4), the wiring tests drive real collaborators and mock only the GitHub boundary, and the construction is careful — length-prefixed hash encoding, AAD-bound envelopes, filesystem CAS via `link()`. Two defects strand data permanently; four inline comments below, ordered by severity.

**First, what landed from the #1479 review.** Four of five blockers are resolved, two better than asked:

- **C4** — `validateRequest` enforces UUIDv4 including variant bits, server-side, and the public marker is a SHA-256 over `(tenant, installation, repo, requestId)` so the raw id never reaches a public body.
- **C3** — `createRelayRequest` stamps `retryDeadlineAt` once; retries replay persisted bytes rather than re-rendering; `persistRelayDraft` dedupes by `sourceKey`. Exactly the proposed remedy.
- **C2 (classification)** — `GitHubCreateError` now separates rejected / retryable / ambiguous outcomes.
- **C1** — encoded as a machine-checked fail-closed gate in `relay-readiness.ts`: prerequisites closed, merge commits ancestors of the evidence commit, measurement artifacts SHA-256-verified at that commit, measured after the last close, under 30 days old. Stronger than the checklist it answers.

**Merge risk is low regardless** — the checked-in manifest ships `enabled: false`, so nothing routes in production until the prerequisites land.

**Minor, not worth inline comments:** `payloadHash` uses `payload-v2` while `associatedData` uses `payload-v1` (harmless now, confusing at the next migration); `recoverRelaySpool(projectDirectory, _now)` takes an unused parameter; repo identity is the normalized `owner/repo` string rather than the node id, so a rename re-buckets the same `requestId` (low odds for a fixed upstream repo, but the code is written multi-tenant).

Everything above was verified by reading this branch; the GitHub status-code semantics come from the REST docs fetched during the review.

packages/retro-relay/src/sqlite.ts
TheMostlyGreat: author — Fixed in 39c5aeaa6.  now checks  before rollback, preserves the original operation error when SQLite already auto-rolled back, and surfaces both errors via  if rollback itself fails. The regression test explicitly rolls back inside the operation and confirms the sentinel error remains authoritative.

packages/retro-relay/src/service.ts
TheMostlyGreat: author — Fixed in 39c5aeaa6. There is now an operator-only  path. It serializes recovery with a durable CAS, performs a fresh complete raw REST marker scan, adopts exactly one match, refuses incomplete or multiple results, and creates on zero only from the encrypted durable request with the original reserved marker/request identity. Real HTTP, GitHub-fixture, restart, role, audit, and simultaneous-recovery tests cover the path.

packages/retro-relay/package.json
TheMostlyGreat: author — Documented in 39c5aeaa6. The README now calls out the Node 22 experimental warning / Node 24 release-candidate maturity and explains that the adapter keeps the unstable API behind one seam.

packages/cli/src/retro/relay-delivery.ts
TheMostlyGreat: author — Fixed in 39c5aeaa6. Delivery now returns  for transitions made in this run and  for standing inventory.  depends only on this run’s retryable/new-dead-letter outcomes, so historical dead letters no longer make fallback permanently sticky.

packages/retro-relay/src/payload.ts
TheMostlyGreat: author — Fixed in 39c5aeaa6. Payload envelopes now carry format version and key ID in authenticated metadata; new writes use the active key and old queued rows decrypt by retained key ID. Schema v3→v4 migration labels legacy rows explicitly, startup fails fast with the exact missing key IDs, and the README documents that the first post-v3 keyring must retain the former key as . Rotation, legacy-AAD, migration, restart, and missing-key tests are green.

packages/retro-relay/src/github.ts
TheMostlyGreat: author — Fixed in 39c5aeaa6. Create failures now inspect GitHub’s response message plus  and : permission 403 and validation 422 reject immediately, while rate/abuse 403 and spam-style 422 remain retryable. Integration tests cover all four cases.

packages/cli/src/retro/relay-delivery.ts
TheMostlyGreat: author — Fixed in 39c5aeaa6. The drain has one monotonic aggregate deadline beginning before spool preparation; each request receives only the smaller of its per-request allowance and the remaining aggregate budget, and untouched requests remain durable for the next run. Expired-claim recovery and directory enumeration happen once before request-specific atomic claims, removing the repeated O(n²) scan. The three-draft fault test proves only work fitting the total budget is attempted.

packages/retro-relay/src/sqlite.ts
TheMostlyGreat: author — Tested against the repository build and runtime qualification path. Static DatabaseSync import from node:sqlite was rewritten to bare sqlite by this tsup artifact and failed with ERR_MODULE_NOT_FOUND. The createRequire node:sqlite seam therefore remains necessary and is covered by the built-runtime test.

packages/retro-relay/src/sqlite.ts
TheMostlyGreat: author — Correction: the guarded APIs are immediateTransaction and DatabaseSync.isTransaction; AggregateError carries both the operation and rollback failures.

packages/cli/src/retro/relay-delivery.ts
TheMostlyGreat: author — Correction: the separate fields are deadLettered and deadLetterBacklog, and agentFilingNeeded uses only current-run outcomes.

packages/retro-relay/src/service.ts
TheMostlyGreat: author — Correction: the new operator-only endpoint is POST /v1/retro-filings/:receiptId/recover.

packages/retro-relay/src/payload.ts
TheMostlyGreat: author — Correction: migrated schema-v3 envelopes require the former key under key ID legacy in the first configured keyring.

packages/retro-relay/src/github.ts
TheMostlyGreat: author — Correction: the rate-limit headers inspected are Retry-After and X-RateLimit-Remaining.

PR #1541, opened 2026-07-27
Prevent retro duplicates when issues close during pagination

TheMostlyGreat:
## Summary

- enumerate the all-state GitHub issue universe in creation order so close/reopen transitions cannot shift page membership
- filter closed issues and pull requests locally, preserving the existing eligible-match policy
- remove the repeated-sweep confirmation state machine and cover the page-boundary regression through the real REST transport
- raise the fail-closed dedup guard to 20,000 repository items, providing about 415 days of measured headroom at the trailing-seven-day growth rate
- make issue state explicit in every REST fixture

Fixes #1481

Follow-up: #1552 tracks Link-header traversal, safe cross-run caching, mutation lifecycle, equal-timestamp ordering, and guard observability.

## Verification

- 5,549/5,549 Vitest tests passed (5 skipped)
- 505/508 acceptance scenarios passed (3 skipped); 15,645/15,645 executed steps passed
- build, ESLint, Prettier, TypeScript, and diff hygiene passed
- focused REST transport suite: 41/41 passed
- audit: 0 change-scoped errors; dependency boundaries, dead code, architecture, docs, learning/domain docs, and changed-test quality clean

## Quality and engineering review

The review feedback and its corrections, evidence, measured cap decision, and post-fix **APPROVE** verdict are recorded in the durable [quality review artifact](https://github.com/ArcadeAI/safeword/blob/codex/1481-stable-retro-pagination/.project/tickets/GS2FGC-keep-retro-dedup-stable-during-issue-closure/quality-review.md). The post-fix review found no critical issues.

The API contract was checked against GitHub primary documentation for [listing repository issues](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues), [REST pagination](https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api), and [REST API versions](https://docs.github.com/en/rest/about-the-rest-api/api-versions).

### DIFF
--- a/packages/cli/src/retro/github-rest.test.ts
+++ b/packages/cli/src/retro/github-rest.test.ts
@@ -8,6 +8,14 @@ interface MockResponse {
   json: () => unknown;
 }
 
+const GITHUB_PAGE_SIZE = 100;
+// Deliberately independent from the production value: an accidental cap change
+// must fail the boundary contract instead of updating its own expectation.
+const EXPECTED_DEDUP_PAGE_BOUND = 200;
+const EXPECTED_DEDUP_PROBE_PAGE = EXPECTED_DEDUP_PAGE_BOUND + 1;
+const EXPECTED_DEDUP_ITEM_BOUND = EXPECTED_DEDUP_PAGE_BOUND * GITHUB_PAGE_SIZE;
+const PAGE_BOUNDARY_FIXTURE_SIZE = 2 * GITHUB_PAGE_SIZE + 1;
+
 function mockFetch(responder: (url: string) => MockResponse): string[] {
   const calls: string[] = [];
   vi.stubGlobal(
@@ -247,33 +255,33 @@ describe('createRestTransport', () => {
     if (!transport) throw new Error('expected a transport');
 
     await expect(transport.searchBySignature('retro:abc123def456')).rejects.toThrow(/truncated/);
-    // 200 bound pages + the probe page, which was also full → a genuine tail.
-    expect(calls).toHaveLength(201);
+    // The configured bound pages + a full probe page → a genuine tail.
+    expect(calls).toHaveLength(EXPECTED_DEDUP_PROBE_PAGE);
   });
 
-  // The boundary the bound alone cannot distinguish: at exactly 200 full pages the
-  // enumeration is COMPLETE, and throwing there would halt every session's filing
-  // over a tail that does not exist.
+  // The boundary the bound alone cannot distinguish: at exactly the configured
+  // number of full pages the enumeration is COMPLETE, and throwing there would
+  // halt every session's filing over a tail that does not exist.
   it('#1453: completes rather than throwing at exactly the page bound', async () => {
-    const fullPage = Array.from({ length: 100 }, (_, i) => ({
+    const fullPage = Array.from({ length: GITHUB_PAGE_SIZE }, (_, i) => ({
       number: i,
       title: `t${i}`,
       body: 'no marker',
       state: 'open',
     }));
     const calls = mockFetch(url => ({
-      json: () => (url.endsWith('page=201') ? [] : fullPage),
+      json: () => (url.endsWith(`page=${EXPECTED_DEDUP_PROBE_PAGE}`) ? [] : fullPage),
     }));
     const transport = createRestTransport('tok');
     if (!transport) throw new Error('expected a transport');
 
     await expect(transport.searchBySignature('retro:abc123def456')).resolves.toEqual([]);
-    expect(calls).toHaveLength(201);
+    expect(calls).toHaveLength(EXPECTED_DEDUP_PROBE_PAGE);
   });
 
-  // The cap has to mean what it says. Appending the probe page instead of rejecting
-  // on it would silently accept 20,001–20,099 items under a "20,000" bound.
-  it('#1453: trips the cap at 20001 items rather than quietly accepting the probe', async () => {
+  // The cap has to mean what it says. Appending the probe page instead of
+  // rejecting it would silently accept items beyond the configured bound.
+  it('#1453: trips the cap at the first item past the bound', async () => {
     const fullPage = Array.from({ length: 100 }, (_, i) => ({
       number: i,
       title: `t${i}`,
@@ -283,15 +291,15 @@ describe('createRestTransport', () => {
     const calls = mockFetch(url => ({
       // Exactly one item past the bound — the smallest genuine tail there is.
       json: () =>
-        url.endsWith('page=201')
-          ? [{ number: 20_001, title: 'tail', body: 'x', state: 'open' }]
+        url.endsWith(`page=${EXPECTED_DEDUP_PROBE_PAGE}`)
+          ? [{ number: EXPECTED_DEDUP_ITEM_BOUND + 1, title: 'tail', body: 'x', state: 'open' }]
           : fullPage,
     }));
     const transport = createRestTransport('tok');
     if (!transport) throw new Error('expected a transport');
 
     await expect(transport.searchBySignature('retro:abc123def456')).rejects.toThrow(/truncated/);
-    expect(calls).toHaveLength(201);
+    expect(calls).toHaveLength(EXPECTED_DEDUP_PROBE_PAGE);
   });
 
   // Ascending creation order appends genuinely new issues to the last page,
@@ -326,7 +334,7 @@ describe('createRestTransport', () => {
 
   it('#1481: finds a still-open marker when an earlier issue closes at a page boundary', async () => {
     const marker = '<!-- safeword-retro-signature: retro:abc123def456 -->';
-    const settled = Array.from({ length: 201 }, (_, i) => ({
+    const settled = Array.from({ length: PAGE_BOUNDARY_FIXTURE_SIZE }, (_, i) => ({
       number: i + 1,
       title: `t${i + 1}`,
       body: i === 100 ? marker : 'no marker',
@@ -388,8 +396,8 @@ describe('createRestTransport', () => {
     expect(calls.length).toBeGreaterThan(1);
   });
 
-  // Truncation is deterministic: re-running it burns another 201 requests to reach
-  // the same answer. Only transient failures earn a retry.
+  // Truncation is deterministic: re-running it burns another full bounded sweep
+  // plus the probe to reach the same answer. Only transient failures earn a retry.
   it('#1453: does not re-run a terminal truncation for every later encounter', async () => {
     const fullPage = Array.from({ length: 100 }, (_, i) => ({
       number: i,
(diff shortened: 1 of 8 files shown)


TheMostlyGreat (commented):

**Verdict: REQUEST CHANGES** — the core change is right, but it inherits a safety bound that was sized for a universe 2.6x smaller, and that bound is now dated.

### The core change is the right call

Replacing the double-sweep confirmation with `state=all` swaps a *probabilistic detector* for a *structural elimination*. `confirmedSnapshot` could only notice a close it happened to catch between two sweeps; with `state=all`, closing never changes membership or creation-order position, so the skip cannot happen at all. Per-encounter cost drops from three sweeps to one, and the removed state machine leaves no orphaned references.

The regression test earns its keep. The `state=open` branch in its fixture is not decoration — revert `state=all` and #101 falls through the page boundary and the test goes red.

API contract checks out against current GitHub docs: `state=all`, `sort=created`, `direction=asc`, `per_page` max 100, and the PR/issue conflation the local filter handles is documented verbatim ("may return both issues and pull requests in the response. You can identify pull requests by the `pull_request` key"). Sources: [List repository issues](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28), [Using pagination in the REST API](https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api).

### Critical: `MAX_DEDUP_PAGES` no longer has real headroom

`MAX_DEDUP_PAGES = 30` was set in #1465, when the enumeration universe was `state=open`. Widening to `state=all` changed what that constant bounds; the constant did not move. Measured against this repo via the GitHub API today (2026-07-27):

| | items | pages | headroom |
|---|---|---|---|
| old universe (`state=open`) | 584 | 6 | 5.1x — and **self-limiting**, closing shrinks it |
| new universe (`state=all`) | **1548** | 16 | **1.94x — and monotonic, nothing ever leaves** |

Breakdown of the new universe: 578 open issues + 6 open PRs + 399 closed issues + 565 closed/merged PRs.

Burn rate: item #483 was created 2026-06-27 and #1548 exists today — 1065 items in 30 days (~35/day; the last 7 days ran ~52/day). Remaining headroom is 1452 items. **Extrapolating that rate, the cap trips in roughly 4-6 weeks.** The date is a projection off an accelerating rate and should be read as such; the 1548-of-3000 measurement and the monotonic-growth argument do not depend on it.

The failure mode is what makes this blocking rather than a nice-to-have. It is a wall, not a degradation:

- `listOpenIssues` throws through `latchTerminal`
- `currentSnapshot` rethrows that latched error for every subsequent encounter in the transport's life
- drafts stay spooled and accumulate

And because `UPSTREAM_REPO` is hardcoded to `ArcadeAI/safeword`, this is a single shared universe: every safeword installation's retro filing stops on the same day, each with a quietly growing spool.

The code comment already reasons about this correctly — "it needs real headroom above the repo's total issue/PR count". The missing step is checking what that count currently is.

**Suggested fix:** raise `MAX_DEDUP_PAGES` to ~150 (15,000 items, comfortably over a year at current burn) and record the measurement and its date in the comment so the next reader can re-check. This costs nothing on the normal path — a real sweep is still 16 pages. The bound only binds when the repo genuinely is that large, at which point the requests are what correctness costs.

### Also blocking: CI lint is red

The description states "build, ESLint, Prettier, TypeScript, and diff hygiene passed", but the `lint` job fails on the ticket-closure gate:

> `.project/tickets/GS2FGC-keep-retro-dedup-stable-during-issue-closure/ticket.md: verify.md changed but ticket.md still has status: in_progress`

Not a code-quality failure — the PR carries completion evidence while the ticket sits at `phase: verify`. Run `/verify` and let the done gate flip it, then include that `ticket.md` change here.

Related: the description's "Independent verdict: **APPROVE**, with no critical findings" has no corresponding review on this PR. Before this one, #1541 had zero reviews, zero issue comments, and zero inline comments, so that verdict is self-asserted in the description with no reviewable record behind it. Worth either linking the artifact it came from or dropping the claim.

### Non-blocking suggestions

- **Enumeration is O(repo lifetime), permanently.** 62% of the new universe (565 merged PRs + 399 closed issues) can never be an eligible match, and every session re-fetches all of it. Raising the cap buys time; a persisted high-water issue number, or ETag conditional requests, would make the steady state O(new items). Worth a follow-up ticket rather than this PR.
- **Prefer `Link: rel="next"` over page-count termination.** GitHub explicitly recommends the link header for traversal. It lets the API decide when the list ends and demotes the page cap to a pure runaway guard, and it retires the `MAX_DEDUP_PAGES + 1` probe.
- **`created_at` tie-breaking is unspecified.** At ~52 items/day in bursts, two items can share a timestamp; if `sort=created` does not tie-break stably, a boundary skip survives in miniature. GitHub almost certainly orders by id here, and this is far smaller than what the PR fixes — flagged only because I could not confirm it from the endpoint docs. Uncertain, not a blocker.
- One test-helper note inline.

### Scores

- Correctness: 3.5/5 — the fix is sound; the un-revisited bound is a real defect
- Security: 5/5
- Performance: 3/5 — 3 sweeps down to 1 is a genuine win, but a sweep is now 16 pages and grows without bound
- Maintainability: 4.5/5 — meaningfully simpler, comments carry the reasoning well

packages/cli/src/retro/github-rest.ts
```diff
@@ -28,17 +28,13 @@ const PER_PAGE = 100;
...
-// endpoint returns PRs alongside issues, and they are filtered after the fetch,
-// so open PRs consume this budget too). Deliberately looser than
+// Safety bound on the dedup enumeration (→ 3000 repository *items*: the stable
+// all-state listing includes closed issues and PRs, which are filtered after the
+// fetch, so both consume this budget). Deliberately looser than
 // MAX_ISSUE_PAGES: hitting this bound THROWS rather than truncating (see
 // listOpenIssues), so it halts filing entirely — it needs real headroom above
-// the repo's open count, not just enough for one sweep.
+// the repo's total issue/PR count, not just enough for one sweep.
 const MAX_DEDUP_PAGES = 30;
```
TheMostlyGreat: **Critical:** this constant was sized in #1465 for a `state=open` universe and has not been revisited now that the universe is `state=all`.

Measured against this repo today (2026-07-27): 578 open issues + 6 open PRs + 399 closed issues + 565 closed/merged PRs = **1548 items, 16 pages, against a 3000-item cap**. The old open-only universe was 584 items — and critically, it was *self-limiting*, since closing an issue removed it. The all-state universe only ever grows.

At the current burn rate (1065 items in the last 30 days, ~52/day over the last 7), the remaining 1452 items of headroom is consumed in roughly 4-6 weeks. That figure is an extrapolation off an accelerating rate; the 1548-of-3000 measurement is not.

When it trips, `listOpenIssues` throws through `latchTerminal` and `currentSnapshot` rethrows for the transport's whole life — retro filing halts and drafts accumulate in the spool. `UPSTREAM_REPO` is hardcoded, so this lands on every safeword installation at once.

Suggest `MAX_DEDUP_PAGES = 150` (15,000 items) plus the measurement and its date in the comment above, so the next reader can re-check rather than re-derive. It costs nothing on the normal path — a real sweep is still 16 pages.

packages/cli/src/retro/github-rest.ts
```diff
@@ -226,7 +232,7 @@ export function createRestTransport(token: string | undefined): IssueTracker | u
       // isolates this as a failed encounter and leaves the draft spooled, which is
       // recoverable. Silently returning [] would file the duplicate this replaces.
       throw latchTerminal(
-        `retro dedup: open items exceed ${MAX_DEDUP_PAGES * PER_PAGE}; enumeration truncated`,
+        `retro dedup: repository items exceed ${MAX_DEDUP_PAGES * PER_PAGE}; enumeration truncated`,
```
TheMostlyGreat, line 242: Message update is right. Worth noting for whoever reads this next: this throw is latched via `latchTerminal`, so it is not a per-encounter failure — `currentSnapshot` rethrows it for every subsequent encounter in the transport's life, and combined with the hardcoded `UPSTREAM_REPO` it halts retro filing for every installation once the repo crosses the cap.

That blast radius is exactly right as a design choice (silently returning `[]` would file the duplicate this module exists to prevent). It is also why the cap value itself needs real headroom — see the comment on `MAX_DEDUP_PAGES`.

packages/cli/src/retro/github-rest.test.ts
```diff
@@ -8,6 +8,19 @@ interface MockResponse {
   json: () => unknown;
 }
 
+function normalizeIssueStates(value: unknown): unknown {
```
TheMostlyGreat: Minor: this papers over the fixtures rather than fixing them. The real API always returns `state`, so a fixture that omits it is unrealistic — and this helper makes the unrealistic version read as `open` silently. A future fixture that drops `state` by mistake will look fine.

Two smaller things:

- The shape sniff (`'number' in item && 'title' in item`) fires on *every* array response in this file, including the reconcile `listIssues` fixtures. Harmless today since reconcile ignores `state`, but it is a wider blast radius than the name suggests.
- `{ state: 'open', ...item }` spreads defensively after a guard that already established `state` is absent, so the spread order carries no meaning.

Adding `state: 'open'` explicitly to the fixtures that need it would be a few more lines and would keep the fixtures honest about the API's actual shape.

packages/cli/src/retro/github-rest.ts
TheMostlyGreat: Resolved in ca7eeef51. I validated the cap against a live all-state census rather than adopting 150 pages verbatim: 1,550 current items, 311 created in the trailing 7 days. The new 200-page / 20,000-item fail-closed guard leaves about 415 days of headroom at that faster rate, while normal operation remains 16 requests. Tests pin completion at exactly 20,000 and rejection at 20,001. The dated calculation and post-fix APPROVE review are recorded in the ticket quality-review artifact.

packages/cli/src/retro/github-rest.ts
TheMostlyGreat: Addressed in ca7eeef51 and follow-up #1552. The immediate outage risk is removed with a measured 20,000-item guard. I kept Link-header traversal, safe ETag/persisted-page caching, reopen/deletion/transfer invalidation, equal-timestamp ordering, and guard observability together in #1552 because those require a wider state/lifecycle design and GitHub does not document snapshot consistency for this endpoint.

packages/cli/src/retro/github-rest.test.ts
TheMostlyGreat: Resolved in ca7eeef51. Removed normalizeIssueStates entirely and added explicit state: 'open' to every GitHub issue fixture (with the existing closed case remaining explicit). The focused REST transport suite passes 41/41, and the fresh whole-repo verification passes 5,549 tests plus 505/508 acceptance scenarios.

PR #1558, opened 2026-07-27
Fix mixed JS/TS architecture truncation

TheMostlyGreat:
## Summary

- Enumerate both immediate directories and loose JS/TS modules in mixed src/lib roots.
- Prefer directories on same-name collisions and fingerprint canonical paths so heals update references.
- Document the generated architecture ownership and preservation contract accurately.
- Add unit and real-CLI coverage for mixed roots, file-to-directory transitions, determinism, and installed-guide parity.

## Root cause

The JS source-root recognizer returned directories or files, never both. Fingerprints also hashed only module names, so a later file-to-directory transition could retain a stale code reference.

## Impact

Generated architecture docs now expose complete mixed JS/TS top-level structure and heal canonical references without ambiguity.

## Validation

- 5,562 tests passed; 5 skipped.
- 510 acceptance scenarios passed; 3 skipped.
- Build, typecheck, ESLint, Gherkin lint, Prettier, architecture freshness, and guide parity passed.
- Independent quality re-review: APPROVE, with no remaining suggestions.

Addresses #1551. Purpose seeding remains intentionally separate.


TheMostlyGreat (commented):
Same findings as my two summary comments above, now anchored inline so each one sits on the code it's about. Eight comments: 2 blocking, 4 worth fixing, 2 FYI.

Order of attack if you want one: the eligibility rule (`architecture-skeleton.ts`) is the only one that changes generated output for every downstream consumer, so decide it before landing — tightening it afterward churns everyone's doc a second time.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -183,13 +178,18 @@ function enumerateJsSourceRoot(
...
-      .toSorted(byNodeName);
+  const byName = new Map(
+    entries
+      .filter(entry => entry.isDirectory())
+      .map(entry => [
+        entry.name,
+        { name: entry.name, path: pathFor(entry.name), purpose: PURPOSE_PLACEHOLDER },
+      ]),
+  );
+  for (const node of jsFileNodes(entries, pathFor)) {
```
TheMostlyGreat: **Blocking — the union admits things that aren't modules.**

`jsFileNodes` filters via `isJsSourceModuleFile`, which excludes dotfiles, `*.d.ts` and `*.config.*` — but **not entry points and not tests**.

Two consequences:

1. **Entry points.** `src/index.ts` (a pure barrel — 7 `export … from` lines, no logic) and `src/cli.ts` now render as top-level architectural concepts. The Rust extractor already rejects exactly this shape: `RUST_ROOT_FILES = {lib.rs, main.rs}`, commented *"entry points, not modules"* (L29-32 of this file). JS has no analogue.

2. **Tests.** `jsModuleName`'s own docstring spells it out: `db.test.ts` → `db.test`. Any project with colocated `src/foo.test.ts` now gains a `### foo.test` section with a placeholder to fill in. Before this PR the file branch only ran on flat/test-only packages, which was #843's stated intent. This repo's `src/` has no root-level tests, so the dogfood can't surface it.

Suggested: an entry-point exclusion set mirroring `RUST_ROOT_FILES` (at minimum `index`; `main` likely; `cli` debatable — it's a real module surface here), plus `*.test.*`/`*.spec.*` excluded **inside `enumerateJsSourceRoot` only**. `topLevelJsModuleNodes` must keep listing test files or you regress #843's integration-tests-package case.

The design doc considers neither case, so I read this as an unexamined gap rather than an accepted trade-off — happy to be corrected.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -97,11 +97,9 @@ export function extractSkeleton(projectDirectory: string): Skeleton {
...
 
-  // The `src/` layout (TS/JS) is authoritative: its child DIRECTORIES are the
-  // modules, unchanged — a package that already has them never churns. Broadened
-  // (issue #843) so a flat `src/` — holding only files, no subdirectories — is
-  // introspected via those files, the same files-and-flat recognition the
-  // Python/Rust extractors above already have.
+  // The `src/` layout (TS/JS) is authoritative. Its immediate child directories
+  // AND loose source files are modules: mixed trees are conventional, and
+  // returning one kind or the other silently truncated them (issue #1551).
   const sourceNodes = enumerateJsSourceRoot(
```
TheMostlyGreat: **Implementation trap for the exclusion above** — flagging so it doesn't bite whoever picks it up.

Four lines below this (`if (sourceNodes.length > 0) return { nodes: sourceNodes };`, outside this hunk) is an early return keyed on node *count*.

A package whose `src/` holds only `index.ts` is common. Excluding `index` naively empties `sourceNodes`, so extraction falls straight past that return into the `go.mod` check, then the `lib/` probe, then `topLevelJsModuleNodes` — and the doc ends up describing root-level scripts instead of the package.

Same shape for the test exclusion: a `src/` holding only `*.test.ts` shouldn't silently fall through either.

The exclusion needs to apply only where it doesn't empty the result, or that early return needs to key off "the root had eligible entries" rather than `length > 0`.

packages/cli/architecture.generated.md
```diff
@@ -37,6 +45,30 @@ No description yet — awaiting prose.
...
+
+### health
+
+<!-- reconciled: 41b56941957c005af2f9c8f78439faada884c97306a75cc3e7c06947aa54ed1d -->
+
+`src/health.ts`
+
+No description yet — awaiting prose.
+
+### index
```
TheMostlyGreat, line 64: Symptom of the eligibility issue — `src/index.ts` is a barrel of re-exports, now sitting as a peer of `commands/`, `utils/`, and `hooks/`. Ten of these sections were added to this file (`cli`, `index`, `health`, `parity`, `schema`, `version`, …).

Some are genuinely modules and belong (`reconcile`, `parity`, `schema`). `index` is the clear-cut one that doesn't.

packages/website/architecture.generated.md
```diff
@@ -15,10 +15,14 @@ fingerprint: ea4bd60c31a3ee3ec521b60ceb7c7dc29864353fab6c50ecef54678ac242e837
 
 No description yet — awaiting prose.
 
+> ⚠ stale: structure changed since this section was reconciled.
```
TheMostlyGreat: **This is the sharpest thing I found.** This package gained *zero* new modules — the whole diff is a fingerprint change plus two of these markers.

Both sections now claim *"structure changed since this section was reconciled."* The structure did not change. The fingerprint **schema** did.

That collides with the contract at the top of `features/architecture-state-docs.feature`: the doc *"may be incomplete but is never silently wrong."* This is it being wrong, quietly, in the one field a reader trusts to tell them their prose needs attention.

The mechanism is pre-existing and coarse by design — any fingerprint move stales every section, not just the affected one — so I'm not calling this a new bug in reconcile. What's new is the **trigger**: a schema change moves the fingerprint for every project at once, so on upgrade every consumer sees every module flagged stale for a non-structural reason. On a repo with real prose written (unlike ours, all placeholders), that's a meaningful false alarm.

Two ways out: a fingerprint schema version that suppresses the stale sweep when only the hash *recipe* changed and the node set is identical; or, if that's too much machinery for one migration, release notes that say plainly "all architecture sections will flag stale once — expected, not a signal about your prose."

packages/cli/src/utils/architecture-fingerprint.ts
```diff
@@ -48,12 +48,10 @@ interface ShapeInputs {
 }
 
 function collectShapeInputs(projectDirectory: string): ShapeInputs {
-  const moduleNames = extractSkeleton(projectDirectory)
-    .nodes.map(node => node.name)
-    .toSorted(byString);
+  const modules = extractSkeleton(projectDirectory).nodes.map(({ name, path }) => ({ name, path }));
```
TheMostlyGreat: **Lost its defensive sort.** This used to be `.toSorted(byString)`; determinism now rests on all six `extractSkeleton` branches sorting internally.

They all do today, so this is correct as written — but the module header three lines up still promises determinism it no longer enforces, and a future branch that forgets produces platform-dependent hashes with no test to catch it. Cheap to keep:

```suggestion
  const modules = extractSkeleton(projectDirectory)
    .nodes.map(({ name, path }) => ({ name, path }))
    .toSorted((a, b) => a.name.localeCompare(b.name) || a.path.localeCompare(b.path));
```

Separately: I did chase whether `localeCompare` makes ordering locale-sensitive here (it's in principle, and this module's whole contract is determinism). Tested across four locales including hyphen cases where collation and code-unit order genuinely diverge — Node resolves the default collator to `en-US` regardless of `LANG`/`LC_ALL`. Not reproducible, no action needed.

planning/design/architecture-mixed-js-coverage-design.md
```diff
@@ -0,0 +1,101 @@
+# Design: Trustworthy mixed JS/TS architecture coverage
```
TheMostlyGreat: **Blocking — this bypasses the ticket system.**

Three brand-new top-level dirs (`planning/{design,test-definitions,user-stories}/`), none on `main`, holding artifacts the bdd skill places at `<namespace-root>/tickets/{ID}-{slug}/` (`bdd/SKILL.md:99`).

The owning ticket already exists: `.project/tickets/QD5DTT-architecture-state-docs/`. (`NTB1` in the new scenario tags is an AC namespace, not a ticket ID — worth double-checking those tags too.)

Two consequences beyond tidiness:
- Coverage/lineage enforcement is ticket-dir-scoped, so a `test-definitions.md` under `planning/` is invisible to it.
- `planning/` isn't in `META_PATHS`, so it counts against the LOC gate.

Move all three into the QD5DTT ticket dir.

packages/cli/tests/templates/architecture-guide.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const guidePath = nodePath.resolve(
+  import.meta.dirname,
+  '../../templates/guides/architecture-guide.md',
+);
+
+describe('architecture guide generated-document ownership', () => {
+  it('distinguishes derived structure from preserved module purpose prose', () => {
+    const guide = readFileSync(guidePath, 'utf8');
+
```
TheMostlyGreat, line 14: Non-blocking. The guide rewrite this asserts is a **genuine correction** — prose preservation really is the behavior (`architecture-document.ts:341`, ticket JT852Q), and the old "never by hand, regenerated deterministically" text was wrong. Good catch.

But this test locks *wording*, not behavior: it asserts the guide contains phrases. If preservation itself regresses, this stays green while the guide keeps promising it. JT852Q's tests cover the actual behavior, so the gap is small — just don't count this file as coverage of the contract it describes.

steps/architecture-state-docs.steps.ts
```diff
@@ -144,6 +184,68 @@ Then('the module is still listed in the skeleton', function (this: ArchitectureW
...
+
+Then(
+  'the generated architecture doc is byte-identical across both runs',
+  function (this: ArchitectureWorld) {
+    assert.equal(readDocument(this), this.documentBefore);
+  },
+);
+
+Given('a project with the installed architecture guide', function (this: ArchitectureWorld) {
+  const result = spawnSync('bun', [CLI_PATH, 'setup', '--yes'], {
```
TheMostlyGreat, line 229: FYI, two things about this step:

1. **It won't run in CI.** The root `features/` + `steps/` lane is local Stop-hook dogfooding, so all five new scenarios are unverified by the pipeline. The unit tests are the only CI-enforced coverage in this PR.

2. **It's the heaviest step in the file** — a full `safeword setup --yes` with a 120s timeout inside a cucumber step, against a command with a history of environment-sensitive exit-1. Worth considering whether the guide-ownership assertion needs a real setup at all, or whether the template-content test already covers it (it does, in `architecture-guide.test.ts`) and this scenario is buying you a second copy of the same assertion at 100× the cost.

packages/cli/src/utils/architecture-skeleton.ts
TheMostlyGreat: Resolved in f1b4227c7. Colocated test/spec files are now excluded only inside src/lib; top-level test-only packages retain #843 behavior. I kept index/main/cli intentionally: TypeScript treats entry files as modules, #843 already covers src/index.ts, and a basename cannot distinguish a pure barrel from real orchestration without violating content-agnostic extraction. Regression coverage documents this policy.

packages/cli/src/utils/architecture-skeleton.ts
TheMostlyGreat: Resolved in f1b4227c7. Source-root enumeration now returns an observed signal separately from emitted nodes, so a src/lib root containing only excluded tests yields an honest empty skeleton and cannot fall through to Go, lib, or unrelated root scripts. Unit and Cucumber regressions cover the case.

packages/cli/architecture.generated.md
TheMostlyGreat: Resolved by explicit policy in f1b4227c7. index remains listed because the extractor models TypeScript modules, not source-text semantics; blanket basename filtering would hide legitimate logic and regress the established flat src/index.ts contract. Test/spec noise is removed, while cli/main/index remain eligible.

packages/website/architecture.generated.md
TheMostlyGreat: Resolved in f1b4227c7. The released name-only fingerprint serialization is restored and pinned by an exact-hash compatibility test. Canonical path drift is reconciled explicitly, so file-to-directory changes heal references without a schema migration or false stale sweep. The unchanged Website document is now byte-identical to the base.

packages/cli/src/utils/architecture-fingerprint.ts
TheMostlyGreat: Resolved in f1b4227c7. The fingerprint layer again defensively sorts module names before hashing. Because the path-schema change was removed for compatibility, the original name sort is the applicable deterministic form.

planning/design/architecture-mixed-js-coverage-design.md
TheMostlyGreat: Resolved in f1b4227c7. The new requirements, RGR ledger, design choices, verification, and reopening provenance are consolidated into QD5DTT spec.md, test-definitions.md, impl-plan.md, verify.md, and ticket.md. The three top-level planning files were deleted.

packages/cli/tests/templates/architecture-guide.test.ts
TheMostlyGreat: Resolved with evidence, no production change needed. The template test remains documentation-contract coverage; JT852Q document tests cover actual prose preservation, and the real setup Cucumber scenario verifies that the corrected guide is installed. All remain green in the full suite.

steps/architecture-state-docs.steps.ts
TheMostlyGreat: Resolved after verification. The premise was incorrect: CI runs bun run test:bdd and the root config includes features/**/*.feature plus steps/**/*.ts. The real setup scenario is retained because it verifies setup-to-installed-guide wiring; the final full lane passed 514 scenarios and 15,832 steps.

packages/cli/src/utils/architecture-document.ts
```diff
@@ -230,9 +244,28 @@ function decideAction(
...
 }
 
+/**
+ * Compare rendered node references with the live skeleton. Canonical paths stay
+ * outside the released name-only fingerprint recipe so upgrades do not falsely
+ * stale every existing section, but path-only drift still self-heals.
+ */
+function hasMatchingNodePaths(content: string, nodes: SkeletonNode[]): boolean {
+  const paths = new Map<string, string>();
+  const pattern = /^### (.+)\r?\n+<!-- reconciled: \S+ -->\r?\n+\r?\n?`([^`]*)`\s*$/gm;
```
TheMostlyGreat: **The `\r?` here is decorative — it doesn't actually give you CRLF tolerance.** Verified by running both regexes against CRLF input:

```
parseSectionStamps on CRLF:   0 matches
hasMatchingNodePaths on CRLF: 0 matches
LF baseline:                  1 / 1
```

`\r?\n+` permits a single `\r` before a *run of `\n`s*, but the rendered shape has two consecutive newlines — `\r\n\r\n` — and the second `\r` has nowhere to go. `(.+)` also greedily swallows the first `\r`, so backtracking can't rescue it either.

Why it matters: on a Windows checkout with `core.autocrlf=true`, this returns `false` on every run → `decideAction` returns `'healed'` every session → the doc gets rewritten to LF perpetually and shows as permanently modified in git. That's a new failure mode, because before this commit a matching fingerprint short-circuited to `'unchanged'` without ever looking at byte shape.

Two-character fix:

```suggestion
  const pattern = /^### (.+?)(?:\r?\n)+<!-- reconciled: \S+ -->(?:\r?\n)+`([^`]*)`\s*$/gm;
```

(`.+?` so the lazy match doesn't eat the trailing `\r` before `.trim()` gets it.)

Worth knowing: `parseSectionStamps` two functions down has the identical flaw and doesn't even claim `\r?`. That one is pre-existing — on CRLF it silently loses every stamp, so all sections get re-stamped current. Not yours to fix here, but the prose parser's "CRLF-tolerant" docstring suggests Windows is a supported platform, so it's probably worth its own issue.

packages/cli/tests/utils/architecture-fingerprint.test.ts
```diff
@@ -100,6 +100,12 @@ afterEach(() => {
 });
 
 describe('shapeFingerprint — captures shape, not noise', () => {
+  it('keeps the released name-only fingerprint recipe stable', () => {
+    expect(shapeFingerprint(context.directory)).toBe(
+      '937a490e9b13dfe240f9abbc4ba05bd7f97e055d5ba3ae1efcfd5a47fdb477d4',
```
TheMostlyGreat: Right instinct — after the churn discussion, locking the released recipe is exactly the guard this needed. Two things make it weaker than it could be:

1. **Opaque failure.** When it breaks, the message is `expected '<hash>' to be '937a490e…'`. A dev who adds a legitimate new shape input (a new manifest source, a new schema extension) gets a hash mismatch with no indication of what changed or whether it was intentional.
2. **Coupled to the shared fixture.** It reads `context.directory`, which other tests in this file also set up. An unrelated change to that fixture breaks this test for a reason that has nothing to do with the recipe.

Cheapest improvement is just to make the failure self-explanatory:

```ts
expect(
  shapeFingerprint(context.directory),
  'Shape-fingerprint recipe changed. Every existing consumer doc will flag ⚠ stale once on upgrade. Only update this hash if that churn is intended.',
).toBe('937a490e…');
```

Stronger version, if you think it's worth it: export `collectShapeInputs` and assert its serialized shape instead of the digest — then the diff shows you *what* moved.

packages/cli/tests/utils/architecture-skeleton.test.ts
```diff
@@ -103,14 +103,52 @@ describe('extractSkeleton — broadened JS/TS recognition (issue #843)', () => {
...
+    const skeleton = extractSkeleton(context.directory);
+
+    expect(skeleton.nodes).toEqual([
+      { name: 'auth', path: `${root}/auth`, purpose: expect.any(String) as string },
+      { name: 'pipeline', path: `${root}/pipeline.ts`, purpose: expect.any(String) as string },
+    ]);
+  });
+
+  it.each(['src', 'lib'])(
+    'excludes colocated %s test files while keeping source modules and entry points',
```
TheMostlyGreat, line 119: Noting that I read this test name as your answer to my entry-point suggestion — `index.ts` stays a module, deliberately, and the assertion below pins it. Fair call, and I'm not re-litigating it: `src/index.ts` being a barrel is a convention, not a guarantee, and excluding it costs you a real module in every package that puts logic there.

One ask, since the decision is now load-bearing: the asymmetry with `RUST_ROOT_FILES` (which *does* exclude `lib.rs`/`main.rs` as "entry points, not modules", `architecture-skeleton.ts:29-32`) is currently only visible in this test's name. A future reader comparing the two extractors will read it as an oversight and "fix" it. One sentence next to `isJsTestFile` — why JS keeps entry points where Rust drops them — would make it stick.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -164,32 +162,39 @@ function declaresWorkspaces(projectDirectory: string): boolean {
...
+  );
+  const byName = new Map(
+    entries
+      .filter(entry => entry.isDirectory())
+      .map(entry => [
+        entry.name,
+        { name: entry.name, path: pathFor(entry.name), purpose: PURPOSE_PLACEHOLDER },
+      ]),
+  );
+  for (const node of jsFileNodes(entries, pathFor, true)) {
```
TheMostlyGreat: Nit, take it or leave it: the bare positional `true` doesn't say what it does at the call site — you have to jump to the signature to learn it means "exclude tests". Since there's exactly one caller passing it, an options object or a named local (`const excludeColocatedTests = true`) reads better. Purely cosmetic; the logic is right, and scoping the exclusion to this call rather than baking it into `isJsSourceModuleFile` is precisely what keeps #843's top-level case working.

TheMostlyGreat (commented):
## Quality Review — third pass (`f1b4227c`)

**Both blockers resolved. One of the fixes is better than either option I proposed.** Verdict moves from REQUEST CHANGES to **APPROVE**, with four non-blocking notes below — one of which is a correction to something I got wrong last round.

### What landed

| # | Finding | Status |
|---|---------|--------|
| 1a | Test files as modules | ✅ `isJsTestFile` + `excludeTests`, scoped to source roots only — `topLevelJsModuleNodes` still lists them, so #843 is intact |
| 1b | Entry points as modules | ➖ Deliberately kept — see note below, not re-litigating |
| 2 | `planning/` outside the ticket tree | ✅ Moved into `.project/tickets/QD5DTT-architecture-state-docs/`, with `impl-plan.md` + `verify.md` added |
| 3 | Fallthrough trap | ✅ `observed` flag replaces the `length > 0` return, with a test proving a test-only `src/` doesn't fall through |
| 4 | Lost defensive sort | ✅ Restored |
| 5 | False `⚠ stale` on every consumer | ✅ Solved better than I suggested |

**On #5** — I offered a fingerprint schema version or release-note framing. You did neither and picked a third option: keep the released name-only recipe untouched, and detect path drift out-of-band via `matchesExisting`/`hasMatchingNodePaths`. So path drift still self-heals, but the fingerprint never moves, so nothing falsely stales. `packages/website/architecture.generated.md` is now byte-identical to `main` — the two bogus markers are gone. The test asserting `action === 'healed'` *and* `not.toContain('⚠ stale')` pins exactly the property that matters. That's the right call and a cleaner design than what I proposed.

### Verified this pass

- **Prettier round-trip is clean.** The generated docs aren't in `.prettierignore`, and `hasMatchingNodePaths` now depends on their exact byte shape — so I ran `prettier --check` against the rendered doc. Passes. No formatter/regex conflict today; just be aware the coupling is new, and a one-line `.prettierignore` entry would make it structural rather than incidental.
- **Leaf docs are covered.** `leafTarget` delegates to `skeletonTarget`, so per-package docs in a monorepo get `matchesExisting` too. No gap.

### Correction — I was wrong about orphan tombstones

Last round I told you removing a module leaves an `⚠ orphaned` section that self-clears on the following heal, and used that to argue the eligibility rule was low-stakes. **That's wrong**, and it holds on this commit too.

The tombstone renders with no `<!-- reconciled: -->` stamp, so it drops out of `priorStamps` — but that only matters if a *re-render happens*. Once the doc is written, `decideAction` sees a matching fingerprint and (now) matching paths, returns `'unchanged'`, and never re-renders. The tombstone sits in the file until some unrelated structural change triggers the next heal.

That's pre-existing, not something this PR introduced, and not a blocker here. But it does mean the entry-point decision has a longer tail than I implied: if `index`/`cli` are ever excluded later, those tombstones persist in every consumer's doc rather than clearing themselves. Worth a follow-up issue for tombstone reaping independent of this PR.

**Next:** the CRLF item below is the only one I'd fix before merge — it's a two-character change. The rest are optional. `test (node 22.22.3)` still running; the other three are green.

packages/cli/src/utils/architecture-document.ts
TheMostlyGreat: Resolved in 9c029646d. Instead of expanding the regex (which tripped the repository unsafe-regex gate), both path matching and section-stamp parsing now operate on an LF-normalized read view. Matching CRLF documents remain byte-identical and unchanged, while structural CRLF heals preserve stamps and mark affected prose stale. Regression tests cover both cases.

packages/cli/tests/utils/architecture-fingerprint.test.ts
TheMostlyGreat: Resolved in 9c029646d. The exact-hash guard now includes an actionable Vitest failure message explaining that recipe changes require an explicit migration plan. collectShapeInputs remains private, avoiding production API surface solely for a test.

packages/cli/tests/utils/architecture-skeleton.test.ts
TheMostlyGreat: Resolved in 9c029646d. Added a durable source comment next to the JS filtering policy: JS index/main/cli names remain eligible because they are ordinary modules that may contain logic, unlike Rust fixed crate roots, and this content-agnostic extractor cannot infer barrel-only semantics.

packages/cli/src/utils/architecture-skeleton.ts
TheMostlyGreat: Resolved in 9c029646d. Replaced the positional true with a named excludeColocatedTests local while retaining the source-root-only scope that preserves the top-level test-package behavior.

PR #1577, opened 2026-07-28
Decouple GitHub token format validation

TheMostlyGreat:
## Summary

- Treat `GITHUB_TOKEN` as an opaque RFC 6750 Bearer credential instead of recognizing GitHub-specific prefixes or lengths.
- Preserve the exact `proxy-injected` fallback and prevent `gh auth token` from reintroducing that rejected value from its child environment.
- Add regression coverage for opaque credentials, RFC-valid nonliteral placeholders, terminal CR/LF rejection, Authorization-header propagation, terminal 401 behavior, and the default `gh` fallback boundary.

## Why

GitHub token formats can change independently of this CLI. The resolver leaves credential validity and authorization to GitHub rather than rejecting future valid token formats locally, while retaining the documented cloud-placeholder escape hatch.

Fixes #1520.

## Validation

- `bun run test` — 5,601 passed, 5 skipped
- `bun run test:bdd` — 505 scenarios passed, 3 skipped
- Build, typecheck, ESLint, Prettier, and `git diff --check`
- Fresh independent quality review against RFC 6750, GitHub token guidance, and GitHub CLI credential precedence
- Verification and audit workflows

### DIFF
--- a/packages/cli/src/retro/github-rest.test.ts
+++ b/packages/cli/src/retro/github-rest.test.ts
@@ -1,5 +1,9 @@
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
+const spawnSyncMock = vi.hoisted(() => vi.fn());
+
+vi.mock('node:child_process', () => ({ spawnSync: spawnSyncMock }));
+
 import { createRestTransport, resolveGitHubToken } from './github-rest.js';
 
 interface MockResponse {
@@ -60,6 +64,8 @@ function mockFetchCapturing(responder: (url: string) => MockResponse): CapturedC
 
 afterEach(() => {
   vi.unstubAllGlobals();
+  vi.unstubAllEnvs();
+  vi.clearAllMocks();
 });
 
 describe('createRestTransport', () => {
@@ -367,8 +373,21 @@ describe('createRestTransport', () => {
 
   // A wrong credential fails the same way every time. Retrying it once per
   // finding burns a whole sweep each and can trip rate limits (#1465 review).
-  it('#1520: sends an opaque Bearer credential to GitHub, where a terminal 401 is latched', async () => {
+  it('#1465: latches a terminal auth failure instead of re-sweeping per encounter', async () => {
     const calls = mockFetch(() => ({ ok: false, status: 401, json: () => ({}) }));
+    const transport = createRestTransport('tok');
+    if (!transport) throw new Error('expected a transport');
+
+    await expect(transport.searchBySignature('retro:abc123def456')).rejects.toThrow('401');
+    expect(calls).toHaveLength(1);
+
+    // The next encounter fails from the latch — no second request.
+    await expect(transport.searchByCanonical('canonical:abc123def456')).rejects.toThrow('401');
+    expect(calls).toHaveLength(1);
+  });
+
+  it('#1520: sends an opaque Bearer credential to GitHub, where its 401 is terminal', async () => {
+    const calls = mockFetchCapturing(() => ({ ok: false, status: 401, json: () => ({}) }));
     const token = resolveGitHubToken({ GITHUB_TOKEN: 'future-token~1' }, () => {
       throw new Error('gh fallback must not be consulted');
     });
@@ -378,6 +397,9 @@ describe('createRestTransport', () => {
 
     await expect(transport.searchBySignature('retro:abc123def456')).rejects.toThrow('401');
     expect(calls).toHaveLength(1);
+    expect((calls[0]?.init.headers as Record<string, string>).Authorization).toBe(
+      'Bearer future-token~1',
+    );
 
     // The next encounter fails from the latch — no second request.
     await expect(transport.searchByCanonical('canonical:abc123def456')).rejects.toThrow('401');
@@ -572,6 +594,8 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', () 
     ['single-character Bearer credential', 'a'],
     ['Bearer credential with every permitted punctuation character and padding', 'a-._~+/==='],
     ['arbitrary opaque Bearer credential', opaqueBearerToken],
+    ['RFC-valid value that is not the exact documented placeholder', 'not-set'],
+    ['case variant of the exact documented placeholder', 'PROXY-INJECTED'],
   ])('accepts a %s from GITHUB_TOKEN without consulting gh', (_label, shaped) => {
     let ghConsulted = false;
     const token = resolveGitHubToken({ GITHUB_TOKEN: shaped }, () => {
@@ -582,23 +606,23 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', () 
     expect(ghConsulted).toBe(false);
   });
 
-  it.each([
-    ['arbitrary opaque credential', opaqueBearerToken],
-    ['stateless GitHub credential', representativeStatelessToken],
-  ])('passes a selected %s GITHUB_TOKEN to the REST transport', async (_label, shaped) => {
-    const calls = mockFetchCapturing(() => ({ json: () => ({ id: 99, body: 'hi' }) }));
-    const token = resolveGitHubToken({ GITHUB_TOKEN: shaped }, () => {
-      throw new Error('gh fallback must not be consulted');
-    });
-    const transport = createRestTransport(token);
-    if (!transport) throw new Error('expected a transport');
+  it.each([['stateless GitHub credential', representativeStatelessToken]])(
+    'passes a selected %s GITHUB_TOKEN to the REST transport',
+    async (_label, shaped) => {
+      const calls = mockFetchCapturing(() => ({ json: () => ({ id: 99, body: 'hi' }) }));
+      const token = resolveGitHubToken({ GITHUB_TOKEN: shaped }, () => {
+        throw new Error('gh fallback must not be consulted');
+      });
+      const transport = createRestTransport(token);
+      if (!transport) throw new Error('expected a transport');
 
-    await transport.createComment(42, 'hi');
+      await transport.createComment(42, 'hi');
 
-    expect((calls[0]?.init.headers as Record<string, string>).Authorization).toBe(
-      `Bearer ${shaped}`,
-    );
-  });
+      expect((calls[0]?.init.headers as Record<string, string>).Authorization).toBe(
+        `Bearer ${shaped}`,
+      );
+    },
+  );
 
   it.each([
     ['a proxy placeholder', 'proxy-injected'],
@@ -606,11 +630,31 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', () 
     ['a value containing a space', 'opaque token'],
     ['a value containing a tab', 'opaque\ttoken'],
     ['a value containing a newline', 'opaque\ntoken'],
+    ['a value ending in a newline', 'opaque-token\n'],
+    ['a value ending in a carriage return', 'opaque-token\r'],
     ['a value containing a NUL control character', 'opaque\0token'],
     ['a value with equals outside the optional suffix', 'opaque=token'],
   ])('rejects %s and falls back to gh', (_label, bogus) => {
     const token = resolveGitHubToken({ GITHUB_TOKEN: bogus }, () => ghToken);
     expect(token).toBe(ghToken);
+  });
+
+  it('#1602: removes a rejected GITHUB_TOKEN before asking gh for its credential', () => {
+    vi.stubEnv('GITHUB_TOKEN', 'proxy-injected');
+    vi.stubEnv('GH_TOKEN', 'explicit-gh-token');
+    spawnSyncMock.mockReturnValue({ status: 0, stdout: 'gh-keyring-token\n' });
+
+    expect(resolveGitHubToken()).toBe('gh-keyring-token');
+    expect(spawnSyncMock).toHaveBeenCalledWith(
+      'gh',
+      ['auth', 'token'],
+      expect.objectContaining({ encoding: 'utf8', timeout: 10_000 }),
+    );
+    const spawnCall = spawnSyncMock.mock.calls[0] as [string, string[], { env: NodeJS.ProcessEnv }];
+    if (!spawnCall) throw new Error('expected gh auth token to be called');
+    const options = spawnCall[2];
+    expect(options.env).toMatchObject({ GH_TOKEN: 'explicit-gh-token' });
+    expect(options.env).not.toHaveProperty('GITHUB_TOKEN');
   });
 
   // invisible-retro-claude.SM1.AC1 (token arm) — GITHUB_TOKEN present → the REST
--- a/packages/cli/src/retro/github-rest.test.ts
+++ b/packages/cli/src/retro/github-rest.test.ts
@@ -367,9 +367,13 @@ describe('createRestTransport', () => {
 
   // A wrong credential fails the same way every time. Retrying it once per
   // finding burns a whole sweep each and can trip rate limits (#1465 review).
-  it('#1465: latches a terminal auth failure instead of re-sweeping per encounter', async () => {
+  it('#1520: sends an opaque Bearer credential to GitHub, where a terminal 401 is latched', async () => {
     const calls = mockFetch(() => ({ ok: false, status: 401, json: () => ({}) }));
-    const transport = createRestTransport('tok');
+    const token = resolveGitHubToken({ GITHUB_TOKEN: 'future-token~1' }, () => {
+      throw new Error('gh fallback must not be consulted');
+    });
+    expect(token).toBe('future-token~1');
+    const transport = createRestTransport(token);
     if (!transport) throw new Error('expected a transport');
 
     await expect(transport.searchBySignature('retro:abc123def456')).rejects.toThrow('401');
@@ -550,10 +554,13 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', () 
   // `ghs_APPID_JWT` rollout shape.
   const minimumStatelessToken = `ghs_${'a'.repeat(10)}_${'b'.repeat(10)}.${'c'.repeat(6)}.${'d'.repeat(7)}`;
   const representativeStatelessToken = `ghs_1234567_${'a'.repeat(160)}.${'b'.repeat(160)}.${'c'.repeat(186)}`;
+  // RFC 6750 Bearer credentials are opaque to this resolver. This deliberately
+  // has no GitHub prefix or minimum length, so a future token format does not
+  // require a resolver release before it can reach GitHub for validation.
+  const opaqueBearerToken = 'future-token~1';
 
-  // #634 — every accepted GitHub token shape must survive resolution, or a
-  // regex that quietly stopped matching one form (e.g. fine-grained PATs) would
-  // break a working auth path with no test to catch it.
+  // Existing GitHub token fixtures remain regression controls; the resolver
+  // must also accept an arbitrary value that satisfies the Bearer grammar.
   it.each([
     ['classic PAT (ghp_)', `ghp_${'a'.repeat(32)}`],
     ['OAuth (gho_)', `gho_${'b'.repeat(32)}`],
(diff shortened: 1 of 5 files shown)


packages/cli/src/retro/github-rest.ts
```diff
@@ -55,20 +55,13 @@ function ghAuthToken(): string | undefined {
...
  */
-function looksLikeGitHubToken(value: string): boolean {
-  return (
-    /^ghs_[\w.-]{36,}$/.test(value) ||
-    /^gh[opusr]_[A-Za-z0-9]{20,}$/.test(value) ||
-    /^github_pat_\w{20,}$/.test(value) ||
-    /^[0-9a-f]{40}$/.test(value)
-  );
+function isUsableBearerCredential(value: string): boolean {
+  return value !== 'proxy-injected' && /^[\w.~+/-]+=*$/.test(value);
```
TheMostlyGreat: Reviewer: **the placeholder guard narrowed from a class to a single literal, and a miss is now unrecoverable.**

I probed the new predicate directly:

```
/^[\w.~+\/-]+=*$/.test('PROXY-INJECTED')  // true
/^[\w.~+\/-]+=*$/.test('proxy_injected')  // true
/^[\w.~+\/-]+=*$/.test('not-set')         // true
/^[\w.~+\/-]+=*$/.test('null')            // true
/^[\w.~+\/-]+=*$/.test('undefined')       // true
```

Each of those now reaches GitHub. That alone would be the accepted trade #1520 describes — except the 401 does not stop at "clearer failure": `isTerminalStatus` includes 401 (line 136), `loadOpenIssues` latches it (line 261), and `currentSnapshot` rethrows it to every later encounter (line 269). There is no post-401 retry against `gh auth token`. So the path is:

- **before:** unrecognized placeholder → fall back to `gh` → filing works wherever `gh` is authenticated
- **after:** unrecognized placeholder → 401 → latched for the transport's life → `zero filed, every finding failed`

That is the exact symptom #568 filed, and #634's stated purpose was avoiding "muddying diagnosis of cloud-filing failures" — which now holds for one string. #568's own wording is the general case, not the literal one: *"GITHUB_TOKEN is the platform's, not a GitHub token"* (echoed in `.safeword/hooks/lib/retro-draft-spool.ts:5`). A platform credential is exactly the thing that satisfies Bearer syntax and 401s.

Two cheap mitigations, either closes it:

1. On a 401 where the token came from `env.GITHUB_TOKEN`, retry once with `getGhToken()` when it yields a *different* value, then latch. That is a fallback on observed failure, not the "authorization preflight" the ticket puts out of scope — the API still decides validity, which is the whole point of the change.
2. Minimum viable: carry the credential *source* into the latched error so the message names `GITHUB_TOKEN` vs `gh auth token`. That preserves #634's diagnosis value without touching control flow.

Also ~free: compare case-insensitively, since `PROXY-INJECTED` passing while `proxy-injected` fails is hard to defend as intentional.

---

Separately, a naming point on the same line: `isUsableBearerCredential` is two rules under one name. `proxy-injected` **is** valid Bearer syntax, so `isUsableBearerCredential('proxy-injected') === false` reads as a contradiction at the call site. Splitting `isBearerCredentialSyntax(value)` from the placeholder check and composing them in `resolveGitHubToken` (line 85) puts the #634 rule where the decision is made, and makes it obvious where a second placeholder goes.

packages/cli/src/retro/github-rest.test.ts
```diff
@@ -367,9 +367,13 @@ describe('createRestTransport', () => {
 
   // A wrong credential fails the same way every time. Retrying it once per
   // finding burns a whole sweep each and can trip rate limits (#1465 review).
-  it('#1465: latches a terminal auth failure instead of re-sweeping per encounter', async () => {
+  it('#1520: sends an opaque Bearer credential to GitHub, where a terminal 401 is latched', async () => {
```
TheMostlyGreat: Reviewer: **this renames the #1465 terminal-latch regression to #1520 and folds a second concern into it.**

The comment directly above still cites `(#1465 review)`, and the sibling test 18 lines down is still `#1465: still retries a transient 5xx on the next encounter`. The latch pair — terminal-401 latched / transient-5xx retried — is the point, and it's now split across two issue numbers. A grep for `#1465` finds only the retryable half plus a dangling comment; the terminal half, which is the one that actually amplifies into a burnt sweep, is no longer findable from its own issue.

The added token-resolution assertions are also largely redundant: `passes a selected %s GITHUB_TOKEN to the REST transport` (line 583) already covers `future-token~1` reaching the `Authorization` header via `resolveGitHubToken`. What's left unique here is "an opaque credential's 401 is terminal" — which is true of `'tok'` too.

Suggest restoring `it('#1465: latches a terminal auth failure instead of re-sweeping per encounter')` with `createRestTransport('tok')`, and if a #1520 end-to-end is wanted, giving it its own `it()`. As written, reverting or adjusting the opaque-acceptance rule breaks a test whose name says #1465's concern.

packages/cli/src/retro/github-rest.test.ts
```diff
@@ -592,14 +602,12 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', ()
 
   it.each([
     ['a proxy placeholder', 'proxy-injected'],
```
TheMostlyGreat, line 628: Reviewer: **the one real regression risk in this change is the only thing the table doesn't pin.**

The reject rows correctly moved from format cases to grammar cases, and `'a proxy placeholder'` still covers the exact literal. But nothing here or in the accept table asserts what happens to a placeholder that *isn't* the documented literal — `'not-set'`, `'null'`, `'PROXY-INJECTED'`, `'undefined'`. That is precisely the behavior that changed, and it's currently unspecified by the suite: a future edit could tighten or loosen it with no test moving.

Add one row documenting the intent, in whichever table matches the decision on the `github-rest.ts:64` thread — e.g. `['a placeholder that is not the documented literal', 'not-set']`. Either placement is fine; leaving it untested is not, given it's the behavioral delta of the PR.

packages/cli/src/retro/github-rest.test.ts
```diff
@@ -592,14 +602,12 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', ()
...
-    ['a 35-character stateless token', minimumStatelessToken.slice(0, -1)],
-    ['a stateless token embedded in prose', `prefix-ghs_${'a'.repeat(40)}.${'b'.repeat(40)}`],
-    [
-      'a stateless token followed by trailing prose',
-      `ghs_${'a'.repeat(40)}.${'b'.repeat(40)} extra`,
-    ],
     ['an empty string', ''],
+    ['a value containing a space', 'opaque token'],
+    ['a value containing a tab', 'opaque\ttoken'],
+    ['a value containing a newline', 'opaque\ntoken'],
```
TheMostlyGreat, line 632: Reviewer: minor — **trailing `\n` / `\r` is the likeliest real-world malformed value and it's unpinned.**

`GITHUB_TOKEN=$(cat token.txt)` and friends produce `"ghp_…\n"`. Today that's correctly rejected: JS `$` without the `m` flag asserts end-of-input and does *not* match before a trailing newline (verified — `/^a$/.test('a\n')` is `false`), so it falls through to `gh`. The embedded-newline row above covers the character but not the terminal position, and they fail for different reasons in a regex.

Worth a row (`'ghp_…\n'`, and `'\r'`) specifically because the tempting "helpful" edit here — adding `m`, or relaxing to `\s*$` — would silently reopen header injection into `` `Bearer ${token}` `` at `github-rest.ts:90`. A test is the cheapest way to keep that edit from looking harmless.

TheMostlyGreat (commented):
Reviewer: quality review of the #1520 decoupling.

The core call is right and the grammar is exactly RFC 6750 `b64token` — I checked `1*( ALPHA / DIGIT / "-" / "." / "_" / "~" / "+" / "/" ) *"="` against `/^[\w.~+/-]+=*$/` and it matches, including that `=` is a suffix only. CRLF injection into `Bearer ${token}` stays blocked, the existing classic/stateless/fine-grained/hex fixtures are correctly kept as regression controls, and CI is green on all four checks (I also ran `src/retro/github-rest.test.ts` locally — 74 passing).

One finding I'd want resolved before merge, plus four smaller ones. All inline.

**The one that matters (inline on `github-rest.ts:64`):** rejecting only the exact literal `proxy-injected` means every *other* placeholder now reaches the API, 401s, and gets latched terminal for the transport's life with no `gh` fallback. #1520 predicted "a clearer API authentication failure"; the actual outcome in that case is `zero filed, every finding failed` — which is verbatim the symptom #568 reported. Worth deciding explicitly rather than inheriting.

**Out-of-scope note, recorded not requested:** the invariant "anything the resolver accepts, `egress.ts` redacts" no longer holds. The scrubber's GitHub patterns (`egress.ts:37-39`) can't match an opaque value, and the entropy backstop (`HIGH_ENTROPY_RUN`, `egress.ts:204`) only weighs runs of ≥20 chars over `[\w+/.-]` — this PR's own fixture `future-token~1` is 14 chars and its `~` isn't in that alphabet, so it falls through both. #1520 deliberately parks the scrubber and I agree with that call (over-redaction is the safe direction there); flagging so the widened gap is a recorded decision rather than a side effect.

Note this review is posted as a COMMENT, not an approval — `gh` authenticates as the PR author, so it carries no merge block.

packages/cli/src/retro/github-rest.ts
TheMostlyGreat: Resolved in c8c36518a. `isBearerCredentialSyntax` now contains only the RFC rule, while `resolveGitHubToken` owns the exact, case-sensitive `proxy-injected` exception required by #1520. I also fixed the #1602 addendum: the default `gh` child removes the rejected `GITHUB_TOKEN` and preserves independent `GH_TOKEN`; the process-boundary test pins both.

packages/cli/src/retro/github-rest.test.ts
TheMostlyGreat: Resolved in c8c36518a. Restored the original #1465 terminal-401 latch regression with `tok`; #1520 now has its own resolver → transport test that asserts the opaque Bearer header and terminal 401 behavior.

packages/cli/src/retro/github-rest.test.ts
TheMostlyGreat: Resolved in c8c36518a. Added distinct terminal-newline and terminal-carriage-return rejection cases to guard the header-construction boundary.

packages/cli/src/retro/github-rest.test.ts
TheMostlyGreat: Resolved in c8c36518a. The acceptance table now explicitly pins `not-set` and `PROXY-INJECTED` as RFC-valid values that are not the exact documented placeholder, so the intentionally changed behavior cannot drift silently.

packages/cli/src/retro/github-rest.ts
```diff
@@ -43,10 +43,23 @@ const MAX_ISSUE_PAGES = 10;
...
-    const result = spawnSync('gh', ['auth', 'token'], { encoding: 'utf8', timeout: 10_000 });
+    const childEnvironment = { ...environment };
+    delete childEnvironment.GITHUB_TOKEN;
+    const result = spawnSync('gh', ['auth', 'token'], {
+      encoding: 'utf8',
+      timeout: 10_000,
+      env: childEnvironment,
+    });
     const token = (result.stdout ?? '').trim();
     return result.status === 0 && token.length > 0 ? token : undefined;
```
TheMostlyGreat, line 64: Reviewer: **the `gh` result is now the only unguarded path into the `Authorization` header.**

This commit tightens `GITHUB_TOKEN` — it must have Bearer syntax and must not be the placeholder — and simultaneously documents (lines 48-50) that `GH_TOKEN` is *preserved* in the child env as "an independent documented gh credential source with higher precedence." Both halves are defensible on their own. Together they leave an asymmetry:

- `GITHUB_TOKEN=proxy-injected` → rejected, falls through
- `GH_TOKEN=proxy-injected` → `gh` echoes it → returned unchecked → straight into `` `Bearer ${token}` `` at line 109

So the value #634 exists to keep out of the API still reaches it, just through the sibling variable. `GH_TOKEN`'s higher precedence in `gh` makes it the *more* likely one to win where a platform sets both.

There's a second, smaller reason to check here: `.trim()` strips surrounding whitespace but not interior control characters, so a `gh` credential carrying an embedded CRLF would reach header construction. undici rejects invalid header values, so this is robustness rather than a vulnerability — but the env path is validated against exactly this and the `gh` path isn't, which is the kind of gap that only stays harmless by accident.

One line covers both, and `token.length > 0` becomes redundant since the pattern's `+` already requires a character:

```ts
return result.status === 0 && isBearerCredentialSyntax(token) ? token : undefined;
```

A `gh` credential that fails this returns `undefined`, which the caller already handles as a graceful no-op — strictly better than passing a value we know is unusable.

Worth stating the intent either way: if `GH_TOKEN` is meant to be trusted verbatim as an operator override, say so in the docblock, because right now the asymmetry reads as an oversight rather than a decision.

packages/cli/src/retro/github-rest.ts
```diff
@@ -77,18 +82,25 @@ function looksLikeGitHubToken(value: string): boolean {
...
-  if (fromEnvironment && looksLikeGitHubToken(fromEnvironment)) return fromEnvironment;
-  return getGhToken();
+  if (
+    fromEnvironment &&
+    fromEnvironment !== 'proxy-injected' &&
+    isBearerCredentialSyntax(fromEnvironment)
+  ) {
+    return fromEnvironment;
+  }
+  return getGhToken(env);
```
TheMostlyGreat, line 103: Reviewer: minor — **the `env` parameter now carries two jobs, and the second one bites any caller that uses the first.**

`env` was "where to read `GITHUB_TOKEN` from." It is now also "the environment `gh` inherits." For production that's invisible: every call site is zero-arg (`retro.ts:401`, `retro.ts:465`, both live smoke tests), so `env` is `process.env` and the child gets a full environment.

But a caller who overrides `env` for the reading job silently rewrites the subprocess environment too:

```ts
resolveGitHubToken({ GITHUB_TOKEN: 'proxy-injected' })
// → ghAuthToken({ GITHUB_TOKEN: 'proxy-injected' })
// → childEnvironment = {}  ← no PATH, no HOME
// → gh not found → undefined
```

That returns `undefined` while a perfectly good `gh` credential exists, and it fails as "no token available" rather than as anything pointing at the cause. Today every test that overrides `env` also injects `getGhToken`, so nothing hits it — the trap is armed but not stepped on.

Cheapest fix is to keep the two environments separate: have `ghAuthToken` read `process.env` for the child and take only the strip-list from the caller, or default the child env independently of the lookup env. Alternatively just document the fusion on the parameter — but a comment is weaker than not having the coupling, given the failure is silent.

packages/cli/src/retro/github-rest.test.ts
```diff
@@ -60,6 +64,8 @@ function mockFetchCapturing(responder: (url: string) => MockResponse): CapturedC
 
 afterEach(() => {
   vi.unstubAllGlobals();
+  vi.unstubAllEnvs();
+  vi.clearAllMocks();
```
TheMostlyGreat, line 68: Reviewer: minor — **`clearAllMocks()` doesn't undo `mockReturnValue`, so the #1602 stub outlives its test.**

Per vitest 4's own typings, `mockClear` — which `vi.clearAllMocks()` calls — resets `mock.calls`/`instances`/`results` only; resetting the *implementation* is `mockReset`'s job (`@vitest/spy` docs the difference explicitly at `mockReset(): "Does what mockClear does and resets inner implementation"`).

So `spawnSyncMock.mockReturnValue({ status: 0, stdout: 'gh-keyring-token\n' })` at line 645 stays in force for every test after it in the file. Nothing currently breaks — every later test injects `getGhToken` explicitly, so the default `ghAuthToken` is never reached — but the file now has two order-dependent failure modes waiting for the next test that calls `resolveGitHubToken()` with defaults:

- placed *before* line 645, `spawnSyncMock()` returns `undefined` and `result.stdout` throws `TypeError`
- placed *after*, it silently resolves to `gh-keyring-token` and passes for the wrong reason

The second is the dangerous one — a test that appears to prove a fallback works while actually reading a neighbour's leftovers.

`vi.resetAllMocks()` in the `afterEach`, or a `beforeEach` setting a safe default return, removes both. Now that `node:child_process` is mocked file-wide (line 3), this is the seam where that decision has to be made once.

packages/cli/src/retro/github-rest.test.ts
```diff
@@ -572,39 +606,57 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', ()
...
-    ['classic opaque', `ghs_${'a'.repeat(40)}`],
-    ['stateless', representativeStatelessToken],
-  ])('passes a selected %s GITHUB_TOKEN to the REST transport', async (_label, shaped) => {
-    const calls = mockFetchCapturing(() => ({ json: () => ({ id: 99, body: 'hi' }) }));
-    const token = resolveGitHubToken({ GITHUB_TOKEN: shaped }, () => {
-      throw new Error('gh fallback must not be consulted');
-    });
-    const transport = createRestTransport(token);
-    if (!transport) throw new Error('expected a transport');
+  it.each([['stateless GitHub credential', representativeStatelessToken]])(
```
TheMostlyGreat, line 609: Reviewer: nit, take or leave — this `it.each` is down to a single row now that the opaque case moved into the `#1520` test. A plain `it('passes a selected stateless GitHub credential to the REST transport')` says the same thing with less indentation and no table to scan.

(Keeping it as an `each` is also fine if you expect rows to come back.)

TheMostlyGreat (commented):
Reviewer: second-pass review at `c8c36518` (my first was against `fb80edf8`). CI green on all four checks; `src/retro/github-rest.test.ts` green locally — 79 passing, up from 74.

**All five findings from the first pass are resolved, and correctly:**

- `#1465` latch regression restored as its own test with `createRestTransport('tok')`, and the `#1520` end-to-end split out beside it — with an `Authorization` assertion added, which is better than what I asked for.
- `isUsableBearerCredential` split into `isBearerCredentialSyntax` plus the placeholder check composed at the call site. `proxy-injected` no longer hides inside a predicate whose name says the opposite.
- Trailing `\n` and `\r` pinned as reject rows.
- The non-literal-placeholder boundary is now pinned, and pinned *explicitly*: `'not-set'` and `'PROXY-INJECTED'` sit in the **accept** table. That is the decision I asked for, made in the direction I did not expect, and stated in a way the next reader can't miss. Good.
- #1602 fixed: `GITHUB_TOKEN` stripped from the `gh` child env, with a test asserting both the removal and `GH_TOKEN`'s survival.

Three new things introduced by this commit, none of them blocking. Details inline.

The first is the one I'd actually action: the credential `gh` returns is the only path into the `Authorization` header that no longer passes any check — including a `GH_TOKEN` placeholder, which the new docblock now deliberately preserves. One line closes it.

Still a COMMENT, not an approval — `gh` authenticates as the PR author.

packages/cli/src/retro/github-rest.ts
TheMostlyGreat: Addressed in #1647: `gh auth token` output is now checked with `isBearerCredentialSyntax()` before it can reach the REST transport. This preserves explicit `GH_TOKEN` precedence while rejecting malformed output safely.

packages/cli/src/retro/github-rest.ts
TheMostlyGreat: Addressed in #1647: the fallback child now always clones real `process.env` and strips only `GITHUB_TOKEN`; the injected resolver environment is lookup-only. A process-boundary regression test pins that separation.

packages/cli/src/retro/github-rest.test.ts
TheMostlyGreat: Addressed in #1647: test teardown uses `vi.resetAllMocks()`, and the next default-resolver test proves no configured `gh` response can leak across test cases.

packages/cli/src/retro/github-rest.test.ts
TheMostlyGreat: Addressed in #1647: the one-row `it.each` is now a plain named `it`, retaining the same stateless-token transport coverage.

PR #1613, opened 2026-07-28
Validate shipped TypeScript hook templates

TheMostlyGreat:
## Summary

Validates every published TypeScript hook template in its installed shape before release.

- Adds baseline lint, typed-preset parser/config, and strict TypeScript release checks.
- Fixes shipped-template diagnostics and preserves dogfood parity.
- Adds package-local Bun types required by the fixture.

## Validation

- `bun run --cwd packages/cli test:release`
- `bun run --cwd packages/cli lint`
- `bun run --cwd packages/cli test:bdd`
- `bun scripts/parity-check.ts`
- `bun audit`

The complete Vitest suite has a runner lifecycle hang in this environment after workers exit; focused/release/BDD checks passed.

### DIFF
--- a/.project/tickets/505-keep-shipped-hooks-host-lintable/ticket.md
+++ b/.project/tickets/505-keep-shipped-hooks-host-lintable/ticket.md
@@ -1,0 +1,78 @@
+---
+id: 505
+slug: keep-shipped-hooks-host-lintable
+type: task
+phase: done
+status: done
+external_issue: https://github.com/ArcadeAI/safeword/issues/505
+created: 2026-07-28T00:00:00Z
+last_modified: 2026-07-28T14:59:03Z
+scope:
+  - Keep every shipped TypeScript hook compatible with the supported ESLint baseline used by host projects.
+  - Add release validation for the physical distributed hook surface: baseline lint, typed-preset parse/config loading, and strict installed-shape TypeScript checking.
+out_of_scope:
+  - Adding a host-project lint ignore for .safeword/hooks.
+  - Normalizing the existing hook corpus to every stylistic and security rule in the full typed host preset.
+done_when:
+  - Shipped TypeScript hooks have no errors from the supported ESLint baseline.
+  - Every physical shipped hook loads without a fatal parser/config diagnostic under the actual typed host preset.
+  - Every physical shipped hook typechecks in a generated installed-shape fixture without customer dependencies.
+---
+
+# Keep shipped hooks host-lintable
+
+**Goal:** Let a Safeword upgrade complete without shipped hooks blocking the host project's lint.
+
+**Why:** Installed hooks are part of the host repository's lint surface and can otherwise break a customer commit after an upgrade.
+
+**Type:** Bug
+
+**Scope:** Remove the reported lint violations from the shipped TypeScript hook
+templates and add release-lane validation for every shipped TypeScript hook:
+the supported host baseline, fatal parser/config loading under Safeword's typed
+preset, and strict typechecking in its installed shape.
+
+**Out of Scope:** Ignoring `.safeword/hooks/**` in generated host configs,
+changing unrelated ESLint policies, or expanding validation to non-hook templates.
+
+## Decision
+
+Use the preferred upstream contract: shipped hooks stay in host lint scope and
+must pass the supported ESLint baseline. The release test validates the exact
+`packages/cli/templates/hooks/**` TypeScript surface that the package publishes,
+including parser/config integration and strict typechecking in a temporary
+installed fixture. Full-preset policy cleanup is separately scoped because the
+pre-existing corpus has a broad backlog unrelated to issue #505.
+
+**Figure-it-out:**
+
+- [x] Phase 1: Frame the choice as lint-clean templates versus a generated host ignore.
+- [x] Phase 2: Compare upstream validation, a host ignore, and both together.
+- [x] Phase 3a: Examine ESLint rule semantics, package distribution, and template/reconciliation ownership.
+- [x] Phase 3b: Validate the two rules with current ESLint documentation and inspect Safeword's published `templates` contract.
+- [x] Phase 4: Commit to upstream lint validation and minimal source fixes.
+
+## Done When:
+
+- [x] The literal BOM and dead-store initializations no longer create host-lint failures.
+- [x] A release test checks every shipped TypeScript hook with the supported baseline.
+- [x] A release test loads every shipped hook through the typed host preset and fails on fatal parser/config diagnostics.
+- [x] A release test strictly typechecks the installed-shape hook tree with package-owned `@types/bun` support.
+
+## Test Definitions:
+
+- [x] RED — The baseline reported five current violations across the shipped hook tree.
+- [x] GREEN — The release test reports no errors after the minimal template fixes.
+- [x] RED — The installed-shape strict typecheck exposed 22 real template diagnostics; the minimal fixes preserve each hook's fail-open behavior.
+- [x] GREEN — The installed fixture typechecks all physical hook templates without customer dependencies.
+- [x] REFACTOR — The test has one owner, discovers the shipped hook tree instead of maintaining a file list, and centralizes Claude session resolution for the two retro filing hooks.
+
+## Work Log
+
+- 2026-07-28T00:00:00Z Started: Triaged GitHub issue #505 and created this scoped task record.
+- 2026-07-28T00:00:00Z Figure-it-out: chose host-lint-clean templates plus upstream release validation; a host ignore would conceal defects in shipped code and leave customers without lint coverage.
+- 2026-07-28T07:22:02Z RED/GREEN: added a release-lane ESLint gate for the shipped hook tree; it found five current errors. Replaced the literal BOM and removed four dead-store initializations without changing fallback behavior.
+- 2026-07-28T07:22:02Z Verified: focused gate, package typecheck, root lint, formatting, and the release suite passed (23 tests). The first full release attempt had one unrelated five-second timing timeout in `pre-tool-git-bare-fix`; its isolated rerun and the clean full rerun passed.
+- 2026-07-28T14:38:21Z Quality review closed a coverage gap: added a package-owned `@types/bun` installed-fixture typecheck and fixed 22 strict diagnostics across the physical hook tree. The typed host preset now also loads that fixture and fails on fatal parser/config errors; full-preset policy diagnostics remain explicitly out of scope because the existing corpus has a separate broad backlog.
+- 2026-07-28T14:38:21Z Refactor/audit: centralized Claude session environment resolution for the two retro filing hooks and synchronized all changed hook templates to dogfood copies.
+- 2026-07-28T14:59:03Z Final verification: release validation (25 tests), direct release-test lint, root lint, build, Gherkin acceptance, parity (195 pairs / 8 contracts), `bun audit`, and two independent quality reviews passed. Reverted the unrelated root patch upgrades found during audit; the required CLI-local Bun types remain. The complete Vitest suite runner hung after workers exited under isolated locks, recorded as a local evidence limitation in `verify.md`.
--- a/.project/tickets/505-keep-shipped-hooks-host-lintable/ticket.md
+++ b/.project/tickets/505-keep-shipped-hooks-host-lintable/ticket.md
@@ -2,17 +2,18 @@
 id: 505
 slug: keep-shipped-hooks-host-lintable
 type: task
-phase: done
-status: done
+phase: verify
+status: in_progress
 external_issue: https://github.com/ArcadeAI/safeword/issues/505
 created: 2026-07-28T00:00:00Z
-last_modified: 2026-07-28T14:59:03Z
+last_modified: 2026-07-29T01:00:00Z
 scope:
-  - Keep every shipped TypeScript hook compatible with the supported ESLint baseline used by host projects.
-  - Add release validation for the physical distributed hook surface: baseline lint, typed-preset parse/config loading, and strict installed-shape TypeScript checking.
+  - Keep every shipped TypeScript template compatible with the supported ESLint baseline used by host projects.
+  - Add release validation for the schema-declared distributed TypeScript surface: baseline lint, typed-preset parse/config loading, and strict installed-shape TypeScript checking.
+  - Keep package-owned Codex runtime template mirrors under parity validation.
 out_of_scope:
   - Adding a host-project lint ignore for .safeword/hooks.
-  - Normalizing the existing hook corpus to every stylistic and security rule in the full typed host preset.
+  - Normalizing the existing template corpus to every stylistic and security rule in the full typed host preset.
 done_when:
   - Shipped TypeScript hooks have no errors from the supported ESLint baseline.
   - Every physical shipped hook loads without a fatal parser/config diagnostic under the actual typed host preset.
@@ -28,21 +29,23 @@ done_when:
 **Type:** Bug
 
 **Scope:** Remove the reported lint violations from the shipped TypeScript hook
-templates and add release-lane validation for every shipped TypeScript hook:
+templates and add release-lane validation for every schema-declared shipped TypeScript template:
 the supported host baseline, fatal parser/config loading under Safeword's typed
 preset, and strict typechecking in its installed shape.
 
 **Out of Scope:** Ignoring `.safeword/hooks/**` in generated host configs,
-changing unrelated ESLint policies, or expanding validation to non-hook templates.
+changing unrelated ESLint policies, or validating arbitrary customer dependency versions.
 
 ## Decision
 
-Use the preferred upstream contract: shipped hooks stay in host lint scope and
-must pass the supported ESLint baseline. The release test validates the exact
-`packages/cli/templates/hooks/**` TypeScript surface that the package publishes,
-including parser/config integration and strict typechecking in a temporary
-installed fixture. Full-preset policy cleanup is separately scoped because the
-pre-existing corpus has a broad backlog unrelated to issue #505.
+Use the preferred upstream contract: shipped templates stay in host lint scope
+and must pass the supported ESLint baseline. The release test derives the exact
+TypeScript install surface from `SAFEWORD_SCHEMA`, including parser/config
+integration and strict typechecking in a temporary installed fixture using
+Safeword's package-pinned development types. Full-preset policy cleanup is
+separately scoped because the pre-existing corpus has a broad backlog unrelated
+to issue #505. Codex runtime assets remain managed (they must not be installed
+in customer repositories) but opt into dogfood parity so template edits cannot drift.
 
 **Figure-it-out:**
 
@@ -51,6 +54,14 @@ pre-existing corpus has a broad backlog unrelated to issue #505.
 - [x] Phase 3a: Examine ESLint rule semantics, package distribution, and template/reconciliation ownership.
 - [x] Phase 3b: Validate the two rules with current ESLint documentation and inspect Safeword's published `templates` contract.
 - [x] Phase 4: Commit to upstream lint validation and minimal source fixes.
+- [x] PR-review follow-up: expand the fixture to all schema-declared distributed
+  TypeScript templates, because statusline and BDD files are customer-facing and
+  the former already exposed the same unchecked-index defect.
+- [x] PR-review follow-up: preserve Codex runtime adapters as managed files while
+  opting them into parity; moving them to owned files would change installation semantics.
+- [x] PR-review follow-up: remove the narrow exported Claude-session convenience
+  helper; the existing pure resolver already owns precedence and two explicit
+  environment projections keep the shipped public surface smaller.
 
 ## Done When:
 
@@ -66,6 +77,8 @@ pre-existing corpus has a broad backlog unrelated to issue #505.
 - [x] RED — The installed-shape strict typecheck exposed 22 real template diagnostics; the minimal fixes preserve each hook's fail-open behavior.
 - [x] GREEN — The installed fixture typechecks all physical hook templates without customer dependencies.
 - [x] REFACTOR — The test has one owner, discovers the shipped hook tree instead of maintaining a file list, and centralizes Claude session resolution for the two retro filing hooks.
+  - [x] RED — The schema-driven fixture caught the statusline unchecked-index error and includes the BDD step templates.
+  - [x] GREEN — The fixture typechecks every schema-declared TypeScript template in its real destination path and the Codex runtime mirrors have parity coverage.
 
 ## Work Log
 
@@ -76,3 +89,5 @@ pre-existing corpus has a broad backlog unrelated to issue #505.
 - 2026-07-28T14:38:21Z Quality review closed a coverage gap: added a package-owned `@types/bun` installed-fixture typecheck and fixed 22 strict diagnostics across the physical hook tree. The typed host preset now also loads that fixture and fails on fatal parser/config errors; full-preset policy diagnostics remain explicitly out of scope because the existing corpus has a separate broad backlog.
 - 2026-07-28T14:38:21Z Refactor/audit: centralized Claude session environment resolution for the two retro filing hooks and synchronized all changed hook templates to dogfood copies.
 - 2026-07-28T14:59:03Z Final verification: release validation (25 tests), direct release-test lint, root lint, build, Gherkin acceptance, parity (195 pairs / 8 contracts), `bun audit`, and two independent quality reviews passed. Reverted the unrelated root patch upgrades found during audit; the required CLI-local Bun types remain. The complete Vitest suite runner hung after workers exited under isolated locks, recorded as a local evidence limitation in `verify.md`.
+- 2026-07-29T00:50:00Z Reopened for PR review: nine actionable findings identified. The release fixture omitted schema-declared statusline and BDD TypeScript templates, and Codex runtime assets were managed-but-not-paired. The follow-up keeps each ownership contract intact while extending validation and parity.
+- 2026-07-29T01:00:00Z Review follow-up GREEN: strict typechecking now covers 106 schema-declared TypeScript templates in real destination paths, including statusline and BDD steps. The new Codex parity opt-in surfaced and healed one additional stale runtime mirror. Focused release tests, parity tests, direct lint, formatting, parity (200 pairs / 8 contracts), and an independent fresh review passed.
--- a/.project/tickets/505-keep-shipped-hooks-host-lintable/ticket.md
+++ b/.project/tickets/505-keep-shipped-hooks-host-lintable/ticket.md
@@ -6,7 +6,7 @@ phase: implement
 status: in_progress
 external_issue: https://github.com/ArcadeAI/safeword/issues/505
 created: 2026-07-28T00:00:00Z
-last_modified: 2026-07-29T02:00:00Z
+last_modified: 2026-07-29T02:20:00Z
 scope:
   - Keep every shipped TypeScript template compatible with the supported ESLint baseline used by host projects.
   - Add release validation for the schema-declared distributed TypeScript surface: baseline lint, typed-preset parse/config loading, and strict installed-shape TypeScript checking.
@@ -79,6 +79,10 @@ in customer repositories) but opt into dogfood parity so template edits cannot d
 - [x] REFACTOR — The test has one owner, discovers the shipped hook tree instead of maintaining a file list, and centralizes Claude session resolution for the two retro filing hooks.
   - [x] RED — The schema-driven fixture caught the statusline unchecked-index error and includes the BDD step templates.
   - [x] GREEN — The fixture typechecks every schema-declared TypeScript template in its real destination path and the Codex runtime mirrors have parity coverage.
+  - [x] RED — The generated `owned-paths.ts` module was absent from both ESLint
+    passes despite being part of the installed hook tree.
+  - [x] GREEN — Every lint/typecheck pass receives the generated module as part
+    of the same installed fixture file list.
 
 ## Work Log
 
@@ -91,13 +95,14 @@ in customer repositories) but opt into dogfood parity so template edits cannot d
 - 2026-07-28T14:59:03Z Final verification: release validation (25 tests), direct release-test lint, root lint, build, Gherkin acceptance, parity (195 pairs / 8 contracts), `bun audit`, and two independent quality reviews passed. Reverted the unrelated root patch upgrades found during audit; the required CLI-local Bun types remain. The complete Vitest suite runner hung after workers exited under isolated locks, recorded as a local evidence limitation in `verify.md`.
 - 2026-07-29T00:50:00Z Reopened for PR review: nine actionable findings identified. The release fixture omitted schema-declared statusline and BDD TypeScript templates, and Codex runtime assets were managed-but-not-paired. The follow-up keeps each ownership contract intact while extending validation and parity.
 - 2026-07-29T01:00:00Z Review follow-up GREEN: strict typechecking now covers 106 schema-declared TypeScript templates in real destination paths, including statusline and BDD steps. The new Codex parity opt-in surfaced and healed one additional stale runtime mirror. Focused release tests, parity tests, direct lint, formatting, parity (200 pairs / 8 contracts), and an independent fresh review passed.
+- 2026-07-29T02:20:00Z Regression follow-up: the generated owned-paths module was typechecked but omitted from both fixture lint passes. Added it to the common fixture file list; the focused release gate and direct lint pass now cover the same installed tree.
 
 ## Refactor Ledger
 
 Scout findings, ordered leaf-first. Each entry is one behavior-preserving change
 with its own test and commit.
 
-- [ ] P1 — Eliminate the duplicate parity-pair membership traversal in
+- [x] P1 — Eliminate the duplicate parity-pair membership traversal in
   `scripts/parity-check.ts`; the CLI count must derive from `runParity` so it
   cannot drift from check/sync coverage.
 - [x] P2 — Rename the broadened release test and its temporary-fixture prefix
@@ -109,5 +114,5 @@ with its own test and commit.
   abstraction would recreate a rejected shipped API.
 - [ ] Struck — Do not merge baseline and fatal-diagnostic formatting: their
   distinct predicates make the test policy legible.
-- [ ] Struck — Keep `as unknown as` for the regex tuple: direct conversion is
+- [x] Struck — Keep `as unknown as` for the regex tuple: direct conversion is
   rejected by the strict fixture compiler.
--- a/.project/tickets/505-keep-shipped-hooks-host-lintable/ticket.md
+++ b/.project/tickets/505-keep-shipped-hooks-host-lintable/ticket.md
@@ -6,7 +6,7 @@ phase: implement
 status: in_progress
 external_issue: https://github.com/ArcadeAI/safeword/issues/505
 created: 2026-07-28T00:00:00Z
-last_modified: 2026-07-29T02:20:00Z
+last_modified: 2026-07-29T02:30:00Z
 scope:
   - Keep every shipped TypeScript template compatible with the supported ESLint baseline used by host projects.
   - Add release validation for the schema-declared distributed TypeScript surface: baseline lint, typed-preset parse/config loading, and strict installed-shape TypeScript checking.
@@ -96,6 +96,7 @@ in customer repositories) but opt into dogfood parity so template edits cannot d
 - 2026-07-29T00:50:00Z Reopened for PR review: nine actionable findings identified. The release fixture omitted schema-declared statusline and BDD TypeScript templates, and Codex runtime assets were managed-but-not-paired. The follow-up keeps each ownership contract intact while extending validation and parity.
 - 2026-07-29T01:00:00Z Review follow-up GREEN: strict typechecking now covers 106 schema-declared TypeScript templates in real destination paths, including statusline and BDD steps. The new Codex parity opt-in surfaced and healed one additional stale runtime mirror. Focused release tests, parity tests, direct lint, formatting, parity (200 pairs / 8 contracts), and an independent fresh review passed.
 - 2026-07-29T02:20:00Z Regression follow-up: the generated owned-paths module was typechecked but omitted from both fixture lint passes. Added it to the common fixture file list; the focused release gate and direct lint pass now cover the same installed tree.
+- 2026-07-29T02:30:00Z Refactor: split release-fixture materialization from tsconfig generation without changing the installed file set, fixture lifetime, or compiler options. Focused release checks and direct lint passed.
 
 ## Refactor Ledger
 
@@ -107,12 +108,12 @@ with its own test and commit.
   cannot drift from check/sync coverage.
 - [x] P2 — Rename the broadened release test and its temporary-fixture prefix
   from “hooks” to “templates”.
-- [ ] P3 — Split the release fixture builder into named materialization and
+- [x] P3 — Split the release fixture builder into named materialization and
   tsconfig-writing helpers without changing fixture lifetime or contents.
-- [ ] Struck — Do not extract the two Claude environment projections: the
+- [x] Struck — Do not extract the two Claude environment projections: the
   exported helper was deliberately removed during review, so reintroducing an
   abstraction would recreate a rejected shipped API.
-- [ ] Struck — Do not merge baseline and fatal-diagnostic formatting: their
+- [x] Struck — Do not merge baseline and fatal-diagnostic formatting: their
   distinct predicates make the test policy legible.
 - [x] Struck — Keep `as unknown as` for the regex tuple: direct conversion is
   rejected by the strict fixture compiler.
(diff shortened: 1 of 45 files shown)


packages/cli/tests/shipped-hooks-validation.release.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      expect(
+        result.status,
+        `tsc exited ${result.status}\nstdout:\n${result.stdout}\nstderr:\n${result.stderr}`,
+      ).toBe(0);
+    } finally {
+      fixture.cleanup();
+    }
+  });
+});
+import eslintJs from '@eslint/js';
```
TheMostlyGreat: **Reviewer:** This import is at the bottom of the file, after the closing `});` of the `describe` block — while `supportedHostBaseline` reads `eslintJs` at line 31, 121 lines *above* its declaration.

It works (ESM hoists imports) and `eslint tests/shipped-hooks-validation.release.test.ts` exits 0 — I verified `simple-import-sort` isn't wired for this path, so nothing catches it. But it reads as a stray patch artifact, and this file's whole job is to be the thing that catches stray patch artifacts in shipped code.

Move it into the import block at line 19.

packages/cli/tests/shipped-hooks-validation.release.test.ts
```diff
@@ -0,0 +1,152 @@
...
+          noUncheckedIndexedAccess: true,
+          noUnusedLocals: false,
+          noUnusedParameters: false,
+          skipLibCheck: true,
+          strict: true,
+          target: 'ESNext',
+          typeRoots: [nodePath.join(cliRoot, 'node_modules', '@types')],
+          types: ['node', 'bun'],
+        },
+        include: ['hooks/**/*.ts'],
```
TheMostlyGreat: **Reviewer:** This `include` is the scope decision, and I think it's cut in the wrong place — the bug #505 describes is still live in a shipped sibling.

Three `.ts` templates land in customer repos but sit outside `templates/hooks/**`:

| template | installs to | registered as |
|---|---|---|
| `templates/statusline/reentry.ts` | `.safeword/statusline/reentry.ts` | `ownedFiles` |
| `templates/cucumber/world.ts` | `steps/world.ts` | `managedFiles` |
| `templates/cucumber/shared.steps.ts` | `steps/shared.steps.ts` | `managedFiles` |

I copied them into a fixture with **this file's exact compilerOptions** and ran the same `tsc`:

```
statusline/reentry.ts(64,42): error TS18048: 'latest' is possibly 'undefined'.
```

Source is `templates/statusline/reentry.ts:59` — `const latest = entries[entries.length - 1];` — then `:64` dereferences `latest.nextImperative`. That is the identical `noUncheckedIndexedAccess` failure mode this PR fixed 22 times inside `hooks/`, in a file that installs into the *same* `.safeword/` directory. Any host tsconfig that lints `.safeword/hooks/**` will lint `.safeword/statusline/**` too.

`steps/*.ts` is arguably worse — those install at the customer's **repo root**, squarely inside their default `eslint .` / `tsc --noEmit` scope. (I saw TS2307 on `@cucumber/cucumber` for those two, but that's my ad-hoc fixture lacking cucumber, not a defect — not claiming it.)

The ticket lists non-hook templates as out of scope, but its stated *Why* is about the host lint surface generally. Either widen this glob to the shipped TS surface, or file the follow-up now — accepting it silently means the next upgrade still breaks a customer commit the exact way #505 reported.

packages/cli/tests/shipped-hooks-validation.release.test.ts
```diff
@@ -0,0 +1,152 @@
...
+        .map(
+          message =>
+            `${nodePath.relative(cliRoot, result.filePath)}:${message.line}:${message.ruleId}`,
+        ),
+    );
+
+    expect(errors).toEqual([]);
+  });
+
+  it('loads every installed hook through Safeword’s actual typed host preset', async () => {
```
TheMostlyGreat: **Reviewer:** The name overclaims relative to the assertion. Filtering `message.fatal` keeps only parse/config-level crashes — every severity-2 rule error from the typed preset is discarded.

My mutation test makes the boundary concrete: injecting a `TS2322` into `templates/hooks/stop-reentry.ts` failed only the third test; this one stayed green.

The *scoping* is deliberate and defensible (the ticket explicitly defers full-preset policy cleanup). The problem is that "loads every installed hook through Safeword's actual typed host preset" reads to a future maintainer as "the hooks pass the typed preset," which will make a green release lane look like coverage it isn't. Rename to what it asserts — e.g. `'parses and resolves type info under the typed host preset without a fatal diagnostic'` — and the deferral documents itself.

TheMostlyGreat (commented):
## Reviewer: /quality-review + engineering review

_Posted as `COMMENT` — `gh` authenticates as the PR author, so `REQUEST_CHANGES` 422s. **This review carries no merge block**; treat the verdict as advisory._

**Verdict: REQUEST CHANGES** — one confirmed parity regression, one confirmed customer-facing bug of the exact class this PR exists to eliminate still shipping in a sibling file, plus craft issues in the new gate.

The core idea is right and the gate genuinely bites. I mutation-tested it: injecting `const mutationProbe: number = 'not a number'` into `templates/hooks/stop-reentry.ts` turned the suite red with `hooks/stop-reentry.ts(140,7): error TS2322`. CI wiring is real (`test:release` runs at `ci.yml:120` and `release.yml:79`). The new test passes locally — 3/3 in 13s.

| | |
|---|---|
| **Currency** | ✓ `@types/bun@1.3.14` is npm-latest and matches the installed `bun 1.3.14` |
| **Sources** | ✓ every claim below was executed in this session, not recalled |
| **Correct** | ⚠️ gate is sound; one mirror desynced, one shipped file left broken |
| **Elegant** | ⚠️ trailing import; test 1 and tests 2–3 validate different trees |
| **No-bloat** | ⚠️ one new shipped-lib export whose necessity I could not confirm |
| **Wiring** | ✓ real collaborators throughout — actual `ESLint` API, actual `recommendedTypeScript` preset, actual `tsc` subprocess, actual `generateOwnedPathsModule`. No internal-seam mocks. This is a genuine wiring test. |

---

### 🔴 Confirmed: `.safeword/hooks/codex/stop.ts` desynced, and parity-check is structurally blind to it

At the merge base the two files were byte-identical:

```
git show b8cc00881:packages/cli/templates/hooks/codex/stop.ts > /tmp/t1.ts
git show b8cc00881:.safeword/hooks/codex/stop.ts        > /tmp/t2.ts
diff /tmp/t1.ts /tmp/t2.ts   # → IDENTICAL_BEFORE_PR
```

This PR edits the template (`offsetState` alias, `let output: string`) and syncs 13 other `.safeword/hooks/*` mirrors, but not this one. `bun scripts/parity-check.ts --mode=all` still prints `All 195 pairs and 8 contracts in sync` — because `.safeword/hooks/codex/*` lives in `managedFiles` via `CODEX_RUNTIME_ASSETS` (`schema.ts:238`, spread at `:1059`), not `ownedFiles`, and the pair check only walks `ownedFiles`:

```
bun -e "import {SAFEWORD_SCHEMA} from './packages/cli/src/schema.js';
  console.log(Object.keys(SAFEWORD_SCHEMA.ownedFiles).filter(k=>k.includes('hooks/codex')).length)"
# → 0
```

So both the PR body's "preserves dogfood parity" and the green parity run in `verify.md` are true-but-blind here. This is a **latent pre-existing gap** the PR happens to be the first to trip — worth fixing at the source, not just re-copying the file.

**Fix:** sync `.safeword/hooks/codex/stop.ts`. Separately consider whether the five `CODEX_RUNTIME_ASSETS` deserve pair enforcement — a mirror nobody diffs will drift again.

---

### 🔴 Confirmed: the bug #505 describes is still shipping in `.safeword/statusline/reentry.ts`

Detail inline on the `include: ['hooks/**/*.ts']` line. Short version: I ran **this PR's exact fixture tsconfig** against the shipped TS that lands in customer repos but sits outside `templates/hooks/**`, and got

```
statusline/reentry.ts(64,42): error TS18048: 'latest' is possibly 'undefined'.
```

That is literally the `noUncheckedIndexedAccess` class this PR fixed 22 times elsewhere, in a file installed into the *same* `.safeword/` tree as the hooks. The ticket declares non-hook templates out of scope, but the ticket's own *Why* — "Installed hooks are part of the host repository's lint surface and can otherwise break a customer commit after an upgrade" — applies verbatim to this file.

---

### Checked and clean (no action)

- All 13 non-codex dogfood mirrors byte-synced; 195 pairs + 8 contracts green.
- Every `let x: T;` de-initialization is behavior-preserving — each `catch` assigns before first read: `codex/stop.ts:270`, `session-codex-start.ts:26`, `prompt-retro-nudge.ts:29`, `stop-retro-filing.ts:38`. Nice touch that these stay fail-open.
- `/^\uFEFF?---/` is exactly equivalent to the literal-BOM regex it replaces — good catch, that one was a real host-lint landmine.
- `parseInt(insMatch?.[1] ?? '0')`, `allEntries.at(-1)`, `currentPhase ? (… ?? null) : null` — all behavior-preserving.
- `knip` clean; `@types/bun` is not flagged unused despite being reachable only through a generated tsconfig `types` array.
- Security: `spawnSync` with array args (no shell), `mkdtempSync` + `rmSync` scoped to the temp dir. Nothing to flag.
- Vacuous-pass risk is low: `ignore: false` disables ignores, ESLint's default `errorOnUnmatchedPattern` throws on an empty match, and `tsc` emits TS18003 on an empty `include`. An explicit file-count assertion would still be cheap insurance and would document the "103 physical hook templates" figure from `verify.md`.

---

**Next:** sync `.safeword/hooks/codex/stop.ts` and move the trailing import to the top block (both one-liners); then decide on the statusline/steps gap — either widen `include` to the shipped TS surface, or open the follow-up issue now so it is tracked rather than silently accepted. The remaining inline notes are optional.

packages/cli/tests/shipped-typescript-templates-validation.release.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  return {
+    cleanup: () => {
+      rmSync(directory, { force: true, maxRetries: 3, recursive: true });
+    },
+    directory,
+  };
+}
+
+describe('shipped TypeScript hooks', () => {
+  it('pass Safeword’s supported host ESLint baseline', async () => {
```
TheMostlyGreat, line 143: **Reviewer:** This test lints `hooksDirectory` (the raw template tree) while tests 2 and 3 lint/typecheck the fixture. That divergence has one concrete consequence: `lib/owned-paths.ts` doesn't exist in `templates/hooks/` — it's synthesized by `createInstalledHooksFixture` — so the generated module is checked by `tsc` and by fatal-only lint, but **never by the supported host baseline**.

That baseline is the ticket's first `done_when` ("Shipped TypeScript hooks have no errors from the supported ESLint baseline"), and `owned-paths.ts` is as shipped as any other hook file. Pointing this test at a fixture too would close the hole and make all three tests validate one identical tree — which also removes a whole class of "why did only one of these go red" confusion later.

packages/cli/tests/shipped-typescript-templates-validation.release.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      expect(fatalErrors).toEqual([]);
+    } finally {
+      fixture.cleanup();
+    }
+  }, 30_000);
+
+  it('typechecks every template in its installed shape', () => {
+    const fixture = createInstalledHooksFixture();
+    try {
+      const result = spawnSync(tscPath, ['--project', 'tsconfig.json'], {
```
TheMostlyGreat, line 198: **Reviewer:** Minor, but this is a release gate, so its failure message matters.

`tscPath` is a hard-coded `packages/cli/node_modules/.bin/tsc`. Today that's correct — I confirmed the package-local binary exists and the workspace root has no `node_modules/.bin/tsc`. But if `typescript` ever hoists (a bun version bump, a dependency change), `spawnSync` returns `status: null` with `error: ENOENT` and empty `stdout`/`stderr`, so the gate fails with `expected null to be 0` and a blank diagnostic block — a genuinely confusing 20 minutes for whoever is cutting the release.

One line fixes it:

```ts
expect(result.error, `failed to spawn ${tscPath}`).toBeUndefined();
```

Or resolve via `import.meta.resolve('typescript')` and drop the path assumption entirely.

packages/cli/tests/shipped-typescript-templates-validation.release.test.ts
```diff
@@ -0,0 +1,152 @@
...
+          noEmit: true,
+          noFallthroughCasesInSwitch: true,
+          noImplicitOverride: true,
+          noUncheckedIndexedAccess: true,
+          noUnusedLocals: false,
+          noUnusedParameters: false,
+          skipLibCheck: true,
+          strict: true,
+          target: 'ESNext',
+          typeRoots: [nodePath.join(cliRoot, 'node_modules', '@types')],
```
TheMostlyGreat, line 105: **Reviewer:** Worth one line in the docblock. The header says the tree is validated "without requiring a customer project's dependencies" — but `typeRoots` points into `cliRoot/node_modules/@types`, so the gate validates against **Safeword's** `@types/node` and the newly-added `@types/bun`, not the customer's.

That's the right pragmatic call, and `@types/bun@1.3.14` is current (npm-latest, matches the installed `bun 1.3.14`). The limitation to name is version skew: a customer on an older `@types/node` can hit diagnostics this gate never sees. Right now the docblock reads as if the check were dependency-independent, which is a stronger guarantee than it makes.

packages/cli/templates/hooks/lib/retro-trigger.ts
```diff
@@ -243,6 +243,21 @@ export function resolveSessionId(
   return firstNonEmpty(input.session_id, env.CLAUDE_CODE_REMOTE_SESSION_ID, env.CLAUDE_SESSION_ID);
 }
 
+/**
+ * Resolve a Claude hook's session id using the process environment. Keeping
+ * this boundary here prevents each Claude-facing hook from rebuilding the same
+ * narrow environment object (and keeps cloud/local precedence in one place).
+ */
+export function resolveClaudeSessionId(
```
TheMostlyGreat: **Reviewer:** Question, not a blocker — flagged as unverified.

This adds a new **exported symbol to a shipped lib**, and its only behavioral difference from `resolveSessionId(input, process.env)` is projecting `process.env` down to the two keys `resolveSessionId` already reads. So it looks like it exists to satisfy a `process.env` assignability error introduced by the fixture's `types: ['node', 'bun']` — i.e. the test harness's typing shaping shipped API surface.

If that's what happened, the narrower fix is a cast or a widened parameter type at the two call sites, not a permanent public helper.

I tried to settle this by reverting both call sites in a scratch copy and re-running `tsc`, but the command was blocked, so **I can't confirm whether the indirection is required** — if it is, say so in the docblock and this becomes a non-issue.

Separately: the new test at `tests/hooks/retro-trigger.test.ts:122` asserts only the cloud branch. For a new exported entry point, input-wins and local-fallback are worth a line each — cheap, and they're the branches that actually differ between the two hooks' payload shapes.

packages/cli/templates/hooks/lib/re-entry.ts
```diff
@@ -38,11 +38,13 @@ const RECENT_TURNS = 10;
 export function parseLogLine(line: string): Entry | null {
   const match = LINE_REGEX.exec(line.trim());
   if (!match) return null;
+  const [timestamp, sessionId, ticket, nextImperative] = match.slice(1);
+  if (!timestamp || !sessionId || !ticket || !nextImperative) return null;
```
TheMostlyGreat, line 43: **Reviewer:** This guard is unreachable. `LINE_REGEX` is

```
/^(\S+)\s+(\S+)\s+(ticket=\S+)\s+Next:\s+(.+)$/
```

— `\S+`, `\S+`, `ticket=\S+`, `.+`; no group can capture an empty string, so if `exec` matched, all four are non-empty by construction.

Harmless at runtime, and it does silence `noUncheckedIndexedAccess`. But it's dead defensive code in a shipped hook that implies empty captures are a real case, which will cost the next reader a minute. `const [, timestamp, sessionId, ticket, nextImperative] = match;` plus a short comment that the regex guarantees non-empty groups would carry the same type-safety with none of the misdirection.

packages/cli/templates/hooks/codex/stop.ts
```diff
@@ -267,7 +264,7 @@ async function main(): Promise<string> {
   return SILENT;
 }
 
-let output = SILENT;
+let output: string;
```
TheMostlyGreat, line 267: **Reviewer:** 🔴 This file's dogfood mirror was not synced. `.safeword/hooks/codex/stop.ts` still has `let output = SILENT;` and the old `pendingOffsetState` dereferences.

The two files were byte-identical at the merge base (`git show b8cc00881:` on both → `diff` clean), and the other 13 `.safeword/hooks/*` mirrors in this PR *were* updated, so this is an omission rather than a deliberate exception.

The reason nothing caught it: `.safeword/hooks/codex/*` is registered in `managedFiles` via `CODEX_RUNTIME_ASSETS` (`schema.ts:238`, spread at `:1059`), not `ownedFiles` — and the parity pair check only walks `ownedFiles`. A schema query returns **0** codex entries there, which is why `parity-check --mode=all` happily reports `All 195 pairs and 8 contracts in sync`.

So the PR body's "preserves dogfood parity" and the green parity run in `verify.md` are both accurate and both blind. Sync the file — and it's worth asking in this ticket or a follow-up whether these five Codex runtime assets should be real parity pairs, since an unenforced mirror will drift again.

packages/cli/templates/hooks/lib/re-entry.ts
```diff
@@ -38,11 +38,20 @@ const RECENT_TURNS = 10;
 export function parseLogLine(line: string): Entry | null {
   const match = LINE_REGEX.exec(line.trim());
   if (!match) return null;
+  // LINE_REGEX has four required, non-empty capture groups. The tuple records
+  // that contract for noUncheckedIndexedAccess without a dead runtime branch.
+  const [, timestamp, sessionId, ticket, nextImperative] = match as unknown as [
```
TheMostlyGreat: **Reviewer:** This resolves my round-one note about the unreachable guard, but I think it trades a safe flaw for an unsafe one.

`as unknown as [...]` is the strongest form of type assertion — it discards *all* relationship to the source type. It's accurate today: `LINE_REGEX` has four non-optional groups, so `match` is a 5-tuple of `string`. The problem is that nothing keeps it accurate.

I probed it in an isolated tsconfig (`strict`, `noUncheckedIndexedAccess`) with the only change being an optional group:

```ts
const LINE_REGEX = /^(\S+)\s+(\S+)\s+(ticket=\S+)?\s+Next:\s+(.+)$/;
//                                              ^ now optional
const [, , , ticket] = match as unknown as [string, string, string, string, string];
```

→ **tsc exits 0.** The cast reports `ticket: string` while the value is `undefined`, and `Entry.ticket` carries that undefined downstream with the compiler fully green.

So the comment on line 41 — "four required, non-empty capture groups" — is now the *only* thing binding the cast to the regex, and it's prose. The previous version was dead code but couldn't lie; this version is live code that can.

Worth weighing against how narrow the trigger is (someone editing `LINE_REGEX` to add an optional group). Cheapest binding is a `parseLogLine` unit test that would fail if a group became optional — then the invariant is checked rather than asserted. Reverting to the guard is also defensible: dead code that can't lie beats live code that can.

See also `templates/statusline/reentry.ts:60`, where this same PR keeps the dead-guard shape for the identical problem — worth landing on one answer.

packages/cli/tests/shipped-typescript-templates-validation.release.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    },
+  },
+];
+
+interface InstalledTemplate {
+  destinationPath: string;
+  templatePath: string;
+}
+
+const shippedTypeScriptTemplates: InstalledTemplate[] = [
```
TheMostlyGreat, line 55: **Reviewer:** Deriving the surface from `SAFEWORD_SCHEMA` is the right call — I verified it's exact (106 schema-declared `.ts` templates, 106 physical on disk, zero uncovered). One gap comes with it: there's no assertion that this list is non-empty, and the previous revision's accidental protection is gone.

That protection was ESLint's default `errorOnUnmatchedPattern` — passing a directory glob meant an empty match threw. Explicit paths don't get that. I checked what actually happens:

```
eslint.lintFiles([]) → RESOLVED with results, NO throw
```

It silently falls back to linting the cwd. So if a schema refactor ever emptied this list (a `.template` key rename, an `ownedFiles`/`managedFiles` restructure), the fixture would contain only `tsconfig.json` and the generated `owned-paths.ts` — and **all three tests would pass green**, having validated one generated file instead of 106 shipped templates. `tsc` doesn't catch it either: `include` still has `ownedPathsTarget`, so no TS18003.

Silent coverage loss in a release gate is the worst failure mode this file can have, since a green lane is exactly what everyone trusts at release time. One line closes it:

```ts
expect(shippedTypeScriptTemplates.length).toBeGreaterThanOrEqual(106);
```

That also pins the count the ticket and `verify.md` cite, so a future template addition or accidental removal shows up as a deliberate edit rather than a silent shift.

packages/cli/templates/statusline/reentry.ts
```diff
@@ -56,7 +56,8 @@ async function main(): Promise<void> {
 
   if (entries.length === 0) return;
 
-  const latest = entries[entries.length - 1];
+  const latest = entries.at(-1);
+  if (!latest) return;
```
TheMostlyGreat, line 58: **Reviewer:** Fixes the TS18048 I flagged last round — confirmed the gate now covers this file (I appended a type error here and the release test went red at `.safeword/statusline/reentry.ts(75,7)`).

Small consistency note: this guard is unreachable. `entries.length === 0` returns two lines above, so `entries.at(-1)` on a non-empty array is always defined — the exact shape that `re-entry.ts:43` *removed* in this same PR by switching to a type assertion.

No action needed on this line specifically; flagging because the two files now answer the same question opposite ways, and the next person to touch either will have to guess which is the house style. My preference is this one — a cheap dead guard beats an unchecked cast, and it stays honest if the code above it ever changes.

packages/cli/tests/shipped-typescript-templates-validation.release.test.ts
```diff
@@ -0,0 +1,188 @@
...
+          noFallthroughCasesInSwitch: true,
+          noImplicitOverride: true,
+          noUncheckedIndexedAccess: true,
+          noUnusedLocals: false,
+          noUnusedParameters: false,
+          skipLibCheck: true,
+          strict: true,
+          target: 'ESNext',
+          paths: {
+            '@cucumber/cucumber': [
```
TheMostlyGreat: **Reviewer:** Minor. This pins `@cucumber/cucumber`'s internal file layout (`lib/index.d.ts`) from a devDependency.

Less brittle than it first looks — I checked the package's own `package.json`, and `types` is `./lib/index.d.ts`, so this mirrors its declared entry point rather than reaching past it. And a break fails loud (TS2307 → red gate), not silent.

Still, mapping to the package *directory* and letting `package.json#types` resolve would survive a layout change on a cucumber upgrade without anyone having to trace why the release gate suddenly complains about the customer's `steps/` files. Take it or leave it.

.project/tickets/505-keep-shipped-hooks-host-lintable/ticket.md
```diff
@@ -0,0 +1,93 @@
...
+last_modified: 2026-07-29T01:00:00Z
+scope:
+  - Keep every shipped TypeScript template compatible with the supported ESLint baseline used by host projects.
+  - Add release validation for the schema-declared distributed TypeScript surface: baseline lint, typed-preset parse/config loading, and strict installed-shape TypeScript checking.
+  - Keep package-owned Codex runtime template mirrors under parity validation.
+out_of_scope:
+  - Adding a host-project lint ignore for .safeword/hooks.
+  - Normalizing the existing template corpus to every stylistic and security rule in the full typed host preset.
+done_when:
+  - Shipped TypeScript hooks have no errors from the supported ESLint baseline.
```
TheMostlyGreat: **Reviewer:** `scope:` was updated to say "template" this round, but all three `done_when` bullets still say "hook" — and one still says "without customer dependencies," which the new docblock correctly walks back to "Safeword's package-pinned type dependencies."

The acceptance criteria now describe a narrower gate than the one that was built: it validates 106 schema-declared templates including `statusline/reentry.ts` and the cucumber `steps/`, not just hooks. Since these bullets are what a future reader checks the work against, worth bringing them along with the scope change.

(Moving the ticket back to `phase: verify` / `status: in_progress` for the second round was the right call — noting it so it doesn't look unintentional.)

TheMostlyGreat (commented):
## Reviewer: re-review of `a12b03d` — full `/quality-review` + engineering review

_Posted as `COMMENT` — `gh` authenticates as the PR author, so `APPROVE` 422s. **No merge block either way**; verdict is advisory._

**Verdict: APPROVE** — both blockers from the previous round are fixed, and fixed at the structural level rather than patched. Two new one-line robustness suggestions below; neither is a live defect.

The response to round one went further than what I flagged, in the right direction twice: instead of hand-syncing the desynced mirror you made the mirror class *unable* to drift, and instead of widening a glob you derived the surface from `SAFEWORD_SCHEMA` so it can't fall behind. Both are the durable version of the fix.

---

## Blocker 1 — codex mirror drift: **fixed, structurally**

Not just the file — the schema now carries a `dogfoodParity?: boolean` opt-in (`packs/types.ts:152`), `parityPairs()` folds opted-in `managedFiles` into the pair walk (`parity.ts:32`), and both `runParity` and `syncParityPairs` consume it. Verified:

```
# all five codex mirrors byte-identical
post-tool-quality.ts OK   post-tool-skill-nudge.ts OK   pre-tool-quality.ts OK
pre-tool-quality-helpers.ts OK   stop.ts OK

bun scripts/parity-check.ts --mode=all
# → All 200 pairs and 8 contracts in sync   (was 195 — the five codex assets are now enforced)
```

`tests/parity.test.ts` covers both directions (detection + `--fix` sync) with real fixtures — 27/27 pass. And the flag is correctly scoped: these stay `managedFiles` so they're still never installed into customer repos, which is the distinction that made this tricky in the first place.

Worth noting the new enforcement immediately earned its keep — `.safeword/hooks/codex/post-tool-skill-nudge.ts` came in at +29/-9 with no template counterpart, i.e. it was a *pre-existing* stale mirror that nothing had been checking. That's a second latent drift found by the fix.

## Blocker 2 — shipped TS outside `templates/hooks/**`: **fixed, and I confirmed the gate now bites there**

`shippedTypeScriptTemplates` derives from `ownedFiles` + `managedFiles` rather than a glob. Coverage is exact:

```
schema-declared .ts templates: 106
physical .ts under templates/: 106
on disk but NOT covered:       none
```

Mutation test against the file I flagged last round — appended a type error to `templates/statusline/reentry.ts`:

```
.safeword/statusline/reentry.ts(75,7): error TS2322: Type 'string' is not assignable to type 'number'.
Tests  1 failed | 2 passed (3)
```

So the scope expansion is real, not nominal. The original TS18048 at `statusline/reentry.ts:59` is also fixed. `steps/*.ts` are pulled in via the `paths` mapping for `@cucumber/cucumber`.

## Round-one secondary findings — all addressed

| # | finding | status |
|---|---|---|
| 3 | trailing `import eslintJs` on line 152 | ✓ moved to line 15 |
| 4 | test 2 name overclaimed a `fatal`-only assertion | ✓ renamed to `parses and resolves type information without fatal errors…` |
| 5 | `owned-paths.ts` never reached the host baseline | ✓ test 1 now lints the fixture |
| 6 | inscrutable ENOENT on `tscPath` | ✓ `expect(result.error).toBeUndefined()` added |
| 7 | `@types` version skew unnamed | ✓ docblock now states it explicitly |
| 8 | was `resolveClaudeSessionId` needed? *(I flagged this unverified)* | ✓ removed — it wasn't; env projection inlined at both call sites |
| 9 | unreachable guard in `re-entry.ts` | ⚠️ replaced with a cast — see the note below |

On #8: I couldn't confirm that one last round because the experiment was blocked, so I raised it as a question rather than a finding. Good call resolving it — the 8 lines of duplication at the two call sites are the cheaper trade against a permanent public export on a shipped lib.

---

## New this round — two one-liners worth taking before merge

Both are inline. Summary:

**`re-entry.ts:43` — the cast is unsound under regex edits.** I probed it in an isolated tsconfig: the identical `as unknown as [string, string, string, string, string]` typechecks **clean** when `LINE_REGEX` gains an optional group. So the fix trades dead-but-safe code for live-but-unchecked code, and nothing binds the comment's contract to the regex.

**`shipped-hooks-validation.release.test.ts:50` — new vacuity risk.** I verified `eslint.lintFiles([])` does *not* throw; it falls back to linting the cwd. With no size assertion on the derived list, a schema refactor that emptied it would leave all three tests green while validating one generated file instead of 106 templates. The previous revision was accidentally protected by `errorOnUnmatchedPattern` on a directory glob; deriving explicit paths removed that safety net.

Also minor: `statusline/reentry.ts:60` keeps an unreachable guard while `re-entry.ts:43` eliminates one via a cast — same problem, opposite answers, same PR. And the ticket's three `done_when` bullets still say "hook" while `scope:` was updated to "template".

---

## Verification run this session

| check | result |
|---|---|
| `test:release` (this file) | ✓ 3/3, 19s |
| `tests/parity.test.ts` | ✓ 27/27 |
| `parity-check --mode=all` | ✓ 200 pairs, 8 contracts |
| schema coverage vs disk | ✓ 106/106, zero uncovered |
| mutation: statusline template | ✓ gate goes red |
| `eslint` on the new test file | ✓ clean |
| `knip` | ✓ clean |
| `@types/bun@1.3.14` | ✓ npm-latest, matches installed `bun 1.3.14` |

**Scorecard —** Currency ✓ · Sources ✓ · Correct ✓ · Elegant ⚠️ (one unchecked cast) · No-bloat ✓ (the speculative export came back out) · Wiring ✓ (real ESLint API, real preset, real `tsc` subprocess, real schema; no internal-seam mocks)

**Next:** add the file-count assertion at `:50` and decide the `re-entry.ts:43` cast — both one-liners. Neither blocks; everything else here is ready to merge.

packages/cli/templates/hooks/lib/re-entry.ts
TheMostlyGreat: Resolved in b0a63c51a: `parseLogLine` now destructures normally and fails closed when any capture is missing. Added direct regression coverage for a missing imperative, so a future optional capture remains silent rather than leaking `undefined`.

packages/cli/tests/shipped-typescript-templates-validation.release.test.ts
TheMostlyGreat: Resolved in b0a63c51a: the release gate now requires at least 106 schema-declared TypeScript templates before linting. This prevents ESLint’s empty-input fallback from producing a vacuous green result while permitting intentional additions.

packages/cli/templates/statusline/reentry.ts
TheMostlyGreat: Resolved in b0a63c51a: removed the redundant `entries.length` condition. The single `entries.at(-1)` absent-value guard is behavior-preserving and matches the parser’s fail-closed pattern.

packages/cli/tests/shipped-typescript-templates-validation.release.test.ts
TheMostlyGreat: Resolved in b0a63c51a: changed the Cucumber path mapping to the pinned package root. Under bundler resolution TypeScript now reads the package-declared types, avoiding a hard-coded `lib/index.d.ts` layout assumption; the release typecheck passes.

.project/tickets/505-keep-shipped-hooks-host-lintable/ticket.md
TheMostlyGreat: Resolved in b0a63c51a: updated all acceptance criteria to the schema-declared shipped TypeScript template surface and made the fixture dependency contract explicit.

packages/cli/tests/hooks/re-entry.test.ts
```diff
@@ -0,0 +1,20 @@
...
+      parseLogLine('2026-07-29T12:34:56Z sess_123 ticket=505/implement Next: run verification'),
+    ).toEqual({
+      timestamp: '2026-07-29T12:34:56Z',
+      sessionId: 'sess_123',
+      ticket: 'ticket=505/implement',
+      nextImperative: 'run verification',
+    });
+  });
+
+  it('rejects a line with a missing imperative', () => {
```
TheMostlyGreat: **Reviewer:** These two cases don't exercise the guard they were added to protect, and I'd rather that be written down than discovered later.

Case two never reaches the guard. `LINE_REGEX` requires `Next:\s+(.+)$`, so this input fails the regex outright and `parseLogLine` returns at `if (!match)` — line 41, not 42. Verified directly:

```
regex match for '…ticket=505/implement Next:' → NULL → returns at !match, guard NOT reached
```

And neither case would catch the regression the guard exists for. I re-ran both against a regex with an optional group — `(ticket=\S+)?`:

```
test1 (canonical line) would still pass: true
test2 (missing imperative → null) would still pass: true
```

So if someone later makes a capture optional, these tests stay green. What actually saves you is the guard itself — it returns `null` instead of emitting `undefined` in a `string` field. That's the right protection and it works; the tests just don't prove it.

This isn't fixable by writing a better test, incidentally: every quantifier in the current regex is `+`, so no input can reach the guard. It's untestable by construction, which is exactly what the line 42 comment claims. The only gap is that this test's name — "rejects a line with a missing imperative" — reads as if it covers that path.

Cheapest fix is a rename (`rejects a line the canonical format does not match`) or one line noting that the guard is defense against future regex edits and is unreachable today. Either way the reader knows the guard is deliberate rather than assuming it has coverage.

scripts/parity-check.ts
```diff
@@ -65,6 +64,7 @@ const result = runParity({
 
 if (result.failures.length === 0) {
   if (mode === 'all') {
+    const pairCount = result.passedCount - contractCount;
```
TheMostlyGreat, line 69: **Reviewer:** Correct, and I like it better than what it replaced — but it's derived arithmetic on an aggregate, so worth one line of thought before it settles in.

I traced `runParity`: `passedCount` increments at exactly two sites — pairs (only under `mode === 'all'`) and contracts. The orphan-template and cursor-rule scans push into `failures` and never touch it. This line runs only inside the zero-failures branch, so `passedCount === pairs + contracts` exactly, and the subtraction is sound.

It's also the more honest number. The old version recomputed from the schema, which reported what was *declared*; this reports what was actually *walked*, so a bug that silently skipped pairs would now show up in the output instead of being masked.

The forward-looking fragility: any future check that also increments `passedCount` — a third category, a new scan that counts successes — would silently inflate the pair count with no test failing. Returning a structured result (`{ pairsPassed, contractsPassed }`) rather than one total would make it immune. Low priority; just noting the coupling now exists between a display string and an internal counter's composition.

packages/cli/tests/shipped-typescript-templates-validation.release.test.ts
```diff
@@ -0,0 +1,211 @@
...
+  ...Object.entries(SAFEWORD_SCHEMA.ownedFiles),
+  ...Object.entries(SAFEWORD_SCHEMA.managedFiles),
+].flatMap(([destinationPath, definition]) => {
+  const templatePath = definition.template;
+  return templatePath?.endsWith('.ts') ? [{ destinationPath, templatePath }] : [];
+});
+
+// This minimum blocks a vacuous ESLint run if schema discovery regresses while
+// allowing deliberate additions to the distributed TypeScript surface.
+const MIN_SHIPPED_TYPESCRIPT_TEMPLATE_COUNT = 106;
```
TheMostlyGreat, line 66: **Reviewer:** This closes the vacuity hole from round two — confirmed, and putting the assertion in its own leading test is the right shape, since it fails before the three expensive passes run.

One property worth naming in the comment: `toBeGreaterThanOrEqual` makes this a **floor, not a census**. It catches the failure mode that mattered (derivation collapses to zero or near-zero), and it correctly stays quiet on deliberate additions. What it can't see is an accidental *removal* offset by an addition — net count holds at 106, one template silently leaves the release surface.

That's the right trade. Exact equality would churn on every legitimate template addition, and this gate should not be a thing people edit reflexively to make green. Just don't let the number's presence imply the surface is pinned — it isn't, and the comment above currently reads as though the count is the contract rather than a lower bound.

Worth a follow-up thought rather than a change here: parity's orphan-template scan already knows about every physical template. If it ever grew a "schema-declared `.ts` templates == physical `.ts` templates" assertion, that would be the real census, and this constant could go back to being a cheap smoke check.

TheMostlyGreat (commented):
## Reviewer: third pass on `b0a63c5` — full `/quality-review` + engineering review

_Posted as `COMMENT` — `gh` authenticates as the PR author, so `APPROVE` 422s. **No merge block**; verdict is advisory._

**Verdict: APPROVE.** Every round-two finding is resolved, and the two judgement calls went the way I'd have argued for. Three notes below, all optional — none changes behavior or coverage.

---

## Round-two findings — resolved

**`re-entry.ts:42` — the unsound cast is gone.** Back to destructuring plus a runtime guard, but reframed:

```ts
// Keep this fail-closed even if a later regex edit makes a capture optional.
if (!timestamp || !sessionId || !ticket || !nextImperative) return null;
```

That comment is the part that matters. My round-one note called this branch dead code; it *is* unreachable today, but the new framing states it as forward-looking fail-closed defense, which makes it correct-by-intent rather than leftover. Better than either previous version — the v1 guard was safe but unexplained, the v2 cast was explained but could lie.

**Vacuity closed** — `MIN_SHIPPED_TYPESCRIPT_TEMPLATE_COUNT = 106` with its own leading test. Since it runs first, an emptied schema derivation now fails loudly instead of quietly linting one generated file.

**Consistency settled** — `statusline/reentry.ts` dropped its redundant `entries.length === 0` check, so both files now use the same `at(-1)` + guard shape. Picking the guard over the cast in both places is the answer I'd have argued for.

**Cucumber `paths`** now maps to the package directory so `package.json#types` resolves. **Ticket `done_when`** updated to "schema-declared shipped TypeScript templates," and the stale "without customer dependencies" phrasing corrected to match the docblock.

## One correction to my own round-two review

I marked round-one finding #5 — *the generated `owned-paths.ts` never reaches the host baseline* — as fixed. It was only half fixed. In `a12b03d`, `templatePaths` didn't include `ownedPathsTarget`, so `tsc` covered the generated module via `include` while both ESLint passes still skipped it. Commit `6f89795` (`templatePaths.push(ownedPathsTarget)`) is what actually closed it. Caught on your side, not mine — flagging because it changes what the gate really covers, and the ticket's new RED/GREEN pair records it accurately.

---

## New this round — three notes, none blocking

**The new `re-entry.test.ts` doesn't bind the invariant I asked it to bind.** I verified by re-running both cases against a regex with an optional group — both still pass, and case two never reaches the guard at all. Detail inline. The runtime protection is real and correct; the tests just don't demonstrate it, and the second test's name implies they do.

The other two are `parity-check.ts:67` (derived arithmetic on an aggregate — correct today, verified) and the `MIN_` constant being a floor rather than a census. Both inline, both take-it-or-leave-it.

---

## Verification run this session

| check | result |
|---|---|
| full release suite (`vitest.release.config.ts`) | ✓ 8 files / 26 tests |
| `tests/hooks/re-entry.test.ts` + `tests/parity.test.ts` | ✓ 2 files / 29 tests |
| `parity-check --mode=all` | ✓ 200 pairs, 8 contracts |
| `eslint` on the renamed release test + new unit test | ✓ clean |
| `knip` | ✓ clean |
| optional-group probe against the new tests | ⚠️ both survive the regression — see inline |
| `passedCount` arithmetic in `runParity` | ✓ traced; exactly two increment sites, subtraction is sound |

**One inconclusive run, reported as such:** I tried a mutation test to prove the *ESLint baseline* (not just `tsc`) reaches non-hook templates. It was SIGTERM'd at 7 minutes — a parallel Codex worktree held the machine-global build lock, so my run was queued, not hung. That's an environment result, not a signal about this PR, and I'm not claiming a verdict from it. The `tsc` half of that question was already settled in round two, where a type error injected into `templates/statusline/reentry.ts` turned the gate red. The mutated file was restored and the tree verified clean.

**Scorecard —** Currency ✓ · Sources ✓ · Correct ✓ · Elegant ✓ · No-bloat ✓ · Wiring ✓

Three rounds in, the shape of the response has been consistently to fix the class rather than the instance — schema-level parity opt-in instead of a file copy, schema-derived surface instead of a wider glob, a stated invariant instead of a silenced diagnostic. The ticket work log tracks all of it accurately.

**Next:** merge. The three notes are optional and none needs a re-review.

packages/cli/tests/hooks/re-entry.test.ts
TheMostlyGreat: Resolved in c71745453: renamed the case to describe the non-matching canonical format and added a comment that it returns before capture validation. The forward-compatible guard remains deliberately untested by the current required-capture regex.

scripts/parity-check.ts
TheMostlyGreat: Resolved in c71745453 after a figure-it-out review: retained the proven aggregate because there is no third successful category today, and documented the exact two-category invariant beside the subtraction. A structured result would be the right change only when that composition expands.

packages/cli/tests/shipped-typescript-templates-validation.release.test.ts
TheMostlyGreat: Resolved in c71745453: the comment now explicitly calls 106 a floor rather than a census and names schema/parity validation as the check responsible for removals offset by additions.

Issue #29, opened 2025-12-08
Review ESLint rules for LLM-optimized coding


TheMostlyGreat:
## Context

Safeword bundles a comprehensive ESLint configuration designed for LLM-assisted development. We should review all linters and rules to ensure they are optimally configured for LLM-generated code patterns.

### ESLint Plugins Installed

**Base (always installed):**
- `@eslint/js` - Core JS rules
- `eslint-plugin-import-x` - Import/export validation
- `eslint-import-resolver-typescript` - TS path resolution
- `eslint-plugin-sonarjs` - Code smells, bugs
- `eslint-plugin-unicorn` - Modern JS patterns
- `eslint-plugin-boundaries` - Architecture boundaries
- `eslint-plugin-playwright` - E2E test rules
- `@microsoft/eslint-plugin-sdl` - Security rules
- `eslint-config-prettier` - Prettier conflict resolver

**Conditional (framework-dependent):**
- `typescript-eslint` - TypeScript (now using `stylisticTypeChecked`)
- `eslint-plugin-react` + `eslint-plugin-react-hooks` + `eslint-plugin-jsx-a11y` - React
- `@next/eslint-plugin-next` - Next.js
- `eslint-plugin-astro` - Astro
- `eslint-plugin-vue` - Vue
- `eslint-plugin-svelte` - Svelte
- `@vitest/eslint-plugin` - Vitest

### Other Linters
- `prettier` - Code formatting
- `markdownlint-cli2` - Markdown linting
- `knip` - Dead code detection

## Review Goals

1. **LLM Pattern Alignment** - Ensure rules match how LLMs actually generate code
2. **Noise Reduction** - Disable rules that create false positives on LLM output
3. **Safety Focus** - Prioritize rules that catch real bugs over style preferences
4. **Auto-fix Priority** - Prefer rules that auto-fix over manual-fix-only rules
5. **Performance** - Consider linting speed impact on hook feedback loops

### TypeScript ESLint
- [ ] Review `stylisticTypeChecked` rules - are any too strict for LLM code?
- [ ] Consider adding explicit any warnings (currently allowed)
- [ ] Evaluate type assertion rules

### Unicorn
Current overrides (in `config.ts:84-97`):
```js
"unicorn/prevent-abbreviations": "off", // ctx, dir, pkg, err are standard
"unicorn/no-null": "off", // null is valid JS
"unicorn/no-process-exit": "off", // CLI apps use process.exit
// ... etc
```
[...]

PR #1614, opened 2026-07-28
Improve architecture narrative purpose seeding

TheMostlyGreat:
## Summary

- Remove the package-by-package narrative drift advisory; decision records no longer need to duplicate generated package inventories.
- Seed generated module purposes from leading documentation comments and package purposes from manifest descriptions.
- Preserve human edits while migrating legacy placeholders, refreshing changed metadata, and safely handling packages without source modules.

## Why

Architecture decision records should capture meaningful choices, while generated documents own structural inventory. Metadata makes newly generated architecture documents useful immediately without overriding human prose.

Fixes #1608
Fixes #1609

## Validation

- `bun run lint`
- `bun run --cwd packages/cli build`
- Focused architecture Vitest suite


packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +291,34 @@ function jsModuleName(filename: string): string {
...
+    : { name, path, purpose, seededPurpose: true };
+}
+
+/**
+ * The first sentence from a file's leading JSDoc comment. This deliberately
+ * reads only a comment at the start of the file (apart from whitespace/BOM),
+ * never arbitrary source comments, so the generated purpose is deterministic
+ * and visibly intentional rather than a guess about implementation details.
+ */
+function purposeFromLeadingDocumentComment(path: string): string | undefined {
```
TheMostlyGreat: Reviewer: **[blocking] The first-sentence heuristic misfires on four common comment shapes.** Verified by probe against this branch:

| Source comment | Generated purpose |
|---|---|
| `/**\n * Reconciliation Engine\n *\n * Computes and executes plans.\n */` | `Reconciliation Engine Computes and executes plans.` |
| `/** Handles retries, e.g. on network failure, with backoff. */` | `Handles retries, e.g.` |
| `/**\n * Copyright (c) 2026 Acme Inc.\n * SPDX-License-Identifier: MIT\n */` | `Copyright (c) 2026 Acme Inc.` |
| `/** <600 A's>. */` | 601-char "one-line purpose" |

Root causes, in order of severity:

1. **Blank lines are erased before sentence-splitting.** `.filter(Boolean)` drops the paragraph break, so a title line with no terminal punctuation runs straight into the next paragraph. This is the dominant JSDoc file-header convention and it is already in this PR's own output — see `packages/cli/architecture.generated.md` L130 and L150.
2. **`(?:\s|$)` after `[.!?]` splits on any abbreviation** — `e.g.`, `i.e.`, `etc.`, `vs.`, `Inc.`, `No.`, initials.
3. **A leading license/SPDX banner is indistinguishable from a purpose.** safeword runs against arbitrary customer repos where a banner is the *first* block comment in every file — every module's "purpose" becomes the copyright line, in a document the SessionStart hook loads as ground truth for every agent session.
4. **No length cap**, despite `SkeletonNode.purpose` being documented as "One-line purpose".

Suggested shape: split the comment body on blank lines first and take the first paragraph; if that paragraph is a single line with no terminal punctuation, use it whole as the title; then apply the sentence split within it. Add a bail-out when the first paragraph matches a banner pattern (`Copyright`, `SPDX-`, `@license`, `Licensed under`) and fall through to the next paragraph or the placeholder. Cap at ~200 chars.

Also silently missed (no seed at all) because `^\s*` can't skip them: a `#!/usr/bin/env node` shebang, and a leading `// eslint-disable` line comment. Shebang files are CLI entry points — exactly the modules most worth describing.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -323,7 +357,14 @@ function rustModuleNodes(projectDirectory: string): SkeletonNode[] {
     }
     const name = entry.name.slice(0, -'.rs'.length);
     if (!byName.has(name)) {
-      byName.set(name, { name, path: `src/${entry.name}`, purpose: PURPOSE_PLACEHOLDER });
+      byName.set(
+        name,
+        sourceFileNode(
```
TheMostlyGreat, line 572: Reviewer: **`sourceFileNode` is wired into the Rust and Python extractors, but the regex only matches `/** */`.**

Rust doc comments are `//!` (inner) and `///` (outer); Python module docs are `"""…"""`. Neither can ever match `/^\s*\/\*\*/`. Probe against this branch, `src/engine.rs` with `//! The Rust engine module.` → `No description yet — awaiting prose.`

So for Rust and Python this call site is dead wiring that still pays a full `readFileSafe` per module on every heal. Either teach `purposeFromLeadingDocumentComment` the `//!`/`///` and `"""` forms (small, and it's what makes the feature actually deliver on #1608's "existing project metadata" for non-JS hosts), or keep `{ name, path, purpose: PURPOSE_PLACEHOLDER }` here and at `pythonModulesFrom` until it does. Same call at L435.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -192,7 +194,7 @@ function enumerateJsSourceRoot(
       ]),
   );
   const excludeColocatedTests = true;
-  for (const node of jsFileNodes(entries, pathFor, excludeColocatedTests)) {
+  for (const node of jsFileNodes(entries, pathFor, excludeColocatedTests, directory)) {
```
TheMostlyGreat, line 206: Reviewer: **Directory nodes are never seedable.** The `byName` map a few lines up hard-codes `purpose: PURPOSE_PLACEHOLDER` for every directory entry; only loose files reach `sourceFileNode`.

On this repo that's the difference between a feature that fires and one that mostly doesn't: `packages/cli/architecture.generated.md` has 26 sections, 16 of them directories. After this PR, 5 are seeded and 21 still read `No description yet` — and 2 of the 5 are the malformed ones from the heuristic finding. Net clean gain on the flagship dogfood repo: 3 of 26 sections.

A directory's `index.ts` (or `mod.rs` / `__init__.py`) leading comment is the obvious source and would roughly quadruple coverage here for a few lines. Worth doing in this PR — without it the cost/benefit on ~200 lines of arbitration machinery is hard to defend.

TheMostlyGreat (commented):
Reviewer: **/quality-review + engineering review** — posted as `COMMENT` (gh authenticates as the PR author, so this carries no merge block).

## Quality Review

**Currency:** ✓ — no new ecosystem surface. Only Node stdlib (`node:crypto` `createHash`, `node:fs`, `node:path`) and language built-ins already used in this file set (`replaceAll` with `/g`, `toSorted`). Nothing deprecated, nothing with a CVE story. (verified: source read this session; no new deps in the diff.)
**Sources:** ✓ — the design claims trace to #1608's scope statement and to the JT852Q/AXRC4D prose-preservation contract already encoded in `architecture-document.ts`. (verified: `gh issue view 1608`, ticket `BY7RNR` spec.)
**Correct:** ⚠️ — the preservation machinery is sound and I could not break it (see below). The *extraction* heuristic is not: it produces wrong or ungrammatical purposes on four common comment shapes, two of which already shipped in this PR's own regenerated doc.
**Elegant:** ⚠️ — human-vs-generated arbitration via content digest is the right mechanism; the 64-hex-per-section marker and the near-duplicate `parseSectionProse` / `parseRootPackageProse` pair are heavier than the job needs.
**No-bloat:** ⚠️ — on the flagship dogfood repo the feature moves **5 of 26** sections off the placeholder, and **2 of those 5 are grammatically broken**. Net clean gain: 3/26. That is a thin return for ~200 lines of new arbitration logic.
**Wiring (code only):** ✓ — `selfHealProject` is exercised end-to-end over a real temp monorepo with real fs (`architecture-document.test.ts`, the new `GitHub #1608` describe block, 8 cases). Real collaborators, only the fs boundary is real-not-mocked. That's a genuine wiring test, and it's the strongest part of this PR.

**Verdict: REQUEST CHANGES** — for finding 1 (blocking) and finding 6 (needs an answer, not necessarily a change).

### What I verified empirically

I built the branch and ran a throwaway probe suite against the real extractor (deleted afterward). Confirmed working: idempotence (`created` → `unchanged`, byte-identical second write); human-prose precedence, including the hard case where a human edits the prose **and** the source comment changes in the same interval (correctly stays `unchanged`, human text wins); placeholder→seed upgrade on legacy docs; seed removal falling back to the placeholder rather than stranding stale prose. CI is green on all four checks.

Confirmed broken — actual probe output:

| Source comment | Generated purpose |
|---|---|
| `/**\n * Reconciliation Engine\n *\n * Computes and executes plans.\n */` | `Reconciliation Engine Computes and executes plans.` |
| `/** Handles retries, e.g. on network failure, with backoff. */` | `Handles retries, e.g.` |
| `/**\n * Copyright (c) 2026 Acme Inc.\n * SPDX-License-Identifier: MIT\n */` | `Copyright (c) 2026 Acme Inc.` |
| `/** <600 A's>. */` | 601-char "one-line purpose" |

### Findings

1. **The first-sentence heuristic misfires on four common comment shapes** (blocking) — inline on `architecture-skeleton.ts`. Two are already in this PR's regenerated `packages/cli/architecture.generated.md`.
2. **`sourceFileNode` is wired into the Rust and Python extractors but only matches `/** */`** — dead wiring plus a wasted file read per module. Rust confirmed by probe.
3. **Directory nodes are never seedable** — 16 of the dogfood doc's 26 sections structurally cannot benefit.
4. **Any leading-doc-comment edit now flips `architecture --check` to non-zero.**
5. **64 hex chars of marker per section** in a human-read document.
6. **Scope: the narrative-drift advisory removal isn't in #1608** and reverts the shipped fix for closed issue #848 with no linked decision.
7. **The feature file now has zero executable scenarios and a fully orphaned 228-line steps file**, and `BY7RNR/spec.md` still declares the deleted ACs.
8. Minor: default param before a required one in `jsFileNodes`.

**Next:** fix finding 1 (paragraph-first extraction + banner guard + length cap), regenerate both `architecture.generated.md` files, and answer finding 6 — if the advisory removal is intended, split it into its own PR that reopens or explicitly closes out #848 and cleans up the steps file, spec ACs, and feature title from finding 7.

packages/cli/src/utils/architecture-document.ts
```diff
@@ -263,7 +272,27 @@ function hasMatchingNodePaths(content: string, nodes: SkeletonNode[]): boolean {
...
-  return paths.size === nodes.length && nodes.every(node => paths.get(node.name) === node.path);
+  return (
+    paths.size === nodes.length &&
+    nodes.every(node => paths.get(node.name) === node.path) &&
+    hasMatchingGeneratedPurposes(nodes, parseSectionProse(content))
+  );
+}
+
+/** Whether no generated purpose has changed since this document was written. */
+function hasMatchingGeneratedPurposes(
```
TheMostlyGreat, line 286: Reviewer: **This makes any leading-doc-comment edit a would-change action.** `matchesExisting` returning false → `decideAction` returns `'healed'` → `isWouldChangeAction` is true → `safeword architecture --check` exits non-zero.

Confirmed by probe: change only a module's leading JSDoc first sentence, nothing structural, and `selfHealProject` reports `['healed']`.

I think this is intended ("refreshing changed metadata" in the PR body) and `--stage` auto-fixes at commit time, so the practical blast radius is small. But it's a real widening of what "the architecture doc is stale" means — previously structural, now prose-of-any-source-comment — and it isn't called out in the PR description or in the `hasMatchingNodePaths` doc comment above, which still explains itself purely in terms of "path-only drift". Please say so in the comment, so the next reader doesn't treat a doc-comment-triggered CI failure as a bug.

packages/cli/src/utils/architecture-document.ts
```diff
@@ -302,20 +331,57 @@ function normalizeLineEndings(content: string): string {
...
+interface PriorPurpose {
+  text: string;
+  /** True only when this exact text is still generated metadata, not human prose. */
+  generated: boolean;
+}
+
+const SEEDED_PURPOSE_PREFIX = '<!-- seeded-purpose: ';
+
+function seededPurposeMarker(purpose: string): string {
+  return `${SEEDED_PURPOSE_PREFIX}${purposeDigest(purpose)} -->`;
```
TheMostlyGreat, line 346: Reviewer: **The digest mechanism is right; the encoding is heavier than it needs to be.**

Using content-addressed identity to tell "still generator-owned" from "a human edited this" is the correct call — a boolean marker genuinely cannot distinguish those, and I couldn't find a way to fool it (including the tricky case where a human edits the prose *and* the source comment changes in the same interval — correctly resolves to the human text).

Two cheaper encodings worth considering:

- **Truncate to 12 hex.** Full sha256 is 64 chars of noise per section in a document humans are meant to read; a 48-bit prefix is far past sufficient for accidental-collision purposes here, and there is no adversary.
- **Fold it into the existing marker** — `<!-- reconciled: <fp> seed:<12hex> -->` — instead of a second comment block plus two blank lines per section. That also removes the `SEEDED_PURPOSE_PREFIX` skip branch now duplicated in `accumulateProseLine` and `parseRootPackageProse`.

Related: `parseSectionProse` and `parseRootPackageProse` are now near-duplicates that differ only in the section-entry condition and the code-reference line. A shared parser taking a "does this section carry a code-ref line" flag would drop ~35 lines and remove the risk of the two skip-lists drifting apart.

Minor: after a human edits prose, the stale seed marker (old digest) stays in the file until the *next* heal, sitting above human-authored text as if it still owned it. Harmless — the digest mismatch is what's load-bearing — but confusing to a reader.

packages/cli/src/commands/architecture.ts
```diff
@@ -49,24 +48,9 @@ export function architecture(
     success(`Architecture state document ${result.action}: ${result.path}`);
   }
   warnUnreadableWorkspaces(cwd);
```
TheMostlyGreat, line 50: Reviewer: **Scope — this removal isn't in #1608, and it reverts a shipped fix for a user-reported issue.**

#1608's scope is four bullets, all about seeding purposes. Removing the narrative-drift advisory is none of them. It undoes the TB2 half of ticket `BY7RNR`, which shipped specifically to close #848 (CLOSED 2026-07-05) — an issue filed by an ArcadeAI/monorepo maintainer after a real incident where six generated packages, two whole product clusters, were absent from the narrative and a human found it by hand-diffing.

The PR's rationale ("decision records no longer need to duplicate generated package inventories") is a reasonable position and may well be the right call — the advisory was a deliberate, documented deviation from AXRC4D's "no deterministic drift module" ruling, and reversing a deviation is legitimate. My objection is procedural, not directional:

- No issue or ticket authorizes it, so there's no record of the decision anywhere except this PR body.
- #848 stays closed while its fix is removed, so the original reporter's problem silently returns with no trace.
- Bundling it with the seeding work means neither change can be reverted independently.

Ask: split this into its own PR that reopens #848 (or closes it wontfix with the reasoning), records the reversal in `BY7RNR`'s Decisions, and carries the cleanup in my comment on the feature file. If you'd rather keep it here, at minimum reference #848 in the PR body and update the ticket.

features/architecture-narrative-blindspots.feature
```diff
@@ -2,20 +2,15 @@
 # resolution) are @wip like bash-ledger-write-gate: their executable backing is
 # Vitest hook coverage — git-backed integration tests over the standalone hook
 # helper plus a differential parity test against the CLI resolver — and cucumber
-# steps would duplicate that harness without adding confidence. The TB2 rules
-# (CLI drift advisory) run in this lane against the real `safeword architecture`
-# command, like architecture-unreadable-workspace.
+# steps would duplicate that harness without adding confidence.
 @architecture-narrative-blindspots
 Feature: Architecture narrative reconciliation reaches configured narratives and pre-existing drift
```
TheMostlyGreat: Reviewer: **This leaves the feature and its steps file in an inconsistent state.** Four concrete items:

1. **Zero executable scenarios remain.** All 13 surviving scenarios are `@wip`. The header comment still reads "The TB1 rules … are `@wip`", implying TB2 is the live lane — but TB2 is gone, so the whole feature is now `@wip` and nothing here runs.
2. **`steps/architecture-narrative-blindspots.steps.ts` is fully orphaned** — 228 lines, ~21 step definitions, zero live scenarios binding them. Cucumber doesn't fail on unused step definitions, and per the repo's own notes the root BDD lane isn't in CI, so nothing will ever flag this. It should be deleted with the scenarios.
3. **The `Feature:` title still says "and pre-existing drift"**, and the rewritten narrative paragraph still opens by framing #848 as the motivating incident — for behavior this PR deletes.
4. **`.project/tickets/BY7RNR-architecture-narrative-blindspots/spec.md` still declares TB2.AC1–AC3** as accepted criteria (`dimensions.md` and `verify.md` too — the latter attests "12 BY7RNR TB2 scenarios driving the real CLI"). The ticket is `done`, so it's out of the in-progress coverage scope and nothing will catch the drift; the spec just quietly becomes wrong.

packages/cli/architecture.generated.md
```diff
@@ -121,7 +125,9 @@ No description yet — awaiting prose.
 
 `src/reconcile.ts`
 
-No description yet — awaiting prose.
+<!-- seeded-purpose: cbbfe7505557376adb3be92a02a7fc95298d294bcd85fcb2e56fb65c3593bafc -->
+
+Reconciliation Engine Computes and executes plans based on SAFEWORD_SCHEMA and project state.
```
TheMostlyGreat: Reviewer: **Evidence for the extraction finding** — this is the heuristic's output on safeword's own source, shipped in this PR.

The source comment in `src/reconcile.ts` is a title line (`Reconciliation Engine`), a blank line, then the description. `.filter(Boolean)` erases the blank line, the title has no terminal punctuation, and the first-sentence regex swallows both — producing a run-on that reads as a typo.

L150 is the same defect and worse: `SAFEWORD Schema - Single Source of Truth All files, directories, configurations, and packages managed by safeword are defined here.`

So 2 of the 5 sections this PR seeds in this file are ungrammatical. Both should read correctly once the extractor splits on paragraphs before sentences — please regenerate this file and `.project/architecture.generated.md` after fixing.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -230,6 +232,7 @@ function jsFileNodes(
   entries: Dirent[],
   pathFor: (entryName: string) => string,
   excludeTests = false,
+  sourceDirectory: string,
```
TheMostlyGreat, line 244: Reviewer: Nit — `excludeTests = false` now sits before a required `sourceDirectory`, so the default can never apply; both call sites pass it explicitly (L197, L218). Either move `sourceDirectory` ahead of it or drop the default and make the flag required, so the signature stops advertising an optionality it doesn't have.

packages/cli/src/utils/architecture-monorepo.ts
```diff
@@ -383,6 +389,14 @@ function packageName(packageDirectory: string): string {
   );
 }
 
+/** A usable package.json description, normalized for an architecture purpose line. */
+function packageDescription(packageDirectory: string): string | undefined {
```
TheMostlyGreat: Reviewer: This one is clean — verbatim manifest text, whitespace-normalized, with an explicit non-empty guard, and it's the highest-signal half of the feature (a `package.json` description is written to describe the package, unlike a file header comment which is written to describe the code). Tests cover both the seed and the multiline normalization.

One small inconsistency: package purposes land without terminal punctuation (`CLI for setting up and managing safeword development environments` in `.project/architecture.generated.md`) while module purposes always end in `.`, since the sentence regex requires it. Not worth forcing a period onto someone's manifest text — just noting the docs read slightly unevenly.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,152 @@ function jsModuleName(filename: string): string {
...
+  flush();
+  const text = paragraphs.find(paragraph => !isDocumentationBanner(paragraph));
+  if (text === undefined) return undefined;
+  const firstSentence = [
+    ...new Intl.Segmenter('en', { granularity: 'sentence' }).segment(text),
+  ][0]?.segment.trim();
+  return truncatePurpose(firstSentence ?? text);
+}
+
+function isDocumentationBanner(paragraph: string): boolean {
```
TheMostlyGreat, line 508: Reviewer: **[bug] The `@license` alternative can never match.** `\b` is a boundary between a word char and a non-word char. `@` is not a word char, so `\b@license` requires the *preceding* character to be a word char — which it never is at the start of a paragraph or after a space.

Probe against this branch:

```
/**
 * @license MIT
 *
 * The real purpose here.
 */
```
→ purpose = `"@license MIT"`

The same three-line fixture with `Copyright (c) …` or `SPDX-License-Identifier: …` correctly yields `"The real purpose here."`, so only this one alternative is dead. `@license` is the canonical JSDoc license tag, so it's the one most likely to appear alone without a `Copyright` line beside it.

Fix: `/(?:\b(?:copyright|spdx-|licensed under)|@license)\b/i`, or drop the leading `\b` and rely on the alternation.

**Why it survived:** the new banner test (`skips a leading license paragraph, preserves abbreviations, and caps a generated purpose`) exercises `Copyright` and `SPDX-` in one fixture and neither of the other two alternatives. Worth one fixture per keyword — a four-way alternation where two branches are untested is exactly where a dead branch hides.

Separately, `Licensed under` **does** work, but note the false positive it implies: `/** Validates copyright headers across source files. */` is a module genuinely about licensing, and it drops to the placeholder (probe-confirmed). Acceptable trade-off as written — if you want to tighten it, anchor the keyword to the paragraph start rather than matching anywhere in it.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,152 @@ function jsModuleName(filename: string): string {
...
+  for (const rawLine of body.split(/\r?\n/)) {
+    const line = rawLine.replace(/^\s*\*\s?/, '').trim();
+    if (line.length === 0) flush();
+    else lines.push(line);
+  }
+  flush();
+  const text = paragraphs.find(paragraph => !isDocumentationBanner(paragraph));
+  if (text === undefined) return undefined;
+  const firstSentence = [
+    ...new Intl.Segmenter('en', { granularity: 'sentence' }).segment(text),
```
TheMostlyGreat: Reviewer: **`Intl.Segmenter` won't break a sentence whose successor starts with a lowercase letter.** That's UAX #29 rule SB8 — and it's the *same* rule that makes `e.g.` work correctly, so it isn't a bug you can fix without giving that back.

It's live in this PR's own output. `packages/cli/architecture.generated.md:202`:

> `Route \`ticket new\` between the local-id path and issue-first creation (KKNFZA TB1). provider:none → the local minter (today's behavior, no tracker client built).`

Two sentences, because `provider` is lowercase. This codebase writes a lot of prose that starts a sentence with an identifier (`provider:none`, `bun run …`, a filename), so it will recur.

I'd **accept this and document it** rather than add a heuristic on top — a second-guess pass over ICU's output is how you reintroduce the `e.g.` bug. One sentence in the `documentationSummary` doc comment naming the trade-off ('a period followed by a lowercase token is treated as an abbreviation, by design') is enough, so the next person doesn't file it as a defect.

One robustness note while you're here: `Intl.Segmenter` is absent on a Node built `--without-intl`, and this runs inside a SessionStart hook — so on such a runtime every architecture heal would throw rather than degrade. ICU is present here (78.3) and CI is green on Node 22 and 24, so this is theoretical; a `typeof Intl.Segmenter === 'function'` guard falling back to the old regex would make it free.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,152 @@ function jsModuleName(filename: string): string {
...
+    const line = lineEnd === -1 ? remaining : remaining.slice(0, lineEnd);
+    const trimmed = line.trim();
+    if (trimmed.length > 0 && !trimmed.startsWith('#')) break;
+    remaining = lineEnd === -1 ? '' : remaining.slice(lineEnd + 1);
+  }
+  return remaining.trimStart();
+}
+
+/** Skip the JS directives that may legally precede a file-level JSDoc block. */
+function afterLeadingJsTrivia(content: string): string {
```
TheMostlyGreat, line 404: Reviewer: The shebang and `// eslint` cases from round 1 are fixed and probe-confirmed. Three neighbours still block seeding, all safe-fail (placeholder, never wrong content):

| Before the JSDoc | Result |
|---|---|
| `/* eslint-disable */` (block form) | no seed |
| `'use client';` / `'use strict';` | no seed |
| any ordinary `// comment` or `// @ts-nocheck` | no seed |

The `'use client'` one is the one I'd actually fix: it's mandatory at the top of every client component in a Next.js app-router repo, so an entire common host shape gets nothing from this feature. `'use strict'` is the same story in older CommonJS codebases.

Both are cheap here — extend the loop to also skip a leading string-literal directive (`/^['"]use [a-z]+['"];?/`) and a `/* … */` block comment that isn't `/**`. Plain `//` comments I'd leave alone: skipping arbitrary comments to reach a JSDoc further down starts guessing about which comment is the module's, which is exactly what this function's doc comment says it won't do.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,152 @@ function jsModuleName(filename: string): string {
...
+    if (paragraph.length > 0) paragraphs.push(paragraph);
+    lines = [];
+  };
+  for (const rawLine of body.split(/\r?\n/)) {
+    const line = rawLine.replace(/^\s*\*\s?/, '').trim();
+    if (line.length === 0) flush();
+    else lines.push(line);
+  }
+  flush();
+  const text = paragraphs.find(paragraph => !isDocumentationBanner(paragraph));
```
TheMostlyGreat: Reviewer: **[bug, low probability] A purpose whose paragraph starts with `### ` is injected verbatim and becomes a phantom section in the generated document.**

Probe against this branch — `/**\n * ### Not a heading\n *\n * body\n */` on `src/a.ts`, then three consecutive heals:

```
actions=[["created"],["unchanged"],["unchanged"]]  stable=true
sections=["a", "Not a heading"]
```

Good news first: it does **not** churn. I checked this specifically because a self-healing document that parses its own output is where an infinite-rewrite loop would live, and it's clean — `hasMatchingNodePaths` requires the `<!-- reconciled: … -->` marker, which a phantom heading lacks, so it never enters the node comparison. Same for `> ⚠`, embedded `-->`, a seeded-marker lookalike, and a `---` fence: all six hostile inputs round-tripped stable.

The damage is downstream. The audit skill states that the generated doc's `### <name>` headings under `## Modules` **"ARE the project's real top-level units"** and are read as "the source of structural truth… deterministic-by-reading, not guessed." A phantom heading feeds that check a module that doesn't exist, and E003 orphan-detection compares `ARCHITECTURE.md` against exactly this list. A deterministic check with a fabricated input is worse than no check.

One-line fix in `documentationSummary`: reject a candidate paragraph matching `/^#{1,6}\s/` (skip to the next paragraph, same as the banner filter), or strip the leading hashes. Cheaper than escaping at render time, and it keeps the invariant where it belongs — the extractor is what promises a purpose is one safe line of prose.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,152 @@ function jsModuleName(filename: string): string {
   return filename.slice(0, filename.lastIndexOf('.'));
 }
 
+/** A source-file node, using its leading module documentation as an initial purpose when present. */
+function sourceFileNode(name: string, path: string, absolutePath: string): SkeletonNode {
+  return nodeFromPurpose(name, path, purposeFromLeadingDocumentation(absolutePath));
+}
+
+/** A directory module may declare its boundary purpose in a conventional entry file. */
+function directoryNode(name: string, path: string, entryPoints: string[]): SkeletonNode {
```
TheMostlyGreat, line 309: Reviewer: `.map()` runs `purposeFromLeadingDocumentation` over **all** eight `JS_SOURCE_EXTENSIONS` candidates before `.find()` picks one, so a directory whose `index.ts` matches on the first try still pays seven more `readFileSafe` calls. `extractSkeleton` runs two-to-three times per heal (fingerprint, render, and the monorepo `introspected` probe), and the heal itself runs at SessionStart — so it multiplies.

Absolute numbers are small (16 directories here → ~128 failed reads per extraction) and I wouldn't block on it, but a `for` loop with an early `return` is the same line count and reads better than map-then-find:

```ts
for (const entryPoint of entryPoints) {
  const purpose = purposeFromLeadingDocumentation(entryPoint);
  if (purpose !== undefined) return nodeFromPurpose(name, path, purpose);
}
return nodeFromPurpose(name, path, undefined);
```

The `mod.rs` and `__init__.py` call sites pass a single candidate, so they're unaffected either way.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,152 @@ function jsModuleName(filename: string): string {
...
+    ...new Intl.Segmenter('en', { granularity: 'sentence' }).segment(text),
+  ][0]?.segment.trim();
+  return truncatePurpose(firstSentence ?? text);
+}
+
+function isDocumentationBanner(paragraph: string): boolean {
+  return /\b(?:copyright|spdx-|@license|licensed under)\b/i.test(paragraph);
+}
+
+function truncatePurpose(purpose: string): string {
```
TheMostlyGreat, line 522: Reviewer: Nit — the cap works (probe-confirmed at exactly 200 with the ellipsis), but it cuts mid-word. `packages/cli/architecture.generated.md:82` ships as:

> `…via a Karpathy-style LLM Wiki index (plain markdown + grep) witho…`

`witho…` reads like corruption rather than an intentional elision. Trimming back to the last space before the cap costs one line:

```ts
const clipped = purpose.slice(0, MAXIMUM_SEEDED_PURPOSE_LENGTH - 1);
const lastSpace = clipped.lastIndexOf(' ');
return `${(lastSpace > MAXIMUM_SEEDED_PURPOSE_LENGTH / 2 ? clipped.slice(0, lastSpace) : clipped).trimEnd()}…`;
```

The `> LENGTH / 2` guard keeps the behavior sane for a 200-character single token.

packages/cli/src/utils/architecture-document.ts
```diff
@@ -302,20 +334,57 @@ function normalizeLineEndings(content: string): string {
...
-  const prose = new Map<string, string>();
+interface PriorPurpose {
+  text: string;
+  /** True only when this exact text is still generated metadata, not human prose. */
+  generated: boolean;
+}
+
+const SEEDED_PURPOSE_PREFIX = '<!-- seeded-purpose: ';
+
+function seededPurposeMarker(purpose: string): string {
```
TheMostlyGreat, line 345: Reviewer: Carrying forward the unactioned half of round 1's finding 5 — not blocking, just noting it wasn't picked up so it's a deliberate call rather than an oversight.

- **64 hex chars per section** in a document humans read. A 12-hex prefix is far past sufficient for accidental-collision purposes, and folding it into the existing marker (`<!-- reconciled: <fp> seed:<12hex> -->`) would drop the second comment block plus two blank lines from every seeded section.
- **`parseSectionProse` and `parseRootPackageProse` remain near-duplicates.** `isPurposeMetadataLine` fixed the skip-list drift risk, which was the sharper half of the concern — what's left is ~35 lines of structural duplication differing only in the section-entry condition and the code-reference line.

The mechanism itself is right and I'll say so again: digest-based arbitration is the only thing that can distinguish 'a human edited this' from 'the source comment changed', and it holds up under the adversarial cases I threw at it.

features/architecture-narrative-blindspots.feature
```diff
@@ -2,20 +2,15 @@
 # resolution) are @wip like bash-ledger-write-gate: their executable backing is
 # Vitest hook coverage — git-backed integration tests over the standalone hook
 # helper plus a differential parity test against the CLI resolver — and cucumber
-# steps would duplicate that harness without adding confidence. The TB2 rules
-# (CLI drift advisory) run in this lane against the real `safeword architecture`
-# command, like architecture-unreadable-workspace.
+# steps would duplicate that harness without adding confidence.
 @architecture-narrative-blindspots
-Feature: Architecture narrative reconciliation reaches configured narratives and pre-existing drift
+Feature: Architecture narrative reconciliation reaches configured narratives
```
TheMostlyGreat, line 7: Reviewer: Round 1's finding 7 is resolved — steps file deleted, `Feature:` title corrected, `spec.md` carries a dated supersession note, `ticket.md` scope and `done_when` updated, `dimensions.md` and `verify.md` follow. That's more thorough than I asked for; the supersession note in particular is the right shape, since it preserves delivery history instead of rewriting it.

All 13 remaining scenarios being `@wip` is now correct rather than a smell: TB1 was always Vitest-backed by design, and TB2 was the only cucumber-executable lane. Withdrawing my round-1 concern on that point.

One leftover nit: this header still reads *"The TB1 rules … are `@wip`"*, phrased to contrast with a TB2 lane that no longer exists. "Every rule here is `@wip`…" would stop implying there's a live cucumber lane in this file.

packages/cli/architecture.generated.md
```diff
@@ -121,7 +127,9 @@ No description yet — awaiting prose.
 
 `src/reconcile.ts`
 
-No description yet — awaiting prose.
+<!-- seeded-purpose: 76b396197de407a1fbd82d9825006ca749d81e0faf97ee7d8006ae6583373657 -->
```
TheMostlyGreat, line 130: Reviewer: Confirming the round-1 blocker is fixed at the point of evidence. This line was:

> `Reconciliation Engine Computes and executes plans based on SAFEWORD_SCHEMA and project state.`

and now reads `Reconciliation Engine` — the title paragraph alone, which is what the source comment actually offers as a summary. `src/schema.ts` likewise went from the run-on to `SAFEWORD Schema - Single Source of Truth`.

Across this file, seeded sections went from 5 of 26 to 15 of 26 (the directory support in round 1's finding 3 did most of that), and the two malformed ones are gone. That moves my no-bloat read from "3 of 26 clean, hard to defend" to "the machinery earns its keep."

Two lines in this regenerated file still need a pass once findings 1 and 4 land — L82 (`witho…`) and L202 (the merged two-sentence purpose). Both are cosmetic, and both resolve by regenerating.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,152 @@ function jsModuleName(filename: string): string {
...
+/** Leading JSDoc is the module-level documentation convention used by JS/TS files here. */
+function jsModuleDocumentation(content: string): string | undefined {
+  return /^\/\*\*([\s\S]*?)\*\//.exec(afterLeadingJsTrivia(withoutByteOrderMark(content)))?.[1];
+}
+
+/**
+ * Rust module documentation uses inner doc comments. Outer line and block doc
+ * comments document the following item and must not be attributed to the file module.
+ */
+function rustModuleDocumentation(content: string): string | undefined {
```
TheMostlyGreat, line 356: Reviewer: This is the part I'd point someone at as the model for the rest. Accepting `//!` and `/*!` while **rejecting** `///` is the distinction that matters — an outer doc comment documents the *following item*, not the file, so attributing it to the module would be quietly wrong in a way nobody would notice in review. Probe confirms `/// Outer doc for the next item.` yields the placeholder.

`pythonModuleDocumentation` earns the same note: it skips the `#` comments and blank lines Python permits ahead of a docstring, handles `'''` and the `r`/`u` prefixes, and correctly declines a string literal that follows `from __future__ import annotations` — which is a real Python rule (the docstring must be the *first statement*) and not one I expected to see handled.

No action. Flagging it because a reviewer who only reads diffs for defects gives no signal about what to keep doing.

TheMostlyGreat (commented):
Reviewer: **/quality-review + engineering review, round 2** on `e791696`. Posted as `COMMENT` (gh authenticates as the PR author), so this carries no merge block.

**Verdict: APPROVE with two one-line fixes.** The blocking finding from [round 1](https://github.com/ArcadeAI/safeword/pull/1614#issuecomment-5111319880) is genuinely fixed — I re-probed every case that failed before and they all pass now. Two new small bugs found, both narrow, both one-liners.

## Scorecard

| | | |
|---|---|---|
| **Currency** | ✓ | Still stdlib-only. New surface is `Intl.Segmenter` — ECMA-402, stable since Node 16, present here (ICU 78.3) and green on CI's Node 22 + 24. |
| **Sources** | ✓ | The Rust and Python conventions are applied correctly per their own language rules, including the distinctions most implementations get wrong (below). |
| **Correct** | ✓ | Round-trip stability verified across six hostile inputs — no churn in any. Two narrow defects remain (findings 1 and 4). |
| **Elegant** | ✓ | `documentationSummary` reads cleanly; `nodeFromPurpose` / `resolveModulePurpose` / `isPurposeMetadataLine` removed the duplication I flagged. |
| **No-bloat** | ✓ | Seeded sections in `packages/cli/architecture.generated.md` went 5 → 15 of 26, and the two malformed ones are now well-formed. The machinery earns its keep. |
| **Wiring** | ✓ | Nine new skeleton cases plus two new document cases, each pinning one behavior from round 1 against real fs. |

## Round 1 findings — status

| # | Finding | Status |
|---|---|---|
| 1 | Extraction heuristic misfires (blocking) | **Fixed** — paragraph-first, banner skip, `Intl.Segmenter`, 200-char cap |
| 2 | Rust/Python wiring matched only `/** */` | **Fixed** — `//!`, `/*!`, `"""`, `'''` |
| 3 | Directory nodes unseedable | **Fixed** — `index.*`, `mod.rs`, `__init__.py` |
| 4 | `--check` widening undocumented | **Fixed** — now stated in `hasMatchingNodePaths` |
| 5 | 64-hex marker + duplicate parsers | **Partial** — `isPurposeMetadataLine` extracted; digest length and the two parsers unchanged |
| 6 | Advisory removal out of scope | **Fixed** — #1609 filed, `Fixes #1609` in the body, spec supersession note, ticket scope/`done_when` updated, audit skill + E006 removed across all three mirrors with a test |
| 7 | Orphaned steps file + stale spec | **Fixed** — 228-line steps file deleted, feature title corrected, spec/dimensions/verify updated |
| 9 | Default param before required | **Fixed** |

The Rust handling deserves a specific call-out: it accepts `//!` and `/*!` (inner, module-level) and **rejects** `///` (outer, documents the next item). Probe confirms `/// Outer doc` yields the placeholder. Same care on the Python side — a string literal after `from __future__ import` is correctly not treated as a docstring. That's the detail most implementations of this get wrong.

## What I re-verified

Built `e791696` and ran two throwaway probe suites against the real extractor, deleted afterward. Every round-1 failure now passes:

| Input | Round 1 | Now |
|---|---|---|
| title line + blank + body | `Reconciliation Engine Computes and executes plans.` | `Reconciliation Engine` |
| `…retries, e.g. on network failure, with backoff.` | `Handles retries, e.g.` | full sentence, intact |
| `Copyright (c) …` / `SPDX-…` banner | banner became the purpose | banner skipped, real paragraph used |
| 600-char sentence | 601 chars | capped at 200 |
| `#!/usr/bin/env node` before JSDoc | no seed | seeded |

Also still holding: idempotence (`created` → `unchanged`, byte-identical), human-prose precedence when the source comment changes in the same interval, and — new this round — **round-trip stability on six markdown-hostile purposes** (heading-like, `> ⚠`, embedded `-->`, seeded-marker lookalike, `---` fence, control). All three heals stable, zero churn.

## New findings

1. **`@license` is unreachable in the banner filter** — one-char fix, inline on `architecture-skeleton.ts`.
2. **`Intl.Segmenter` merges a sentence whose successor starts lowercase** — inherent to UAX #29, visible in this PR's own doc. Document, don't fix.
3. **Leading-trivia skipping misses `/* eslint */` blocks, `'use client'`, and plain `//` comments** — safe-fail, but `'use client'` costs the whole Next.js app-router class.
4. **A purpose starting with `### ` injects a phantom section heading** into the generated doc — no churn, but it feeds the audit skill a fake module.
5. **`directoryNode` reads all 8 index candidates before picking one.**
6. **Carried from round 1:** 64-hex digest per section; `parseSectionProse` / `parseRootPackageProse` still near-duplicates.
7. **Nits:** mid-word truncation, banner false-positive on modules about licensing, stale feature-file header comment.

**Next:** fix findings 1 and 4 (both one-liners, both with an obvious test), regenerate the two `architecture.generated.md` files, and merge — the rest are optional.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,157 @@ function jsModuleName(filename: string): string {
...
+  for (const rawLine of body.split(/\r?\n/)) {
+    const line = rawLine.replace(/^\s*\*\s?/, '').trim();
+    if (line.length === 0) flush();
+    else lines.push(line);
+  }
+  flush();
+  return paragraphs;
+}
+
+function isDocumentationBanner(paragraph: string): boolean {
```
TheMostlyGreat, line 508: Reviewer: **Still open from round 2 — unchanged since `e791696`.** `\b` before `@` requires a word character in front of it, which never occurs at a paragraph start or after a space, so this alternative is dead.

Re-probed on `4e4f998`:

| First paragraph | Purpose |
|---|---|
| `@license MIT` (alone) | `"@license MIT"` ← banner leaks in |
| `@license MIT` + `Copyright (c) 2026 Acme.` | `"The real purpose here."` ← caught, by `copyright` |

That second row narrows the real exposure usefully: any file pairing `@license` with a `Copyright`, `SPDX-`, or `Licensed under` line in the *same paragraph* is already handled, because the regex tests the whole paragraph. What leaks is the `@license`-only header — common in minified-adjacent and library-distributed JS, and in projects that keep the copyright line in a separate `LICENSE` file.

Fix, unchanged from round 2: `/(?:\b(?:copyright|spdx-|licensed under)|@license)\b/i`.

The test gap is the part I'd fix regardless of the bug: `skips a leading license paragraph, preserves abbreviations, and caps a generated purpose` covers `Copyright` and `SPDX-` in a single fixture, leaving two of four alternatives unexercised. One fixture per keyword would have caught this before review.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,157 @@ function jsModuleName(filename: string): string {
...
+  while (true) {
+    const trimmed = remaining.trimStart();
+    if (!/^\/\/\s*eslint(?:-|\s)/.test(trimmed)) return trimmed;
+    const lineEnd = trimmed.indexOf('\n');
+    remaining = lineEnd === -1 ? '' : trimmed.slice(lineEnd + 1);
+  }
+}
+
+/** Derive one meaningful, bounded summary without treating banners as module purpose. */
+function documentationSummary(body: string): string | undefined {
```
TheMostlyGreat, line 480: Reviewer: **Still open from round 2 — a purpose whose first paragraph starts with `### ` is emitted verbatim and becomes a phantom section.**

Re-probed on `4e4f998`, `/**\n * ### Not a heading\n *\n * body\n */` on `src/a.ts`, three consecutive heals:

```
actions=[["created"],["unchanged"],["unchanged"]]
document headings=["a", "Not a heading"]
```

The stability result from round 2 holds — no churn, because `hasMatchingNodePaths` requires the `<!-- reconciled: … -->` marker that a phantom heading lacks. The cost is downstream: the audit skill treats `### <name>` headings under `## Modules` as "the project's real top-level units" and "the source of structural truth… deterministic-by-reading, not guessed," and E003 orphan-detection diffs `ARCHITECTURE.md` against that list. Feeding a deterministic check a module that doesn't exist is worse than not running it.

The fix belongs right here, in the `.find()` predicate — a candidate paragraph that opens with markdown block syntax isn't a purpose:

```ts
const text = documentationParagraphs(body).find(
  paragraph => !isDocumentationBanner(paragraph) && !/^(?:#{1,6}\s|>\s|[-*+]\s)/.test(paragraph),
);
```

That also covers the `> ⚠` and list-item cases I probed, which are stable today but equally not-a-purpose. Keeping the guard in the extractor rather than escaping at render time preserves the invariant this function already promises: what it returns is one safe line of prose.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,157 @@ function jsModuleName(filename: string): string {
...
+  const text = documentationParagraphs(body).find(paragraph => !isDocumentationBanner(paragraph));
+  if (text === undefined) return undefined;
+  const firstSentence = [
+    ...new Intl.Segmenter('en', { granularity: 'sentence' }).segment(text),
+  ][0]?.segment.trim();
+  return truncatePurpose(firstSentence ?? text);
+}
+
+/** Normalize documentation text into its non-empty paragraphs. */
+function documentationParagraphs(body: string): string[] {
```
TheMostlyGreat, line 487: Reviewer: The extraction is clean and worth keeping as-is. `documentationSummary` now reads as three plain steps — pick the first paragraph that isn't a banner, take its first sentence, bound it — and the line-accumulating loop lives on its own where it belongs.

Verified behavior-preserving: all eight round-2 probe cases return byte-identical output on `4e4f998` versus `e791696`, including the banner-skip and cap paths that route through both functions.

One small thing now that the seam exists: `documentationParagraphs` strips the JSDoc `*` gutter via `/^\s*\*\s?/`, which is a JS-comment concern living in a function named for generic documentation text. It's harmless for Rust and Python today (their extractors hand over gutter-free text, and a `*`-leading line in a docstring is vanishingly rare), but the naming invites a future caller to assume it's language-neutral. Either move the strip into `jsModuleDocumentation` or say so in the doc comment.

TheMostlyGreat (commented):
Reviewer: **/quality-review + engineering review, round 3** on `4e4f998`. Posted as `COMMENT` (gh authenticates as the PR author), so this carries no merge block.

**Verdict: APPROVE — unchanged from [round 2](https://github.com/ArcadeAI/safeword/pull/1614#issuecomment-5114366702).** Nothing new found. The two commits since round 2 are a pure refactor plus a ticket-index chore, and I verified the refactor changes no behavior. **Both round-2 findings are still open** — neither was picked up.

## What changed since round 2

| Commit | What | Assessment |
|---|---|---|
| `0a7c99a` | Extract `documentationParagraphs` out of `documentationSummary` | Clean. `documentationSummary` is now three statements and reads as what it is; the paragraph-splitting loop stands alone. Verified behavior-preserving. |
| `4e4f998` | Regenerate `.project/tickets/INDEX.md` | Correct — picks up the `BY7RNR` title/summary change from round 2's supersession work. Regenerated, not hand-edited, and the diff is exactly the two lines it should be. |

No source behavior changed, so I re-ran the round-2 probe set as a regression check rather than a fresh review. All eight cases produce byte-identical output to `e791696`:

| Case | Output | Same as round 2 |
|---|---|---|
| title + body | `Reconciliation Engine` | ✓ |
| `…retries, e.g. on network failure…` | full sentence intact | ✓ |
| `Copyright (c) …` banner | `The real purpose here.` | ✓ |
| `SPDX-…` banner | `The real purpose here.` | ✓ |
| `Licensed under …` banner | `The real purpose here.` | ✓ |
| shebang before JSDoc | `The real purpose.` | ✓ |
| 300-char sentence | capped at 200 with `…` | ✓ |
| lowercase next sentence | still merged (UAX #29 SB8) | ✓ |

Idempotence (`created` → `unchanged`, byte-identical) and human-prose precedence over a changed source comment both still hold.

## Still open from round 2

1. **`@license` is unreachable in the banner filter** — `isDocumentationBanner` is byte-identical to round 2. Re-probed: `@license MIT` as the sole first paragraph still becomes the module's purpose. Inline below with the narrowed exposure.
2. **A purpose starting with `### ` still injects a phantom section heading.** Re-probed on the current head: the rendered document's headings are `["a", "Not a heading"]`. Still no churn across three heals — the stability result holds.

Both remain one-line fixes with an obvious test. Neither is severe enough to block, which is why my verdict doesn't move — but they also didn't get harder, so if they're being deliberately deferred rather than missed, that's worth saying in the PR body so they don't evaporate at merge.

Also still cosmetic, and resolved by regenerating after either fix: `packages/cli/architecture.generated.md:82` ends `witho…` (mid-word truncation) and `:202` carries the merged two-sentence purpose.

**Next:** decide explicitly on the two open findings — fix them here (two lines plus two fixtures), or file them as a follow-up issue and merge. Everything else is ready.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,234 @@ function jsModuleName(filename: string): string {
...
+  }
+  flush();
+  return paragraphs;
+}
+
+function isPurposeParagraph(paragraph: string): boolean {
+  return !isDocumentationBanner(paragraph) && !/^(?:#{1,6}\s|>\s|[-*+]\s)/.test(paragraph);
+}
+
+function isDocumentationBanner(paragraph: string): boolean {
```
TheMostlyGreat, line 508: Reviewer: **The `^` anchor is the right fix — noting what it trades away.** Anchoring fixed both the dead `@license` branch and the `Validates copyright headers…` false positive at once, which is a better outcome than the two separate patches I proposed. Probe confirms both.

The cost is that a banner only counts when a keyword is its *first* token. Three shapes now pass through as module purposes:

| Leading paragraph | Purpose produced |
|---|---|
| `All rights reserved.`<br>`Copyright (c) 2026 Acme.` | `"All rights reserved."` |
| `(c) 2026 Acme Inc.` | `"(c) 2026 Acme Inc."` |
| `MIT License` | `"MIT License"` |

Only the first is a regression — the unanchored version caught it because `copyright` appeared on the paragraph's second line. The other two were never caught.

I'd leave this alone unless you want the coverage: the anchor is what makes the check safe, and these are strictly less common than the `Copyright`-first and `@license`-first forms you now handle. If you do want them, they extend the alternation without weakening the anchor:

```ts
/^(?:copyright\b|\(c\)\s*\d{4}|spdx-|@license\b|licensed under\b|all rights reserved\b|(?:mit|apache|bsd|isc|gpl)[\w. ]*license\b)/i
```

Since it stays anchored, `Validates copyright headers…` keeps working.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,234 @@ function jsModuleName(filename: string): string {
...
+  return /^(?:copyright\b|spdx-|@license\b|licensed under\b)/i.test(paragraph);
+}
+
+function firstDocumentationSentence(text: string): string {
```
TheMostlyGreat, line 512: Reviewer: **The guard is right; the fallback produces a different document than the Intl path.** Probe on the same input:

| Runtime | Purpose |
|---|---|
| `Intl.Segmenter` present | `"Handles retries, e.g. on network failure, with backoff."` |
| `Intl.Segmenter` absent | `"Handles retries, e.g."` |

The fallback is the pre-round-1 regex, so it reintroduces exactly the abbreviation bug on a runtime that lacks ICU. That's a defensible choice — degraded output beats a crash in a SessionStart hook — but it has a second-order effect worth a sentence in the code: this document is committed and `architecture --check` compares it for equality, so a contributor on a `--without-intl` Node would rewrite these purposes and CI would flip them back on the next run. A quiet churn loop between environments rather than a visible failure.

Two options, both fine:
- Say so in the doc comment — "generation on a runtime without `Intl.Segmenter` yields shorter purposes; regenerate on a standard Node before committing."
- Or make the fallback not split at all (whole paragraph, then `truncatePurpose`). It would still diverge, but by *including* more rather than truncating at an abbreviation, which reads as intentional rather than broken.

Smaller point on the test: `degrades to a basic sentence splitter when the runtime has no Intl.Segmenter` uses `Keeps architecture generation available. Details are secondary.` — no abbreviation, so both paths agree and the assertion passes either way. It pins that the fallback *runs*, not that it differs. The `Object.getOwnPropertyDescriptor` save/restore is exactly right, so it's worth one extra expectation on an `e.g.` fixture to make the divergence a stated behavior instead of an accident.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,234 @@ function jsModuleName(filename: string): string {
...
+  }
+  const blockEnd = content.indexOf('*/');
+  return blockEnd === -1 ? undefined : content.slice(blockEnd + 2);
+}
+
+function isEslintDirective(content: string): boolean {
+  return content.startsWith('eslint-') || content.startsWith('eslint ');
+}
+
+function afterJsUseDirective(content: string): string | undefined {
```
TheMostlyGreat, line 448: Reviewer: No action — flagging a design choice I think is correct and would otherwise read as sloppy.

`isJsUseDirective` matches on *shape* (`use ` plus all-lowercase-ASCII words) rather than an allowlist of known directives. That means `'use cache'`, `'use server'`, and whatever React or a bundler ships next all work with no code change. The cost is that a leading `'use the force';` is also skipped — probe-confirmed — but skipping a leading string expression can only ever help find the JSDoc below it, so the false positive is free.

An allowlist would have looked more careful and been worse: it would silently stop seeding for a whole framework the day a new directive lands. The `isLowercaseAsciiWord` helper is what keeps this from over-matching arbitrary strings, and it's the reason the shape check is safe.

TheMostlyGreat (commented):
Reviewer: **/quality-review + engineering review, round 4** on `f70001d`. Posted as `COMMENT` (gh authenticates as the PR author), so this carries no merge block.

**Verdict: APPROVE.** Every finding I've raised across three rounds is now closed. I re-probed 40 cases against the real extractor on this head; nothing regressed and the two open bugs are gone. Two small observations below, both NOTE-level, neither worth another round.

## Every prior finding — final status

| Round | Finding | Status |
|---|---|---|
| 1 | Extraction heuristic misfires (blocking) | Fixed |
| 1 | Rust/Python wiring matched only `/** */` | Fixed |
| 1 | Directory nodes unseedable | Fixed |
| 1 | `--check` widening undocumented | Fixed |
| 1 | 64-hex digest + duplicate parsers | **Accepted as-is** — raised three times, not actioned; reading that as deliberate |
| 1 | Advisory removal out of scope | Fixed (#1609, spec supersession, ticket + audit skill updates) |
| 1 | Orphaned steps file + stale spec | Fixed |
| 1 | Default param before required | Fixed |
| 2 | `@license` unreachable in banner filter | **Fixed** |
| 2 | `Intl.Segmenter` merges lowercase-led sentences | **Documented** in `documentationSummary` — the right call |
| 2 | Trivia skipping misses eslint blocks / `'use client'` | **Fixed** |
| 2 | `### ` injects a phantom section | **Fixed** |
| 2 | `directoryNode` reads all 8 candidates | **Fixed** — early-return loop |
| 2 | `Intl.Segmenter` absent on `--without-intl` | **Fixed** — guarded with a regex fallback |
| 2 | Mid-word truncation | **Fixed** — word boundary with the `> LENGTH/2` guard |
| 2 | Banner false-positive on licensing modules | **Fixed** (see below — one change fixed two findings) |
| 3 | `documentationParagraphs` strips a JS gutter under a neutral name | **Fixed** — doc comment now says so |
| 3 | Feature header implied a live cucumber lane | **Fixed** |

The banner fix is the one worth calling out as a piece of design rather than a patch. Anchoring the pattern to the paragraph start — `/^(?:copyright\b|spdx-|@license\b|licensed under\b)/i` — kills the `\b@` dead branch **and** the false positive on modules that merely mention copyright, in a single change. Two findings I reported separately turned out to have one root cause, and the fix found it. Probe: `/** Validates copyright headers across source files. */` now keeps its purpose, while `@license MIT` as a leading paragraph is skipped.

## What I probed

40 cases on `f70001d` against the real extractor — banner forms, markdown blocks, JS preambles, truncation, the Intl fallback, and a document-level round trip. Throwaway suite, deleted after.

**The two round-3 bugs are gone:**

| Case | Round 3 | Now |
|---|---|---|
| `@license MIT` as sole first paragraph | `"@license MIT"` | `"Real purpose."` |
| `### Heading` first paragraph | doc headings `["a", "Not a heading"]` | `["a"]` |

**Markdown block rejection** — `###`, `>`, `-`, `+` all skipped to the next paragraph; a purpose of *only* blocks correctly falls back to the placeholder; an em-dash-led purpose (`— the core reconciler.`) is correctly **not** treated as a list.

**JS preambles** — `'use strict'`, `'use client'`, `"use server"`, `/* eslint-disable */`, `/*eslint-disable*/`, and all of them stacked behind a shebang: every one now reaches the JSDoc. Degrades safely on an unterminated block comment, a bare quote, an empty file, and a shebang-only file. I also threw 500 stacked directives at `afterLeadingJsTrivia` to check the `while (true)` loop terminates — 0 ms, no hang; every branch strictly consumes.

**Still holding:** idempotence (`created` → `unchanged`, byte-identical), and human prose winning over a source comment that changed in the same interval.

## Notes (not blocking)

1. **The anchor narrows what counts as a banner** — three common shapes now pass through. Inline below with the probe output; net still the right trade.
2. **The `Intl.Segmenter` fallback produces a different document than the Intl path** on the same input, and the test doesn't pin that divergence. Inline below.

CI: all four checks green — lint, dogfood parity, and both test jobs (Node 22.22.3 and Node 24).

**Next:** merge. The notes below are optional polish, not follow-ups I'd track.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,234 @@ function jsModuleName(filename: string): string {
...
+    const line = lineEnd === -1 ? remaining : remaining.slice(0, lineEnd);
+    const trimmed = line.trim();
+    if (trimmed.length > 0 && !trimmed.startsWith('#')) break;
+    remaining = lineEnd === -1 ? '' : remaining.slice(lineEnd + 1);
+  }
+  return remaining.trimStart();
+}
+
+/** Skip recognized JS preambles that may legally precede a file-level JSDoc block. */
+function afterLeadingJsTrivia(content: string): string {
```
TheMostlyGreat, line 404: Reviewer: The rewrite is a real improvement over the previous inline regex — one named predicate per preamble form, each returning the remainder or `undefined`, with the loop doing nothing but iterate. Easy to add the next form to.

I checked the thing this shape can get wrong: `while (true)` with a non-consuming branch. Every path strictly shortens the input — `afterEslintLineDirective` and `afterLeadingShebang` cut past a newline, `afterEslintBlockDirective` past `*/`, `afterJsUseDirective` past the closing quote (index ≥ 1) — so no branch can spin. Probe with 500 stacked `'use strict';` lines completes in 0 ms and still finds the JSDoc.

One subtlety worth noting because it's load-bearing and non-obvious: `afterEslintBlockDirective` tests `startsWith('/*')`, which also matches `/**`. It doesn't eat a JSDoc block only because `content.slice(2).trimStart()` on `/** eslint-disable` yields `* eslint-disable`, which fails `isEslintDirective`. That's correct but accidental-looking — a future edit that trims `*` characters there would silently start swallowing JSDoc headers. A one-line comment saying "`/**` reaches here and is rejected by the `*` prefix" would protect it.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,234 @@ function jsModuleName(filename: string): string {
...
+  if (typeof Intl.Segmenter === 'function') {
+    return (
+      [...new Intl.Segmenter('en', { granularity: 'sentence' }).segment(text)][0]?.segment.trim() ??
+      text
+    );
+  }
+  return /^.*?[.!?](?:\s|$)/.exec(text)?.[0].trim() ?? text;
+}
+
+function truncatePurpose(purpose: string): string {
```
TheMostlyGreat, line 522: Reviewer: Word-boundary truncation resolved. `packages/cli/architecture.generated.md:82` regenerated from `…(plain markdown + grep) witho…` to `…(plain markdown + grep)…`, which reads as an elision rather than corruption.

The `> MAXIMUM_SEEDED_PURPOSE_LENGTH / 2` guard does its job on the degenerate case — probe with a single 300-character token still hard-clips at 200 rather than collapsing to nothing. That's the branch most implementations of this forget.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,234 @@ function jsModuleName(filename: string): string {
...
+  for (const rawLine of body.split(/\r?\n/)) {
+    const line = rawLine.replace(/^\s*\*\s?/, '').trim();
+    if (line.length === 0) flush();
+    else lines.push(line);
+  }
+  flush();
+  return paragraphs;
+}
+
+function isPurposeParagraph(paragraph: string): boolean {
```
TheMostlyGreat, line 504: Reviewer: This closes the phantom-section finding, and the predicate is placed where I hoped — in the paragraph selector, so the invariant ("what this returns is one safe line of prose") stays owned by the extractor rather than by whoever renders it later.

Probe on the current head: `### Heading`, `> note`, `- item`, and `+ item` all skip to the next paragraph; a comment containing *only* block syntax falls back to the placeholder; and the document-level round trip now yields headings `["a"]` instead of `["a", "Not a heading"]`. Three consecutive heals remain stable.

Nice detail: `— the core reconciler.` (em dash) is correctly **not** rejected — `[-*+]` is ASCII-only, so prose that opens with a typographic dash still seeds. That's the kind of thing a broader `\p{Dash}` class would have quietly broken.

packages/cli/src/utils/architecture-skeleton.ts
```diff
@@ -285,6 +300,234 @@ function jsModuleName(filename: string): string {
   return filename.slice(0, filename.lastIndexOf('.'));
 }
 
+/** A source-file node, using its leading module documentation as an initial purpose when present. */
+function sourceFileNode(name: string, path: string, absolutePath: string): SkeletonNode {
+  return nodeFromPurpose(name, path, purposeFromLeadingDocumentation(absolutePath));
+}
+
+/** A directory module may declare its boundary purpose in a conventional entry file. */
+function directoryNode(name: string, path: string, entryPoints: string[]): SkeletonNode {
```
TheMostlyGreat, line 309: Reviewer: Early-return loop landed — a directory whose `index.ts` matches no longer pays seven speculative reads for the remaining extensions. Same line count as the map-then-find it replaced, and it now short-circuits.

With that, `extractSkeleton` reads at most one file per node instead of up to eight per directory, which matters because it runs two-to-three times per heal and the heal runs at SessionStart.

PR #1616, opened 2026-07-28
Run acceptance coverage locally for contributors

TheMostlyGreat:
## Summary

- add `bun run test:all` to run the existing unit suite before the Gherkin acceptance lane
- document root-level and `packages/cli` test commands separately, so every command is runnable from its stated directory
- protect the command and documentation contract with a real-file test that verifies sequencing without coupling to incidental wording

## Why

Contributors previously had no single documented local command for the unit suite plus Gherkin acceptance lane. `test:all` intentionally does not replace CI-only install-proof and release-gate lanes.

## Validation

- `bun run test tests/local-test-contract.test.ts` (2/2 passing post-rebase)
- `bun run lint`
- `safeword sync-config --check`
- diff-scoped `/audit` (configuration and dependency-cruiser clean)
- CI is running on the rebased branch and is the authoritative aggregate result.

Closes #1455


README.md
```diff
@@ -521,7 +521,8 @@ Without these binaries, the scripts print a message and skip.
 
 ```bash
 # Important: Use `bun run test` (Vitest), NOT `bun test` (Bun's runner)
-bun run test                      # All tests
+bun run test:all                  # Unit and acceptance tests
```
TheMostlyGreat: Reviewer: **Blocking — this makes the block unrunnable from any single directory.**

`test:all` exists only in the root `package.json`. But `test:integration` and `test:watch` (lines 527–528) exist only in `packages/cli/package.json`, and `bunx vitest run tests/foo.test.ts` (line 526) only resolves from `packages/cli` — that's the rule CLAUDE.md spells out.

Before this PR the block was internally consistent: all five lines ran from `packages/cli`. Adding the first root-only command breaks that. Verified:

```
$ cd packages/cli && bun run test:all
error: Script not found "test:all"

$ cd <repo root> && bun run test:integration
error: Script not found "test:integration"
```

Since the point of this PR is to make this exact block trustworthy for contributors, it's worth fixing here rather than in a follow-up.

Suggested shape:

```bash
# From the repo root
bun run test:all                  # Unit + Gherkin acceptance
bun run test:bdd                  # Acceptance lane only

# From packages/cli
bun run test                      # Vitest suite
bunx vitest run tests/foo.test.ts # Single file
bun run test:integration          # Integration tests
bun run test:watch                # Watch mode
```

One warning: **don't** fix this by adding `test:all` to `packages/cli/package.json`. It has both `test` and `test:bdd`, so it looks free — but `packages/cli/cucumber.mjs` is a narrower lane than the root config (which globs `packages/*/features/**`), so a package-level `test:all` would silently skip the root `features/` directory. That's a worse failure than the current one because it's quiet.

Separately: `test:bdd` still isn't documented anywhere, so there's no listed way to run just the acceptance lane after a unit-only change. Cheap to add while you're here.

package.json
```diff
@@ -8,6 +8,7 @@
   ],
   "scripts": {
     "test": "bun run --cwd packages/cli test",
+    "test:all": "bun run test && bun run test:bdd",
```
TheMostlyGreat, line 11: Reviewer: The composition is correct — I verified `&&` short-circuits under `bun run` (built a throwaway manifest where the first script exits 3; the second never ran, and the non-zero status propagated out of the aggregate). Ordering is also load-bearing in a good way: root `test` goes through `run-vitest-with-build-lock.mjs`, which rebuilds `dist/` first, and CI notes some cucumber step defs spawn `dist/cli.js`. So unit-before-acceptance happens to satisfy that dependency. Worth a short comment somewhere so a later reorder doesn't quietly break the BDD lane.

One accuracy note, non-blocking: this isn't CI parity. `.github/workflows/ci.yml` runs four lanes in that job — `packages/cli test`, `test:slow:install-proof`, `test:bdd`, and `test:release`. `test:all` covers the first and third. A contributor can still pass `test:all` and go red in CI on the release-gate tests, which is the exact failure mode #1455 is about. Either fold `test:release` in, or soften the PR body's "the same unit-plus-acceptance coverage CI runs" — the README's own wording ("Unit and acceptance tests") is already the honest version.

packages/cli/tests/local-test-contract.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+describe('local complete-test contract (#1455)', () => {
+  it('runs the unit suite before the Gherkin acceptance suite', () => {
+    const manifest = JSON.parse(readRepoFile('package.json')) as {
+      scripts?: Record<string, string>;
+    };
+
+    expect(manifest.scripts?.test).toMatch(/\S/);
+    expect(manifest.scripts?.['test:bdd']).toMatch(/\S/);
+    expect(manifest.scripts?.['test:all']).toBe('bun run test && bun run test:bdd');
```
TheMostlyGreat: Reviewer: Exact-string equality over-couples the test to one spelling of the script.

`planning/test-definitions/…` states the behavior as two properties: `test:all` runs `test` before `test:bdd`, **and** stops if the unit suite fails. This assertion proves neither directly — it proves a string — while simultaneously being stricter than the spec. `"bun run --cwd packages/cli test && bun run test:bdd"` is behaviorally identical and would fail here.

Asserting the properties instead reads closer to the spec and survives harmless rewrites:

```ts
const all = manifest.scripts?.['test:all'] ?? '';
expect(all).toMatch(/\btest\b.*&&.*\btest:bdd\b/); // order + short-circuit
```

Not blocking — the current version does catch the one regression that matters (someone deleting `test:all`).

packages/cli/tests/local-test-contract.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  it('documents distinct complete and unit test commands', () => {
+    const readme = readRepoFile('README.md');
+
+    expect(readme).toMatch(/^bun run test:all[ \t]+# Unit and acceptance tests$/m);
```
TheMostlyGreat: Reviewer: These two regexes lock the *comment text* of a README line, which makes any wording tweak a red CI run for no defect caught. The load-bearing claim is "the README names `test:all` as the complete command" — the trailing prose isn't part of that contract.

```ts
expect(readme).toMatch(/^bun run test:all\b/m);
expect(readme).toMatch(/^bun run test[ \t]/m);
```

Worth noting the repo has already been bitten by tests asserting shipped doc content directly — those go red in CI during otherwise-fine content migrations. Keeping the assertion to the command token avoids re-buying that.

(Pass 2 of the ticket's own `quality-review.md` flagged the alignment-space coupling and fixed it with `[ \t]+`. This is the same class of problem one level up: the comment string is as incidental as the spaces were.)

packages/cli/tests/local-test-contract.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  return readFileSync(nodePath.join(repoRoot, relativePath), 'utf8');
+}
+
+describe('local complete-test contract (#1455)', () => {
+  it('runs the unit suite before the Gherkin acceptance suite', () => {
+    const manifest = JSON.parse(readRepoFile('package.json')) as {
+      scripts?: Record<string, string>;
+    };
+
+    expect(manifest.scripts?.test).toMatch(/\S/);
```
TheMostlyGreat: Reviewer: Nit — if `scripts.test` is ever missing, `toMatch` on `undefined` throws a type error rather than failing with a readable message, which buries the actual regression. `expect(manifest.scripts).toHaveProperty('test')` (or `toBeTruthy()`) reports the real cause. Same on the next line.

planning/test-definitions/TQQGZS-local-acceptance-verification.md
```diff
@@ -0,0 +1,25 @@
+# Test definitions: local acceptance verification
```
TheMostlyGreat: Reviewer: **Blocking — wrong location, and a new top-level directory.**

`planning/` doesn't exist on `main` and isn't referenced by `schema.ts`, `knip.json`, `.prettierignore`, or any guide. The canonical path is `<namespace-root>/tickets/{ID}-{slug}/test-definitions.md`:

- `.safeword/guides/planning-guide.md:327` states it explicitly
- `packages/cli/src/health.ts:401` reads `join(ticketDirectory, 'test-definitions.md')`
- `packages/cli/src/boundary/engine.ts:27` lists it as a ticket artifact
- 142 existing tickets follow it

So this file is invisible to every check it exists to satisfy — it reads as coverage without being coverage.

Move it to `.project/tickets/TQQGZS-local-acceptance-verification/test-definitions.md`, or drop it: the ticket is `type: task`, and the feature done-gate that requires `test-definitions.md` doesn't fire for tasks. Given the two scenarios here restate what the 29-line test already asserts, dropping is the cleaner call.

The stakes are higher than usual in this repo specifically — it's the one that ships the workflow, so its own layout is the worked example.

planning/user-stories/TQQGZS-local-acceptance-verification.md
```diff
@@ -0,0 +1,17 @@
+# Run acceptance coverage locally for contributors
```
TheMostlyGreat: Reviewer: **Blocking — no precedent for this artifact.**

There is no `user-stories` directory anywhere in the repo, and no ticket has ever carried one. The JTBD/acceptance-criteria artifact in this system is `spec.md`, inside the ticket directory. `ticket.md` already carries Goal / Why / Scope / Out of Scope / Done When, and its content overlaps this file almost completely.

Suggest deleting it. If AC3 ("`bun run lint` behavior and cost remain unchanged") is the part worth keeping, it already lives in the ticket's **Out of Scope**.

Together with the `planning/test-definitions/` file, this is most of the PR's 208 added lines standing behind a 1-line script change.

.project/tickets/TQQGZS-local-acceptance-verification/verify.md
```diff
@@ -0,0 +1,25 @@
+## Verify Checklist
+
+**Test Suite:** ✓ 2/2 focused contract tests pass. The full unit suite was also run through Safeword's test plan and `test:all`, but the runner did not retain its final summary.
```
TheMostlyGreat: Reviewer: Credit where due — flagging the missing full-suite summary rather than writing ✅ is the right call, and the **Evidence limits** section at the bottom is exactly how that should be handled.

Two things now fill part of the gap:

1. I ran the root acceptance lane on this head: **508 scenarios, 505 passed, 3 skipped, 4m 6s**. The "Gherkin: ✅" line is confirmed.
2. The full Vitest suite is running; I'll post the result as a follow-up comment.

The part that isn't filled: CI has never run on this PR (`gh pr checks 1616` → *no checks reported on the branch*). So "CI remains the authoritative aggregate result" in the PR body is currently a forward-looking statement, not evidence. A rebase should both refresh the stale `CONFLICTING` state and get the checks to fire.

TheMostlyGreat (commented):
Reviewer: **/quality-review + engineering review — REQUEST CHANGES** (posted as `COMMENT`; `gh` authenticates as the PR author, so this carries no merge block).

Reviewed at `7d30028`, against `main` `b8cc0088`.

## Quality Review

**Currency:** ✓ No dependency or runtime version changes. `&&` short-circuit under `bun run` verified empirically against the installed Bun, not from docs.
**Sources:** ⚠️ The PR body's central claim — "the same unit-plus-acceptance coverage CI runs" — is not accurate against `.github/workflows/ci.yml`. See finding 4.
**Correct:** ⚠️ The script composes correctly, but the README edit it ships is not runnable as documented. See finding 1.
**Elegant:** ⚠️ The contract test asserts literal strings (script body, README comment text) rather than the properties the test-definitions actually state.
**No-bloat:** ❌ 208 additions for a 1-line script + 2-line README edit, including a **new top-level `planning/` directory** that does not exist anywhere in this repo. See finding 2.
**Wiring:** ✓ Accepting the recursion justification — `test:all` cannot spawn itself from inside the suite it runs. The existence assertions on `test` / `test:bdd` are a reasonable substitute.

**Verdict:** REQUEST CHANGES

---

## What I verified green myself

Ran on the PR head, not taken on trust:

| Check | Result |
| --- | --- |
| `bun run test:bdd` (root acceptance lane) | ✅ 508 scenarios, 505 passed, 3 skipped — 4m 6s |
| `bun packages/cli/src/cli.ts architecture --check` | ✅ current |
| `bun scripts/parity-check.ts --mode=all` | ✅ 195 pairs, 8 contracts in sync |
| `bun scripts/check-pr-ticket-done.ts` | ✅ pass |
| `prettier --check`, `markdownlint-cli2`, `eslint` on the diff | ✅ clean |
| `tests/local-test-contract.test.ts` | ✅ 2/2 |
| Three-way merge vs `main` `b8cc0088` | ✅ clean, 0 conflicts |

The full Vitest suite is queued behind another checkout's run under the machine-global build lock; I'll append the result as a follow-up comment.

**The premise checks out.** I went looking for whether the unit suite already executes acceptance scenarios (which would have made `test:all` redundant) — it does not. `packages/cli/tests/integration/cucumber-bdd.test.ts` is a `--dry-run` wiring check only, and says so in its own docstring. So the gap #1455 describes is real and this is the right shape of fix.

---

## Blocking

**1. The README edit breaks the block it's fixing** (`README.md:524`)

Before this PR, all five lines in that block were runnable from `packages/cli`. `test:all` exists **only** at the repo root, so the block is now runnable from no single directory. Verified:

```
$ cd packages/cli && bun run test:all   → error: Script not found "test:all"
$ cd <root>       && bun run test:integration → error: Script not found "test:integration"
$ cd <root>       && bun run test:watch       → error: Script not found "test:watch"
```

A contributor following the fixed doc top-to-bottom now hits an error on line 1 or line 4, depending on where they stand.

**2. `planning/` is an invented top-level directory** (`planning/test-definitions/…`, `planning/user-stories/…`)

`planning/` does not exist on `main` and appears in no schema, guide, or tool config. The canonical path is `<namespace-root>/tickets/{ID}-{slug}/test-definitions.md` — `planning-guide.md:327` states it, and 142 existing tickets follow it. Consequences: `health.ts:401` and `boundary/engine.ts:27` read the ticket directory, so these files are invisible to the health, boundary, and coverage checks they exist to satisfy. `user-stories` has no precedent at all in this repo — `spec.md` is the JTBD/criteria artifact.

This is the repo that ships the workflow. A stray convention here becomes the example.

**3. The PR has never been checked and GitHub reports it unmergeable**

`gh pr checks 1616` → *no checks reported on the branch*. `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. The body says "CI remains the authoritative aggregate result" — CI has not run, so nothing is backing the aggregate claim right now. A local three-way merge against `main` `b8cc0088` is clean, so the conflict is most likely a stale mergeability computation; a rebase/push should refresh it and trigger CI.

---

## Non-blocking

**4. `test:all` is not CI parity.** CI also runs `test:slow:install-proof` and `test:release` in the same job. `test:all` omits both, so a contributor can still be green locally and red in CI. The README label ("Unit and acceptance tests") is honest — it's the PR body and the ticket's *Why* that overreach. Either soften the claim or fold `test:release` in.

**5. Don't add `test:all` to `packages/cli` as the fix for finding 1.** It looks like the obvious move — `packages/cli` already has both `test` and `test:bdd` — but `packages/cli/cucumber.mjs` is a *narrower* lane than the root config, which globs `packages/*/features/**`. You'd get a `test:all` that silently skips the root `features/` directory. Annotate the cwd in the README block instead, or split it into "from the repo root" and "from `packages/cli`".

**6. Contract-test coupling.** Three inline comments below on `local-test-contract.test.ts` — exact script-string equality, README comment-text matching, and a `toMatch(/\S/)` nit.

**7. `verify.md` records the aggregate as unverified.** Honest, and I appreciate the explicit evidence-limit note. The acceptance-lane half is now covered by my run above.

---

**Provenance:**
- (verified: `.github/workflows/ci.yml:94–120`, read this session) — CI runs `packages/cli test`, `test:slow:install-proof`, `test:bdd`, `test:release`.
- (verified: `bun run` executed against the installed Bun this session) — `&&` short-circuits; the missing-script errors above are real output.
- (verified: `.safeword/guides/planning-guide.md:327`, `packages/cli/src/health.ts:401`, `packages/cli/src/boundary/engine.ts:27`, and `find .project/tickets -name test-definitions.md` → 142 hits) — canonical artifact location.
- (verified: `gh pr checks 1616`, `git merge-tree --write-tree origin/main pr-1616`) — check and merge state.

**Next:** move the two `planning/` files into `.project/tickets/TQQGZS-local-acceptance-verification/` (or drop them — this is a `type: task`, so `test-definitions.md` isn't gated), give the README block a cwd, rebase to clear the stale conflict and get CI to actually run. Findings 4–7 are optional.

README.md
TheMostlyGreat: Resolved in 6d54057. The README now splits the block into “From the repo root” (`test:all` and `test:bdd`) and “From packages/cli” commands, so every documented command runs from its stated manifest.

package.json
TheMostlyGreat: Resolved in 6d54057. `test:all` remains deliberately scoped to the unit and acceptance lanes; the PR body now says it does not replace CI-only install-proof or release-gate checks. The README names the required order, and the contract test verifies the two `&&`-separated commands without adding a narrower package-level aggregate.

packages/cli/tests/local-test-contract.test.ts
TheMostlyGreat: Resolved in 6d54057. The contract now validates nonempty collaborators, splits `test:all` at `&&`, and asserts `test` then `test:bdd`. That proves sequence and short-circuit composition while allowing harmless command-spelling changes.

packages/cli/tests/local-test-contract.test.ts
TheMostlyGreat: Resolved in 6d54057. The README contract no longer asserts either trailing comment. It checks only the root command grouping and package-level `bun run test` token, which are the load-bearing documentation facts.

packages/cli/tests/local-test-contract.test.ts
TheMostlyGreat: Resolved in 6d54057. The test now uses `objectContaining` with nonempty string matchers for all three scripts, producing a structured assertion failure instead of invoking `toMatch` on `undefined`.

planning/test-definitions/TQQGZS-local-acceptance-verification.md
TheMostlyGreat: Resolved in 6d54057. I deleted the redundant task test-definitions file rather than creating a noncanonical top-level `planning/` tree or preserving duplicate coverage metadata.

planning/user-stories/TQQGZS-local-acceptance-verification.md
TheMostlyGreat: Resolved in 6d54057. I deleted the nonstandard user-stories artifact; the existing task ticket remains the canonical scope and acceptance record.

.project/tickets/TQQGZS-local-acceptance-verification/verify.md
TheMostlyGreat: Resolved operationally in 6d54057: rebasing and pushing refreshed GitHub to `MERGEABLE` and started the CI matrix. I kept the local verification record appropriately cautious and updated the PR body to describe CI as the authoritative aggregate result.

packages/cli/tests/local-test-contract.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    expect(commands).toHaveLength(2);
+    expect(commands[0] ?? '').toMatch(/(?:^|\s)test(?:\s|$)/);
+    expect(commands[1] ?? '').toMatch(/\btest:bdd\b/);
+  });
+
+  it('documents the root and package test commands from their matching directories', () => {
+    const readme = readRepoFile('README.md');
+
+    expect(readme).toMatch(/^# From the repo root\nbun run test:all\b[^\n]*\nbun run test:bdd\b/m);
+    expect(readme).toMatch(/^bun run test[ \t]/m);
```
TheMostlyGreat: Reviewer: **Non-blocking, but this is the half of the fix that isn't guarded.**

Line 37 structurally pins the root block — heading, then `test:all`, then `test:bdd`, consecutive. Good. Line 38 only asserts a `bun run test ` line exists *somewhere* in the file. Nothing ties the four package-scoped commands to the `# From packages/cli` heading.

I mutation-tested it against the shipped README, with a control:

| Mutation | Result |
| --- | --- |
| Delete the `# From packages/cli` heading | **passes** — all 4 package commands now read as root commands |
| Move `bun run test:integration` into the root block | **passes** |
| Merge both blocks under the root heading | **passes** |
| *Control:* rename the root heading | fails, as expected |

All three mutations are the round-1 regression coming back on the other side of the block.

This pair closes it — I checked each line against all three mutations *and* against the shipped README, so it isn't a suggestion I haven't run:

```ts
// root block: also require it to END, so nothing can be appended into it
expect(readme).toMatch(
  /^# From the repo root\nbun run test:all\b[^\n]*\nbun run test:bdd\b[^\n]*\n\n/m,
);
// package block: anchor the command to its heading, tolerating the `# Important:` line
expect(readme).toMatch(/^# From packages\/cli\n(?:#[^\n]*\n)*bun run test[ \t]/m);
```

Result: shipped README passes, all three mutations fail.

The trailing `\n\n` on the first one is what catches the `test:integration` move — without it, the root block has no defined end and anything can be appended. That's the non-obvious part; my first attempt missed that mutation.

Worth weighing against the cost: this does couple the test to two heading strings. I think it's the right trade here because those headings *are* the fix — they're what makes each command unambiguous — but reasonable to disagree and leave it as-is.

packages/cli/tests/local-test-contract.test.ts
```diff
@@ -0,0 +1,40 @@
...
+        test: expect.stringMatching(/\S/),
+        'test:bdd': expect.stringMatching(/\S/),
+        'test:all': expect.stringMatching(/\S/),
+      }),
+    );
+    const commands = (manifest.scripts?.['test:all'] ?? '')
+      .split('&&')
+      .map(command => command.trim());
+
+    expect(commands).toHaveLength(2);
```
TheMostlyGreat, line 29: Reviewer: This is better than the fix I suggested, and I want to be specific about why rather than just say "nice."

I proposed a single loose regex (`/\btest\b.*&&.*\btest:bdd\b/`). Splitting on `&&` and asserting the arity is strictly stronger — it makes the separator itself load-bearing, so a `;` or `||` collapses the split to one element and fails. I replicated these assertions and ran mutations through them:

| `test:all` mutation | Result |
| --- | --- |
| `bun run test; bun run test:bdd` | caught |
| `bun run test \|\| bun run test:bdd` | caught |
| `bun run test:bdd && bun run test` | caught (line 30's `(?:^\|\s)test(?:\s\|$)` won't match `test:bdd` — the `:` blocks it) |
| `bun run test:bdd` alone | caught |
| `bun run --cwd packages/cli test && bun run test:bdd` | correctly tolerated |

That covers both clauses in the test-definitions — ordering *and* "stops if the unit suite fails" — with no false positive on an equivalent rewrite.

One note for future readers, not a change request: the `?? ''` on lines 30–31 looks like dead defensive code after `toHaveLength(2)`, but `noUncheckedIndexedAccess: true` in `packages/cli/tsconfig.json` requires it. Leave it.

README.md
```diff
@@ -520,8 +520,13 @@ Without these binaries, the scripts print a message and skip.
 **Running Tests:**
 
 ```bash
+# From the repo root
```
TheMostlyGreat, line 523: Reviewer: Round-1 blocker resolved — verified rather than assumed. I cross-checked every documented command against both manifests:

```
root   test:all         OK        cli  test              OK
root   test:bdd         OK        cli  test:integration  OK
                                  cli  test:watch        OK
```

And confirmed the cross-scope leakage that caused the original problem is gone: `test:all` is root-only, `test:integration` and `test:watch` are package-only, and each now sits under the heading that matches.

One leftover worth a line, low priority: `packages/cli` *also* defines `test:bdd`, and it runs a narrower lane — `packages/cli/cucumber.mjs` rather than the root config that globs `packages/*/features/**`. Putting `test:bdd` under the root heading is exactly right, but a contributor working in the `# From packages/cli` section could still run it there and silently get a subset of the acceptance lane with no error. A short parenthetical on the `test:bdd` line — something like `# Acceptance lane only (root only — the packages/cli one is narrower)` — would close the last version of this trap.

TheMostlyGreat (commented):
Reviewer: **Round 2 — /quality-review + engineering review at `7cb8d13`. APPROVE with two non-blocking notes.**

All three blocking findings from my first review are resolved, and I verified each rather than taking the commit messages for it.

| Round-1 blocker | Status |
| --- | --- |
| README block unrunnable from any single directory | ✅ Fixed — split into `# From the repo root` / `# From packages/cli`. I checked all five documented commands resolve in their stated manifest. |
| `planning/` invented top-level directory | ✅ Deleted, both files. Diff is down to 8 files, +181/-1. |
| Never checked, reported `CONFLICTING` | ✅ Rebased onto `b8cc0088`, now `MERGEABLE`. CI green: Dogfood parity, lint, test (node 22.22.3), test (node 24). |

Also addressed: the exact-string coupling on the script, the README comment-prose lock, the `toMatch` on a possibly-undefined value, and the PR body's CI-parity over-claim (now explicit that `test:all` doesn't replace install-proof and release-gate).

## Quality Review

**Currency:** ✓ No dependency or runtime changes.
**Sources:** ✓ The body's claims now match `.github/workflows/ci.yml`.
**Correct:** ✓ Every documented command resolves from its stated directory — verified against both manifests.
**Elegant:** ✓ The sequencing assertion is now property-based, and better than what I suggested. See below.
**No-bloat:** ✓ 27 lines lighter, and the two orphan artifacts are gone.
**Wiring:** ✓ Unchanged and still justified — `test:all` can't spawn itself from inside the suite it runs.

**Verdict:** APPROVE — the two notes below are optional and neither blocks merge.

---

## Credit where it's due: the sequencing assertion is stronger than my suggestion

I proposed a single loose regex. The `split('&&')` + `toHaveLength(2)` approach that landed is better, and I mutation-tested it to confirm rather than eyeballing it — I replicated the assertions and ran real regressions through them:

| Mutation of `test:all` | Result |
| --- | --- |
| `bun run test; bun run test:bdd` (loses short-circuit) | ✅ caught |
| `bun run test \|\| bun run test:bdd` (inverts it) | ✅ caught |
| `bun run test:bdd && bun run test` (order reversed) | ✅ caught |
| `bun run test:bdd` (unit lane dropped) | ✅ caught |
| `bun run --cwd packages/cli test && bun run test:bdd` (equivalent rewrite) | ✅ correctly tolerated |

That's the whole behaviour the test-definitions describe, including the short-circuit clause, with no false positive on a harmless rewrite.

---

## Note 1 — the doc-contract test guards the root block but not the `packages/cli` block

The regression I found in round 1 was *a command listed under the wrong working directory*. The fix is symmetric — two labelled blocks — but the test is not: it structurally verifies the root block and only loosely checks that a `bun run test ` line exists somewhere.

Mutation-tested against the shipped README, with a control to prove the harness works:

| Mutation | Test result |
| --- | --- |
| Delete the `# From packages/cli` heading (all 4 package commands now read as root commands) | ❌ **passes** |
| Move `bun run test:integration` up into the root block | ❌ **passes** |
| Merge both blocks under the root heading | ❌ **passes** |
| *Control:* rename the root heading | ✓ fails, as it should |

So the precise regression this PR exists to fix could be reintroduced on the `packages/cli` side without the test noticing. Details and a validated replacement in the inline comment.

## Note 2 — the ticket's evidence artifacts weren't refreshed after this round

`ticket.md`, `verify.md`, `quality-review.md`, and `refactor-ledger.md` are byte-identical to the version I reviewed at `7d30028` (`git diff 7d3002804..7cb8d133 -- <each>` is empty), but the implementation changed substantially since: two files deleted, the contract test rewritten, the README block restructured, and a rebase onto main.

Concretely stale:

- `quality-review.md` ends at **Pass 2 → APPROVE**, "Suggested improvements: None — both Pass 1 items are complete." Two substantive commits landed *after* that approval, with no Pass 3.
- `verify.md` **PR Scope** still says the diff contains "supporting planning/review evidence" — that's the `planning/` directory, now deleted.
- `verify.md` **Evidence limits** still flags the lost full-suite summary. Moot now: CI is green on both Node majors.
- `refactor-ledger.md` lists only the two original Pass-1 items; nothing for this round's refactors.

Nothing here affects the shipped behaviour. Flagging it because this is the repo that ships the workflow, and the done-gate treats these as the evidence of record — a `verify.md` describing a diff that no longer exists is the failure mode the gate is meant to prevent.

---

**What I ran on `7cb8d13`:** the contract test (**2/2 pass**); both manifests cross-checked against all five documented commands; `eslint` and `tsc --noEmit` (clean); `prettier --check` (clean); mutation tests on both assertions with controls; `git diff` on each ticket artifact. I didn't re-run the full local suite — CI already ran it green on node 22 and node 24, which is the stronger signal here since my machine is contended and CI's isn't.

**Provenance:**
- (verified: `gh pr checks 1616` this session) — 4/4 green; `mergeable: MERGEABLE`.
- (verified: `node -e` cross-check of root and `packages/cli` `package.json` this session) — all documented commands resolve where stated.
- (verified: mutation runs against the shipped `README.md` and the test's own regexes this session) — the pass/fail tables above are real output, not reasoning.
- (verified: `git diff 7d3002804..pr-1616 -- <artifact>` empty for all four) — artifacts unchanged.

**Next:** merge when you're ready — neither note blocks. If you take Note 1, the inline comment has a replacement regex pair I validated against all three mutations plus the shipped file.

.project/tickets/TQQGZS-local-acceptance-verification/verify.md
```diff
@@ -0,0 +1,25 @@
...
+
+**Gherkin:** ✅ Acceptance lane passes. It ran directly and through Safeword's BDD test plan.
+
+**Build:** ✅ Success during each focused contract run.
+
+**Lint:** ✅ Clean (`eslint`, Gherkin lint, and package typecheck).
+
+**Scenarios:** All 0 scenarios marked complete — this task uses a document-contract test rather than a scenario ledger.
+
+**PR Scope:** ✅ Diff matches ticket scope. It contains the root test command, contributor guidance, its contract test, and supporting planning/review evidence only.
```
TheMostlyGreat: Reviewer: This file is byte-identical to the version at `7d30028` — `git diff 7d3002804..7cb8d133 -- <this file>` is empty — but the diff it describes has changed underneath it.

**PR Scope** (this line) says the diff contains "supporting planning/review evidence". That's `planning/test-definitions/` and `planning/user-stories/`, both deleted this round. The scope claim now describes files that aren't in the PR.

**Evidence limits** (line 23) still flags the lost full-suite summary. That's moot — CI is green on node 22.22.3 and node 24, and I posted a local full-suite result on the previous head.

**Test Suite** (line 3) still reads "2/2 focused contract tests pass" against the *old* two tests; both were rewritten this round.

Not a behaviour problem, and it doesn't block. But this repo ships the workflow that treats `verify.md` as the evidence of record, and a verify record describing a superseded diff is the exact thing that gate exists to catch. A quick re-run would let you replace the whole **Evidence limits** paragraph with the CI links.

.project/tickets/TQQGZS-local-acceptance-verification/quality-review.md
```diff
@@ -0,0 +1,58 @@
...
+
+**Correct:** ✓ The contract now proves that both named collaborators exist and that unit tests precede acceptance tests.
+
+**Elegant:** ✓ The README assertions retain their content contract without depending on alignment spaces.
+
+**No-bloat:** ✓ The change remains one root script and one small real-file contract test.
+
+**Wiring:** ✓ `packages/cli/tests/local-test-contract.test.ts` reads the real root manifest and README. It does not recursively spawn `test:all`; direct aggregate execution is recorded separately in `verify.md`.
+
+**Verdict:** APPROVE
```
TheMostlyGreat, line 47: Reviewer: Pass 2 closes at **APPROVE** with "Suggested improvements: None — both Pass 1 items are complete."

Two substantive commits landed after that approval — `6d54057` (review findings) and `7cb8d13` (relaxed doc contract) — which deleted two files, rewrote the contract test, and restructured the README block. No Pass 3 records them, so the file reads as though the work ended at Pass 2.

The skill's own loop is review → fix → **re-review**, and the re-review is what's missing. A short Pass 3 noting what changed and why the verdict still holds would make the record match the branch. Low priority; the code itself is in good shape, which is why my round-2 verdict is APPROVE.

.project/tickets/TQQGZS-local-acceptance-verification/refactor-ledger.md
```diff
@@ -0,0 +1,4 @@
+# Refactor ledger
+
+- [x] Assert that the composed root scripts exist before asserting the aggregate command.
+- [x] Match README command lines without coupling the test to alignment spaces.
```
TheMostlyGreat, line 4: Reviewer: Both entries here are from the original Pass-1 round, and the second one — "Match README command lines without coupling the test to alignment spaces" — has been overtaken: the README assertions were rewritten again this round to drop the comment-prose coupling entirely, which is a different and larger change than the alignment-space fix it describes.

This round's two refactors (property-based sequencing assertion; split root / `packages/cli` documentation blocks) have no entries. Two lines would bring it current. Cosmetic — flagging only because the ledger is one of the artifacts the done-gate reads.

packages/cli/tests/local-test-contract.test.ts
TheMostlyGreat: Resolved in edc35e5. The contract now requires the root block to end after `test:bdd`, then anchors the package `bun run test` command to `# From packages/cli` while allowing its explanatory comment. This rejects the three documented scope-leak mutations.

packages/cli/tests/local-test-contract.test.ts
TheMostlyGreat: Resolved without a code change. Confirmed: the `split('&&')` arity assertion is intentionally stronger than the prior regex, and the `?? ''` fallbacks are required by `noUncheckedIndexedAccess` for readable matcher inputs.

README.md
TheMostlyGreat: Resolved in edc35e5. The root `test:bdd` line now says that it is root-only because the `packages/cli` lane is narrower; package behavior is unchanged.

.project/tickets/TQQGZS-local-acceptance-verification/verify.md
TheMostlyGreat: Resolved in edc35e5. `verify.md` now records the final focused contract result, green CI on both Node majors, accurate scope without the deleted planning files, and the CI run as the authoritative aggregate evidence.

.project/tickets/TQQGZS-local-acceptance-verification/quality-review.md
TheMostlyGreat: Resolved in edc35e5. Added Pass 3 with the current-source review, both working-directory contracts, CI provenance, and an APPROVE verdict after this re-review.

.project/tickets/TQQGZS-local-acceptance-verification/refactor-ledger.md
TheMostlyGreat: Resolved in edc35e5. The ledger now records the scoped README-block contract and the root-only acceptance-lane clarification.

.project/tickets/TQQGZS-local-acceptance-verification/verify.md
```diff
@@ -0,0 +1,25 @@
...
+
+**Dep Drift:** ✅ Clean — no dependency changes.
+
+**Parent Epic:** N/A.
+
+**Reconcile:** ✅ No pattern deviation.
+
+**Experience:** ⏭️ N/A — internal contributor tooling.
+
+**Evidence limits:** ✅ None — the full CI matrix passed: Dogfood parity, lint, and test jobs on Node 22.22.3 and Node 24 ([run 30429519111](https://github.com/ArcadeAI/safeword/actions/runs/30429519111)).
```
TheMostlyGreat: Reviewer: Good refresh — **Evidence limits** going from a ⚠️ caveat to a ✅ with a CI link is exactly what round 2 was asking for.

One correction: run **30429519111** is the *previous* head's run.

```
run 30429519111  head=7cb8d1333  success   ← cited here
run 30431181570  head=46c1a6a8b  success   ← this PR's current head
```

This line was written in `46c1a6a8`, the same commit that rewrote the README block and the contract test — so the run it cites finished before the code it vouches for existed. Both are green, so nothing is actually unverified; the pointer is just one commit stale.

```
[run 30431181570](https://github.com/ArcadeAI/safeword/actions/runs/30431181570)
```

Same swap needed at `quality-review.md:83`.

Worth noting the general shape, since it'll recur: a verification artifact written *in* the commit it verifies can't cite that commit's CI run, because the run doesn't exist yet. Either cite the run after it completes, or reference the PR's checks rather than a specific run id.

.project/tickets/TQQGZS-local-acceptance-verification/quality-review.md
```diff
@@ -0,0 +1,86 @@
...
+
+**Critical issues:** None.
+
+**Suggested improvements:** None.
+
+**Provenance:**
+
+- (verified: https://bun.sh/docs/runtime) — fetched 2026-07-29; `bun run <script>` executes named package scripts.
+- (verified: `.github/workflows/ci.yml` and CI run 30429519111) — current CI covers build, unit, acceptance, install-proof, release-gate, lint, and dogfood parity lanes.
```
TheMostlyGreat: Reviewer: Pass 3 addresses the round-2 note properly — the loop now reads review → fix → re-review, and the per-angle findings match what actually changed.

Same stale citation as `verify.md:23`: run **30429519111** has head `7cb8d133`, one commit behind this file. The run covering `46c1a6a8` is **30431181570**, also green.

One more thing worth checking on this line: it claims CI covers "build, unit, acceptance, install-proof, release-gate, lint, and dogfood parity lanes." That's accurate for `ci.yml` as written — I confirmed all seven steps exist. But `test:all` itself only covers unit + acceptance, which the PR body now states correctly. The Pass 3 wording is about CI, not `test:all`, so it's right as written; flagging only because those two lists are easy to conflate on a later read.

packages/cli/tests/local-test-contract.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    expect(commands[0] ?? '').toMatch(/(?:^|\s)test(?:\s|$)/);
+    expect(commands[1] ?? '').toMatch(/\btest:bdd\b/);
+  });
+
+  it('documents the root and package test commands from their matching directories', () => {
+    const readme = readRepoFile('README.md');
+
+    expect(readme).toMatch(
+      /^# From the repo root\nbun run test:all\b[^\n]*\nbun run test:bdd\b[^\n]*\n\n/m,
```
TheMostlyGreat, line 38: Reviewer: This closes the gap — I re-ran the full mutation battery against it, and all five regressions are caught while the shipped README passes.

But I owe you a correction on my own suggestion: the `\n\n` anchor I proposed pins the root block to *exactly two commands in this order*. Legitimate edits now fail:

| Edit | Result |
| --- | --- |
| Add a third root command (`bun run test:done # …`) | fails |
| Swap the order of `test:all` / `test:bdd` | fails |

Neither is a regression. That's a false positive aimed at someone who did nothing wrong, and the natural reaction is to loosen the regex — losing the coverage this round bought.

Extracting the block and asserting on its *contents* keeps the coverage without the brittleness:

```ts
const rootBlock =
  readme.match(/^# From the repo root\n([\s\S]*?)\n# From packages\/cli\n/m)?.[1] ?? '';

expect(rootBlock).toMatch(/^bun run test:all\b/m);
expect(rootBlock).toMatch(/^bun run test:bdd\b/m);
expect(rootBlock).not.toMatch(/^bun run (?:test:integration|test:watch)\b/m);
expect(readme).toMatch(/^# From packages\/cli\n(?:#[^\n]*\n)*bun run test[ \t]/m);
```

I validated it 8/8 before suggesting it this time — shipped README passes; all five mutations caught; both legitimate edits pass. The `not.toMatch` line is what carries the real contract: package-scoped commands must not appear in the root block.

Totally optional. The current version is correct for the README as it stands today, and this only matters if the block ever grows.

README.md
```diff
@@ -520,8 +520,13 @@ Without these binaries, the scripts print a message and skip.
...
+bun run test:all # Unit suite, then acceptance tests
+bun run test:bdd # Acceptance lane only (root only; packages/cli is narrower)
```
TheMostlyGreat, line 525: Reviewer: Resolved — and the wording is better than what I suggested. "(root only; `packages/cli` is narrower)" tells a contributor both the rule and the reason to care, in six words, without explaining cucumber config discovery at them.

Re-confirmed the underlying fact on this head: `test:bdd` exists in both manifests, and `packages/cli/cucumber.mjs` doesn't carry the root config's `packages/*/features/**` glob — so running it from the package directory really does silently cover less. The warning is accurate.

TheMostlyGreat (commented):
Reviewer: **Round 3 — /quality-review + engineering review at `46c1a6a8`. APPROVE — ship it.**

Both round-2 notes are fully addressed, and I re-ran my mutation battery against the new code rather than assuming the fix landed correctly.

| Round-2 note | Status |
| --- | --- |
| Doc test guarded only the root block | ✅ Fixed. **5/5** mutations now caught, shipped README still passes — table below. |
| Ticket evidence artifacts stale | ✅ Refreshed. Pass 3 added, ledger gained 2 entries, `verify.md` rewritten, work-log entry added. |
| `packages/cli` `test:bdd` is a narrower lane (minor) | ✅ Warned inline in the README. |

CI green on the current head: run [30431181570](https://github.com/ArcadeAI/safeword/actions/runs/30431181570) — Dogfood parity, lint, test (node 22.22.3), test (node 24).

## The hardening works — verified, not assumed

I re-ran every mutation from round 2 against the shipped regexes, plus a new one and a control:

| Mutation | Result |
| --- | --- |
| Delete the `# From packages/cli` heading | ✅ caught |
| Move `test:integration` into the root block | ✅ caught |
| Merge both blocks under the root heading | ✅ caught |
| Remove `test:all` from the root block (new this round) | ✅ caught |
| *Control:* rename the root heading | ✅ caught |
| Shipped README | ✅ passes |

The round-1 regression can no longer come back on either side of the block. Local checks on this head: contract test **2/2**, `eslint` clean, `tsc --noEmit` clean, `prettier --check` clean, `markdownlint` clean. I also re-verified command scoping against both manifests — `test:all` root-only, `test:integration` and `test:watch` package-only, each under the matching heading.

## Quality Review

**Currency:** ✓ No dependency or runtime changes.
**Sources:** ✓ Claims match the manifests and `ci.yml`. One citation points at the wrong CI run — note 1.
**Correct:** ✓ Verified against both manifests and by mutation.
**Elegant:** ✓ Block-boundary assertions, no prose coupling.
**No-bloat:** ✓ One README clause, one regex pair, evidence refresh. Nothing new invented.
**Wiring:** ✓ Unchanged, still justified.

**Verdict:** APPROVE. Two notes below, both optional; neither is worth another round.

---

## Note 1 — the refreshed evidence cites the previous head's CI run

`verify.md:23` and `quality-review.md:83` both cite run **30429519111**. That run's head is **`7cb8d133`** — the commit *before* this one. Pass 3 and the `verify.md` rewrite were both authored in `46c1a6a8`, which also changed the README and the contract test, so the cited run predates the code it vouches for by one commit.

```
run 30429519111  head=7cb8d1333  success   ← cited
run 30431181570  head=46c1a6a8b  success   ← actually covers this head
```

Benign in outcome — both runs are green — but it's the same shape as the round-2 note: the evidence record pointing at a superseded artifact. Swapping the run id in two places closes it.

## Note 2 — my round-2 suggestion has a false-positive cost, and I should own that

I proposed the `\n\n` anchor on the root block, and it was taken as-is. It does close the gap, but it also pins the root block to *exactly these two commands in this order*. Legitimate future edits now fail:

| Future edit | Result |
| --- | --- |
| Add a third root-scoped command (e.g. `test:done`) | ❌ fails |
| Reorder `test:all` / `test:bdd` within the block | ❌ fails |

Neither is a regression, so that's the test crying wolf at a contributor who did nothing wrong — and the likely response is to loosen the regex under time pressure, losing the coverage.

I worked out a version that keeps all five mutations caught *and* tolerates both edits, and validated it 8/8 before suggesting it this time. It's in the inline comment on the assertion.

Entirely optional — the current version is correct today, and this is polish on polish. Fine to merge as-is and pick it up if the block ever grows.

---

**Provenance:**
- (verified: `gh api .../actions/runs/{30429519111,30431181570}` this session) — the head SHA of each run, quoted above.
- (verified: mutation runs against the shipped `README.md` and the regexes as they now ship, this session) — both tables are real output.
- (verified: contract test, `eslint`, `tsc --noEmit`, `prettier`, `markdownlint` on `46c1a6a8`) — all clean.
- (verified: `node -e` cross-check of both manifests) — command scoping holds.

**Next:** Merge. If you want the two-line evidence fix from note 1 first, that's the only thing I'd bother changing before it lands.

PR #1647, opened 2026-07-29
Protect retro filing from malformed GitHub CLI credentials

TheMostlyGreat:
## Summary

- reject gh auth token output that is unsafe for a Bearer header
- keep fallback child-process context independent from injected token lookup input
- reset subprocess mocks between tests and simplify the single-case test table

This follows up the four late review threads on merged PR #1577 without prompting users for credentials. GH_TOKEN remains an opaque, explicit GitHub CLI credential; only RFC-invalid output is rejected locally.

Closes #1637

## Validation

- focused resolver/transport suite: 58/58
- bun run test exercised; two unrelated integration tests timed out under shared load, and both passed in isolation
- bun run test:bdd: 93 scenarios and 1,109 steps passed
- bun run lint, TypeScript typecheck, Prettier, dependency-cruiser, Knip, and audit
- fresh independent quality reviews: approved

### DIFF
--- a/.project/tickets/5EEKMD-harden-gh-fallback-credentials/test-definitions.md
+++ b/.project/tickets/5EEKMD-harden-gh-fallback-credentials/test-definitions.md
@@ -8,6 +8,8 @@
 | TD-2 | Preserve real child environment | Supply a lookup-only environment; set a process-only marker; mocked `gh` returns a valid credential | Resolution succeeds and `gh` receives the process marker but no `GITHUB_TOKEN` |
 | TD-3 | Preserve explicit `GH_TOKEN` | Rejected `GITHUB_TOKEN`; set `GH_TOKEN`; mocked `gh` returns a valid credential | `gh` child environment includes `GH_TOKEN` and excludes `GITHUB_TOKEN` |
 | TD-4 | Isolate mocks between tests | Reset mocks in `afterEach` | Every subprocess-dependent test explicitly sets its own response |
+| TD-5 | Accept one terminal line ending only | Mocked `gh` returns a valid credential followed by one LF or CRLF | Resolution succeeds; leading/trailing whitespace and repeated line endings remain rejected |
+| TD-6 | Remove token keys case-insensitively | Lookup-only rejected token; process context has a differently cased `github_token` | The `gh` child environment excludes differently cased `GITHUB_TOKEN` keys |
 
 ## Completion
 
@@ -15,5 +17,7 @@
 - [x] TD-2 — added RED→GREEN process-boundary coverage with a process-only marker.
 - [x] TD-3 — retained and passed the #1602 child-environment regression.
 - [x] TD-4 — added a regression proving no earlier mock response can leak.
+- [x] TD-5 — added terminal LF/CRLF acceptance and malformed-whitespace rejection coverage.
+- [x] TD-6 — added a case-insensitive Windows child-environment regression.
 
-Each new behavior test is added and run red before its implementation changes. The test scope is integration because it validates the real resolver-to-transport behavior and mocks only the external subprocess boundary.
+The test scope is integration because it validates the real resolver-to-transport behavior and mocks only the external subprocess boundary.
--- a/.project/tickets/5EEKMD-harden-gh-fallback-credentials/test-definitions.md
+++ b/.project/tickets/5EEKMD-harden-gh-fallback-credentials/test-definitions.md
@@ -7,7 +7,7 @@
 | TD-1 | Reject unsafe `gh` output | Rejected `GITHUB_TOKEN`; mocked `gh` returns a value with a space | `resolveGitHubToken()` is undefined and `createRestTransport()` is undefined |
 | TD-2 | Preserve real child environment | Supply a lookup-only environment; set a process-only marker; mocked `gh` returns a valid credential | Resolution succeeds and `gh` receives the process marker but no `GITHUB_TOKEN` |
 | TD-3 | Preserve explicit `GH_TOKEN` | Rejected `GITHUB_TOKEN`; set `GH_TOKEN`; mocked `gh` returns a valid credential | `gh` child environment includes `GH_TOKEN` and excludes `GITHUB_TOKEN` |
-| TD-4 | Isolate mocks between tests | Reset mocks in `afterEach` | Every subprocess-dependent test explicitly sets its own response |
+| TD-4 | Reset mock implementations during test cleanup | Configure a `gh` response, then run the shared cleanup | The next fallback call receives no retained subprocess response |
 | TD-5 | Accept one terminal line ending only | Mocked `gh` returns a valid credential followed by one LF or CRLF | Resolution succeeds; leading/trailing whitespace and repeated line endings remain rejected |
 | TD-6 | Remove token keys case-insensitively | Lookup-only rejected token; process context has a differently cased `github_token` | The `gh` child environment excludes differently cased `GITHUB_TOKEN` keys |
 
--- a/.project/tickets/5EEKMD-harden-gh-fallback-credentials/test-definitions.md
+++ b/.project/tickets/5EEKMD-harden-gh-fallback-credentials/test-definitions.md
@@ -1,0 +1,19 @@
+# Test Definitions: Protect retro filing from malformed GitHub CLI credentials
+
+**Scope:** Integration tests in `packages/cli/src/retro/github-rest.test.ts`, using the real resolver and REST transport while mocking only the `gh` subprocess boundary.
+
+| ID | Behavior | Setup | Assertion |
+| --- | --- | --- | --- |
+| TD-1 | Reject unsafe `gh` output | Rejected `GITHUB_TOKEN`; mocked `gh` returns a value with a space | `resolveGitHubToken()` is undefined and `createRestTransport()` is undefined |
+| TD-2 | Preserve real child environment | Supply a lookup-only environment; set a process-only marker; mocked `gh` returns a valid credential | Resolution succeeds and `gh` receives the process marker but no `GITHUB_TOKEN` |
+| TD-3 | Preserve explicit `GH_TOKEN` | Rejected `GITHUB_TOKEN`; set `GH_TOKEN`; mocked `gh` returns a valid credential | `gh` child environment includes `GH_TOKEN` and excludes `GITHUB_TOKEN` |
+| TD-4 | Isolate mocks between tests | Reset mocks in `afterEach` | Every subprocess-dependent test explicitly sets its own response |
+
+## Completion
+
+- [x] TD-1 — added RED→GREEN integration coverage at the resolver-to-transport boundary.
+- [x] TD-2 — added RED→GREEN process-boundary coverage with a process-only marker.
+- [x] TD-3 — retained and passed the #1602 child-environment regression.
+- [x] TD-4 — added a regression proving no earlier mock response can leak.
+
+Each new behavior test is added and run red before its implementation changes. The test scope is integration because it validates the real resolver-to-transport behavior and mocks only the external subprocess boundary.
--- a/.project/tickets/5EEKMD-harden-gh-fallback-credentials/ticket.md
+++ b/.project/tickets/5EEKMD-harden-gh-fallback-credentials/ticket.md
@@ -1,0 +1,43 @@
+---
+id: 5EEKMD
+slug: harden-gh-fallback-credentials
+type: task
+phase: implement
+status: in_progress
+created: 2026-07-29T17:33:42.554Z
+last_modified: 2026-07-29T17:39:00.000Z
+external_issue: https://github.com/ArcadeAI/safeword/issues/1637
+external_prs:
+  - https://github.com/ArcadeAI/safeword/pull/1577
+---
+
+# Protect retro filing from malformed GitHub CLI credentials
+
+**Goal:** Keep gh-based retro filing header-safe without prompting users for a token.
+
+**Why:** Malformed gh output and environment coupling can make automatic filing fail or send unusable credentials.
+
+## Scope
+
+- Accept only RFC 6750 Bearer-syntax output from `gh auth token`.
+- Run `gh auth token` with the real process environment, excluding only the rejected `GITHUB_TOKEN`.
+- Reset subprocess mocks after each test and simplify the one-row parameterized test.
+
+## Out of Scope
+
+- Prompting for, storing, or creating GitHub credentials.
+- Guessing whether opaque `GH_TOKEN` values are valid; GitHub remains the authority.
+
+## Done When
+
+- A malformed `gh` result cannot create a REST transport or Authorization header.
+- An injected lookup environment cannot remove process context needed to invoke `gh`.
+- The focused and repository verification lanes pass with isolated mocks.
+
+## Work Log
+
+- 2026-07-29T17:33:42.554Z Started: Created ticket 5EEKMD
+- 2026-07-29T17:39:00.000Z Defined: Captured the late #1577 review findings as #1637 scope and acceptance criteria.
+- 2026-07-29T17:42:14.000Z Implemented: Added RED→GREEN coverage for unsafe gh output, lookup-environment isolation, and mock reset; simplified the single-case test table.
+- 2026-07-29T17:48:00.000Z Reviewed: Fresh independent quality review approved the implementation with no blocking findings.
+- 2026-07-29T18:15:00.000Z Verified: Focused tests, canonical test plan, BDD acceptance lane, lint, formatting, and audit completed; recorded evidence in verify.md.
--- a/.project/tickets/5EEKMD-harden-gh-fallback-credentials/ticket.md
+++ b/.project/tickets/5EEKMD-harden-gh-fallback-credentials/ticket.md
@@ -2,10 +2,10 @@
 id: 5EEKMD
 slug: harden-gh-fallback-credentials
 type: task
-phase: implement
-status: in_progress
+phase: done
+status: done
 created: 2026-07-29T17:33:42.554Z
-last_modified: 2026-07-29T17:39:00.000Z
+last_modified: 2026-07-30T00:46:48Z
 external_issue: https://github.com/ArcadeAI/safeword/issues/1637
 external_prs:
   - https://github.com/ArcadeAI/safeword/pull/1577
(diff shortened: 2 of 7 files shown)


packages/cli/src/retro/github-rest.ts
```diff
@@ -49,19 +50,24 @@ const MAX_DEDUP_PAGES = 200;
...
       timeout: 10_000,
       env: childEnvironment,
     });
-    const token = (result.stdout ?? '').trim();
-    return result.status === 0 && token.length > 0 ? token : undefined;
+    // Accept at most one terminal LF or CRLF. Broad trimming could turn malformed
+    // credential output into a token-shaped value by silently discarding
+    // whitespace or control characters.
+    const token = (result.stdout ?? '').replace(/\r?\n$/, '');
+    return result.status === 0 && isBearerCredentialSyntax(token) ? token : undefined;
```
TheMostlyGreat, line 74: Reviewer: NOTE — **the `proxy-injected` sentinel still reaches the Authorization header through `GH_TOKEN`, and the code no longer reads as if that were deliberate.**

Reproduced live this session:

```
$ GH_TOKEN=proxy-injected gh auth token | od -c
0000000    p   r   o   x   y   -   i   n   j   e   c   t   e   d  \n
```

That survives `replace(/\r?\n$/, '')`, and `isBearerCredentialSyntax('proxy-injected')` is `true` (hyphen and ALPHA are both in `b64token`) — so it returns here and lands in `` `Bearer proxy-injected` ``. `GH_TOKEN` > `GITHUB_TOKEN` precedence is documented, and go-gh's `TokenForHost` reads env before config, so `gh` echoes it verbatim. (verified: https://cli.github.com/manual/gh_help_environment and https://raw.githubusercontent.com/cli/go-gh/trunk/pkg/auth/auth.go, fetched this session; plus the local run above)

This is the exact asymmetry the `github-rest.ts:64` thread on #1577 raised — the one that ended "the value #634 exists to keep out of the API still reaches it, just through the sibling variable." **Issue #1637 explicitly decided against `GH_TOKEN` placeholder heuristics, so I read this as knowingly accepted, not a defect.** Real-world impact is low: the value is a documented public placeholder, and the resulting 401 is latched by `isTerminalStatus`, costing one sweep rather than amplifying.

The problem is only that the thread is now marked "Addressed in #1647" while the doc comment above (lines 47-51) reads as though the sentinel is universally rejected. Two ways to close the record honestly:

- **Cheapest:** one line in the doc comment — "an explicit `GH_TOKEN` carrying the `proxy-injected` placeholder is still returned; #1637 keeps `GH_TOKEN` opaque and lets GitHub's 401 be the authority." Next reader stops re-litigating it.
- **Or close it:** `token !== 'proxy-injected' && isBearerCredentialSyntax(token)`. Worth noting the #1637 rationale ("don't guess whether `GH_TOKEN` is valid") doesn't quite cover this — an exact-literal match on a documented sentinel is not a validity heuristic, and it is already what the `GITHUB_TOKEN` path does 30 lines down. Your call; I'd take the comment.

packages/cli/src/retro/github-rest.ts
```diff
@@ -49,19 +50,24 @@ const MAX_DEDUP_PAGES = 200;
...
-): string | undefined {
+function ghAuthToken(): string | undefined {
   try {
-    const childEnvironment = { ...environment };
-    delete childEnvironment.GITHUB_TOKEN;
+    // Node's Windows environment keys are case-insensitive, while this copied
+    // JavaScript object has case-sensitive keys. Exclude every casing so a
+    // rejected token cannot re-enter `gh` under a different key spelling.
+    const childEnvironment = Object.fromEntries(
+      Object.entries(process.env).filter(([key]) => key.toUpperCase() !== GITHUB_TOKEN_ENV_KEY),
```
TheMostlyGreat: Reviewer: nit, take or leave — this comparison silently depends on `GITHUB_TOKEN_ENV_KEY` already being uppercase. It is today, so the filter is correct, but a future rename to a mixed-case constant would make every key survive the filter and the failure is invisible: `gh` just receives the rejected token back.

`key.toUpperCase() !== GITHUB_TOKEN_ENV_KEY.toUpperCase()` costs nothing and removes the trap. A trailing comment on the constant declaration works too.

Related asymmetry, also harmless today: the strip here is case-*insensitive* while the lookup at line 101 (`env[GITHUB_TOKEN_ENV_KEY]`) is case-*sensitive*. Correct for `process.env` on both platforms; only diverges for a hand-built injected object, which no caller passes.

packages/cli/src/retro/github-rest.test.ts
```diff
@@ -650,11 +653,56 @@ describe('resolveGitHubToken (7D8PJP — no hard GITHUB_TOKEN requirement)', ()
...
+
+  it('removes differently cased rejected token keys before asking gh', () => {
+    vi.stubEnv('github_token', 'proxy-injected');
+    spawnSyncMock.mockReturnValue({ status: 0, stdout: 'gh-keyring-token\n' });
+
+    expect(resolveGitHubToken({ GITHUB_TOKEN: 'proxy-injected' })).toBe('gh-keyring-token');
+    expect(getGhChildEnvironment()).not.toHaveProperty('github_token');
+  });
+
+  it('starts without a gh response configured by an earlier test', () => {
```
TheMostlyGreat: Reviewer: **this test is a real guard, but it passes for a reason it doesn't assert — and it goes vacuous under `--shuffle`.**

It does catch the leak it was written for: a surviving `mockReturnValue` would make this return `'gh-keyring-token'` and fail. Credit where due. Two things weaken it:

1. **It passes via a swallowed exception.** With the mock reset, `spawnSync` returns `undefined`, `(result.stdout ?? '')` throws a `TypeError`, and `ghAuthToken`'s bare `catch {}` eats it. So `toBeUndefined()` is also satisfied by any *other* breakage that yields undefined — the assertion can't distinguish "mock was reset" from "resolver threw."
2. **It only detects a leak because three preceding tests happen to set `mockReturnValue`.** That's declaration-order coupling with nothing naming it. Reorder the file, or run with `--shuffle`, and it silently stops testing anything.

One line fixes (1) and documents the mechanism:

```ts
expect(spawnSyncMock).toHaveBeenCalledTimes(1);
```

plus a comment noting it must stay *after* a test that configures a response. Cheap insurance on a test whose whole job is to prove isolation.

.project/tickets/5EEKMD-harden-gh-fallback-credentials/verify.md
```diff
@@ -0,0 +1,19 @@
+## Verify Checklist
+
+**Test Suite:** ⚠️ Local environment limitation: 58/58 focused resolver/transport tests pass; the canonical suite under shared load timed out in two unrelated integration tests, both of which pass in isolation
+**Gherkin:** ✅ Acceptance lane passes (`bun run test:bdd`: 93 scenarios, 1,109 steps)
+**Build:** ✅ Success (`tsup`)
+**Lint:** ✅ Clean (`bun run lint`; TypeScript typecheck and Prettier check pass)
+**Scenarios:** All 6 scenarios marked complete
+**PR Scope:** ✅ Diff matches ticket scope
```
TheMostlyGreat, line 8: Reviewer: **this checkbox is false as written — the diff carries two refactors that are in neither #1637's Decision list nor the ticket's Scope.**

In `createRestTransport`: `matchesIn` → `findMarkerMatches` (`github-rest.ts:311`) and `createdThisRun: { number; title; body }[]` → `OpenIssue[]` (`:309`). Both are pure cleanups in the dedup path, unrelated to credential handling, and not among the four #1577 threads this PR exists to close.

I think both changes are *improvements* — `OpenIssue` is the same shape and `findMarkerMatches` reads better. The issue is only the claim: a reviewer trusting "Diff matches ticket scope" won't look for them. Either drop them into their own patch, or add a Scope bullet and keep the ✅ honest.

.project/tickets/5EEKMD-harden-gh-fallback-credentials/test-definitions.md
```diff
@@ -0,0 +1,23 @@
+# Test Definitions: Protect retro filing from malformed GitHub CLI credentials
+
+**Scope:** Integration tests in `packages/cli/src/retro/github-rest.test.ts`, using the real resolver and REST transport while mocking only the `gh` subprocess boundary.
+
+| ID | Behavior | Setup | Assertion |
+| --- | --- | --- | --- |
+| TD-1 | Reject unsafe `gh` output | Rejected `GITHUB_TOKEN`; mocked `gh` returns a value with a space | `resolveGitHubToken()` is undefined and `createRestTransport()` is undefined |
+| TD-2 | Preserve real child environment | Supply a lookup-only environment; set a process-only marker; mocked `gh` returns a valid credential | Resolution succeeds and `gh` receives the process marker but no `GITHUB_TOKEN` |
+| TD-3 | Preserve explicit `GH_TOKEN` | Rejected `GITHUB_TOKEN`; set `GH_TOKEN`; mocked `gh` returns a valid credential | `gh` child environment includes `GH_TOKEN` and excludes `GITHUB_TOKEN` |
+| TD-4 | Isolate mocks between tests | Reset mocks in `afterEach` | Every subprocess-dependent test explicitly sets its own response |
```
TheMostlyGreat: Reviewer: minor — **TD-4's assertion column contradicts the test it describes.**

"Every subprocess-dependent test explicitly sets its own response" is the opposite of what `starts without a gh response configured by an earlier test` does: it deliberately sets *none*, and that absence is the whole assertion. Anyone reading TD-4 as the spec would look for the wrong test.

The behavior TD-4 actually pins is closer to: "a test that configures no `gh` response observes no response from an earlier test."

TheMostlyGreat (commented):
Reviewer: **/quality-review — APPROVE. No correctness or security defect found. Two passes (mine + a fresh independent reviewer) converged on the same verdict.**

Severity split: **zero blockers, one accepted-by-design NOTE, four artifact/test-hygiene nits.** The credential hardening itself is right; every remaining item is documentation or evidence accuracy.

### What I verified sound — don't re-chase these

- **The header-injection surface is fully closed.** `buildCall` is the only `` `Bearer ${token}` `` construction site, both `createRestTransport` and `createReconcileTransport` compose it, and every input now passes `isBearerCredentialSyntax`. The regex `/^[\w.~+/-]+=*$/` matches RFC 6750 §2.1 `b64token = 1*( ALPHA / DIGIT / "-" / "." / "_" / "~" / "+" / "/" ) *"="` character-for-character — `\w` without `/u` is exactly `[A-Za-z0-9_]`. (verified: https://datatracker.ietf.org/doc/html/rfc6750#section-2.1, fetched this session)
- **JS `$` has no Python-style trailing-newline leniency**, so the anchor is strict end-of-input. Ran the table locally: `'tok\n'`, `'tok\r'`, `'tok\r\n'`, `'a b'`, `'tok '`, `' tok'`, `'ghp_a\tb'`, `'%41'`, `''` → all rejected. After the terminal strip, `'tok\n\n'` → `'tok\n'` → rejected, `'tok\r\r\n'` → `'tok\r'` → rejected. No CRLF path survives. (verified: node run this session)
- **The accept path matches real `gh` output.** `gh auth token` writes `fmt.Fprintf(opts.IO.Out, "%s\n", val)` — exactly one LF, always — and the "new release" notice goes to **stderr**, so stdout is never polluted. Nothing real-world is wrongly rejected by the narrow `replace(/\r?\n$/, '')`. (verified: https://raw.githubusercontent.com/cli/cli/trunk/pkg/cmd/auth/token/token.go and .../internal/ghcmd/cmd.go, fetched this session)
- **The #1577 `env`-conflation regression is genuinely fixed.** `ghAuthToken` now reads real `process.env`, so `PATH`/`HOME` survive a lookup-only override.
- **The signature change breaks nothing.** Every caller is zero-arg: `src/commands/retro.ts:401`, `:465`, `tests/smoke/reconcile.live.test.ts:48`, `tests/smoke/retro-dedup.live.test.ts:53`; `tests/commands/retro.test.ts:32` mocks the export. Every in-file test seam passes a zero-param lambda, still assignable.
- **Child-env construction is correct on Windows.** `Object.entries(process.env)` preserves stored casing while the `process.env` proxy is case-insensitive, so `key.toUpperCase() !== 'GITHUB_TOKEN'` strips every spelling. Values are always strings at runtime, and `spawnSync` **omits** an `undefined` env value rather than stringifying it — no `KEY=undefined` leak. (verified: node run this session)
- **`vi.clearAllMocks()` → `vi.resetAllMocks()` weakens nothing.** `spawnSyncMock` is `vi.hoisted(() => vi.fn())` with no baked-in implementation, there is no `beforeEach`, and fetch mocks are created per-test by `vi.stubGlobal` and torn down by the preceding `vi.unstubAllGlobals()`.
- **The RED proof is real, not asserted.** I ran the head test file against `main`'s source: **5 of the new tests fail** (the four `does not build a REST transport from gh output with …` rows plus `uses process context when a lookup-only environment falls back to gh`), 53 pass. TD-1 and TD-2's RED→GREEN claims hold.
- **Objective check: 58/58 green** on `packages/cli/src/retro/github-rest.test.ts` at head, and all four CI checks pass on `8459dcc`.

### Evidence accuracy (top-level, no line to anchor to)

The PR body's **"focused resolver/transport suite: 53/53" is wrong — the real number is 58/58.** `verify.md` has it right. 53 is the count that *passes against the pre-change source* (see the RED proof above); the base commit `2f9e29f6d` had 50 tests, so 53 matches neither side. Worth correcting since it is the headline validation claim.

Inline comments follow: one accepted-by-design NOTE on the `GH_TOKEN` path, one naming nit, one test-hygiene item, two artifact-accuracy items. Nothing blocks the merge.

_Posted as `COMMENT` — `gh` authenticates as the PR author, so this carries no merge block._

.project/tickets/5EEKMD-harden-gh-fallback-credentials/ticket.md
```diff
@@ -0,0 +1,43 @@
+---
+id: 5EEKMD
+slug: harden-gh-fallback-credentials
+type: task
+phase: implement
```
TheMostlyGreat: Reviewer: pre-merge reconcile — **`phase: implement` / `status: in_progress` while this ticket's own `verify.md` is fully filled in and the work log records a `Verified:` entry.**

Correct for a draft PR, so not a finding *now*. Flagging it because the repo convention for finished tickets is `phase: done` / `status: done` — that's what `ZCYD5P`, `ZRW21K`, `ZD70P1`, `ZJMZ50`, `ZRXM6Q` all carry. Merging as-is leaves a ticket that says the work is unfinished next to evidence that it is.

Worth noting the done-flip guard won't catch this for you: it's warn-mode, and `Dogfood parity` passed on `8459dcc`. Flip it when you undraft.

.project/tickets/5EEKMD-harden-gh-fallback-credentials/verify.md
```diff
@@ -0,0 +1,19 @@
+## Verify Checklist
+
+**Test Suite:** ⚠️ Local environment limitation: 58/58 focused resolver/transport tests pass; the canonical suite under shared load timed out in two unrelated integration tests, both of which pass in isolation
+**Gherkin:** ✅ Acceptance lane passes (`bun run test:bdd`: 93 scenarios, 1,109 steps)
+**Build:** ✅ Success (`tsup`)
+**Lint:** ✅ Clean (`bun run lint`; TypeScript typecheck and Prettier check pass)
+**Scenarios:** All 6 scenarios marked complete
```
TheMostlyGreat: Reviewer: wording, lowest priority — "scenarios" here means the six `TD-` rows in `test-definitions.md`, not Gherkin scenarios. There is no `harden-gh-fallback-credentials.feature`, which is fine for a `task`-type ticket on the legacy `test-definitions.md` path.

It reads as a Gherkin claim next to the line above it, which *is* one (`**Gherkin:** ✅ Acceptance lane passes`). "All 6 test definitions complete" removes the ambiguity.

Not false, just loose — take it or leave it.

TheMostlyGreat (commented):
Reviewer: **completeness pass — three artifact items from the same /quality-review that weren't in the first review. Verdict is unchanged: APPROVE, still zero blockers.**

Posting them so the PR is the complete record rather than the chat transcript. Two are inline below; the third has no line in the diff to anchor to:

**Ticket `INDEX.md` gets no entry for `5EEKMD`.** `.project/tickets/INDEX.md` is auto-generated by `safeword sync-tickets`, is git-tracked, and every recent ticket appears in it (`ZCYD5P`, `ZRW21K`, `Z10A9Q`, `ZFGWS1` all present). `5EEKMD` is absent at head, so the new ticket ships undiscoverable from the index.

**I'd defer this rather than fix it here.** `INDEX.md` on `main` is already markdown-corrupted — it carries three stacked `## Tickets (458)` / `(457)` / `(455)` headers, the known out-of-tree-CLI version-skew corruption from #397. Regenerating in this PR would drag that unrelated churn into a credential-hardening diff. Cleaner as its own patch that regenerates from a tree-source `sync-tickets`, which would pick up `5EEKMD` along with whatever else has drifted.

With the four checks green on `8459dcc` and 58/58 on the focused suite, nothing here changes the merge decision.

packages/cli/src/retro/github-rest.ts
TheMostlyGreat: Addressed in b75e2f0d0: the resolver now explicitly documents that `proxy-injected` is rejected only from direct `GITHUB_TOKEN`; per #1637, syntax-valid explicit `GH_TOKEN` values remain opaque and GitHub authorizes them.

packages/cli/src/retro/github-rest.ts
TheMostlyGreat: Addressed in b75e2f0d0: normalize both sides of the comparison, so a future mixed-case key constant cannot let a rejected token back into the gh child environment.

packages/cli/src/retro/github-rest.test.ts
TheMostlyGreat: Addressed in b75e2f0d0: extracted the exact afterEach cleanup into resetTestState, then test it directly after configuring a response. The assertion proves the next fallback invokes gh without a retained implementation; the shuffled focused suite passes.

.project/tickets/5EEKMD-harden-gh-fallback-credentials/verify.md
TheMostlyGreat: Addressed in b75e2f0d0: reverted the unrelated OpenIssue alias and findMarkerMatches rename. The remaining diff is credential hardening and supporting test/artifact accuracy only.

.project/tickets/5EEKMD-harden-gh-fallback-credentials/test-definitions.md
TheMostlyGreat: Addressed in b75e2f0d0: TD-4 now describes the deterministic shared-cleanup behavior it actually specifies—configure a gh response, run cleanup, and observe no retained response on the next fallback.

.project/tickets/5EEKMD-harden-gh-fallback-credentials/ticket.md
TheMostlyGreat: Deferred intentionally: this PR remains a draft, so ticket 5EEKMD stays phase: implement/status: in_progress. I will flip it to done only when the PR is made ready for delivery.

.project/tickets/5EEKMD-harden-gh-fallback-credentials/verify.md
TheMostlyGreat: Addressed in b75e2f0d0: verify.md now says `All 6 test definitions complete`, separating that artifact count from the Gherkin acceptance result above it.

PR #1685, opened 2026-07-30
Reuse monorepo discovery during architecture healing

TheMostlyGreat:
Closes #1667

## What changed

- Build one operation-scoped monorepo architecture snapshot and reuse its model for root rendering and leaf enumeration.
- Extract each leaf skeleton once, then reuse it for fingerprinting, matching, and rendering.
- Carry the already-observed workspace-root state through the zero-leaf fallback so workspace managers are not probed again.
- Add filesystem-boundary regression tests for readable, unreadable, zero-leaf, and source-purpose discovery.

## Why

Architecture healing independently rediscovered workspace topology and leaf structure at several stages. Besides redundant filesystem work, those reads could observe different filesystem states during one operation.

## Impact

Architecture output remains byte-identical: package ordering, dependency edges, fingerprints, coverage gaps, nested/polyglot behavior, and zero-leaf behavior are preserved. The heal path now observes each relevant manager or source boundary once.

## Validation

- 5,631 Vitest tests passed; 5 skipped
- 499/502 Cucumber scenarios passed; 3 skipped
- 15,444 executed Cucumber steps passed; 4 skipped
- Focused architecture suite: 191 tests passed
- ESLint, Prettier, TypeScript, build, dependency-cruiser, Knip, and Safeword audit passed
- Fresh independent quality review: approved after resolving its zero-leaf re-probe finding

### DIFF
--- a/packages/cli/src/utils/architecture-document.ts
+++ b/packages/cli/src/utils/architecture-document.ts
@@ -141,11 +141,7 @@ function planTarget(target: HealTarget): SelfHealAction {
 }
 
 /** A `src/`-skeleton doc: the skeleton of `directory`, rendered to `path`. */
-function skeletonTarget(
-  directory: string,
-  path: string,
-  skeleton = extractSkeleton(directory),
-): HealTarget {
+function skeletonTarget(directory: string, path: string, skeleton: Skeleton): HealTarget {
   const fingerprint = shapeFingerprintOf(directory, skeleton);
   const nodes = skeleton.nodes;
   return {
--- a/packages/cli/src/utils/architecture-document.ts
+++ b/packages/cli/src/utils/architecture-document.ts
@@ -159,8 +159,12 @@ function skeletonTarget(
 }
 
 /** The single-repo doc: the project's `src/` skeleton at the namespace-root path. */
-function singleRepoTarget(projectDirectory: string): HealTarget {
-  return skeletonTarget(projectDirectory, resolveGeneratedArchitecturePath(projectDirectory));
+function singleRepoTarget(projectDirectory: string, workspaceRoot = false): HealTarget {
+  return skeletonTarget(
+    projectDirectory,
+    resolveGeneratedArchitecturePath(projectDirectory),
+    extractSkeleton(projectDirectory, { workspaceRoot }),
+  );
 }
 
 /** A colocated leaf: the package's own skeleton at `packages/<pkg>/architecture.generated.md`. */
@@ -190,13 +194,13 @@ function rootIndexTarget(projectDirectory: string, model: MonorepoModel): HealTa
 
 /** The targets a project heals: single-repo → one; monorepo → root index + per-leaf. */
 function projectTargets(projectDirectory: string): HealTarget[] {
-  const { model, leaves } = extractMonorepoArchitectureSnapshot(projectDirectory);
+  const { model, leaves, workspaceRoot } = extractMonorepoArchitectureSnapshot(projectDirectory);
   // A repo whose ONLY workspace signal is an unparseable manager (zero discovered leaves)
   // is still a monorepo we must not mistake for a single-repo: render the root index so the
   // "config unreadable" advisory has a home, rather than silently emitting a single-repo doc
   // that omits the whole declared-but-unreadable workspace (UWP4XK).
   if (model.packages.length === 0 && model.unreadableWorkspaces.length === 0) {
-    return [singleRepoTarget(projectDirectory)];
+    return [singleRepoTarget(projectDirectory, workspaceRoot)];
   }
   return [
     rootIndexTarget(projectDirectory, model),
--- a/packages/cli/src/utils/architecture-document.ts
+++ b/packages/cli/src/utils/architecture-document.ts
@@ -16,8 +16,6 @@ import nodePath from 'node:path';
 
 import { shapeFingerprint } from './architecture-fingerprint.js';
 import {
-  discoverLeafDirectories,
-  discoverUnreadableWorkspaces,
   extractMonorepoModel,
   monorepoFingerprintOf,
   type MonorepoModel,
@@ -169,8 +167,7 @@ function leafTarget(packageDirectory: string): HealTarget {
 }
 
 /** The derived root index: the package graph at the namespace-root path. */
-function rootIndexTarget(projectDirectory: string): HealTarget {
-  const model = extractMonorepoModel(projectDirectory);
+function rootIndexTarget(projectDirectory: string, model: MonorepoModel): HealTarget {
   const fingerprint = monorepoFingerprintOf(projectDirectory, model);
   return {
     path: resolveGeneratedArchitecturePath(projectDirectory),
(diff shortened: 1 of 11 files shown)


packages/cli/src/utils/architecture-document.ts
```diff
@@ -142,9 +141,13 @@ function planTarget(target: HealTarget): SelfHealAction {
 }
 
 /** A `src/`-skeleton doc: the skeleton of `directory`, rendered to `path`. */
-function skeletonTarget(directory: string, path: string): HealTarget {
-  const fingerprint = shapeFingerprint(directory);
-  const nodes = extractSkeleton(directory).nodes;
+function skeletonTarget(
+  directory: string,
+  path: string,
+  skeleton = extractSkeleton(directory),
```
TheMostlyGreat: Reviewer: **Suggested improvement (elegance/no-bloat).** `skeleton = extractSkeleton(directory)` is a default that's never actually evaluated — both call sites (`singleRepoTarget`, `leafTarget`) always pass it explicitly, and `skeletonTarget` is private with no other callers. Consider making `skeleton` a required parameter; the optional-with-fallback shape is misleading and, if a future caller omitted it, would silently reintroduce the redundant re-extraction this PR removes.

packages/cli/src/utils/architecture-monorepo.ts
```diff
@@ -308,17 +325,33 @@ function hasRecognizedManifest(directory: string): boolean {
 
 /** The package/edge model the root index renders over. */
 export function extractMonorepoModel(projectDirectory: string): MonorepoModel {
```
TheMostlyGreat: Reviewer: **Suggested improvement (no-bloat).** `extractMonorepoModel` and `discoverLeafDirectories` lost their last production call site in this refactor (both were only called from `architecture-document.ts`, now routed through `extractMonorepoArchitectureSnapshot`). They're still correct and tested, but now have zero production callers — kept alive only by their own tests, which `knip` won't flag. Worth an explicit decision: delete them and re-point their tests at the snapshot API, or keep them intentionally as public surface.

TheMostlyGreat (commented):
Reviewer: /quality-review

**Currency:** ✓ No external dependency/API claims — pure internal refactor, no new packages or version-sensitive code.
**Sources:** ✓ N/A — no load-bearing external claims to trace.
**Correct:** ✓ Traced every reuse path (snapshot → root render, skeleton → fingerprint+render+introspection flag). No divergence between the fast (reused) and slow (fresh-extract) paths. The `workspaceRoot` short-circuit in `architecture-skeleton.ts` is strictly conservative versus the old `declaresWorkspaces()` probe — it only skips the re-probe when the caller has already positively confirmed workspace-root status; every other case falls through to the original check unchanged (verified by walking JS/pnpm/go.work/Cargo/uv detection paths and by the new "readable zero-leaf" regression test).
**Elegant:** ⚠️ see inline comment on `architecture-document.ts`.
**No-bloat:** ⚠️ see inline comment on `architecture-monorepo.ts`.
**Wiring (code only):** ✓ `architecture-project.test.ts` adds `vi.mock('node:fs', {spy:true})` — a real-fs spy, not a stub — and asserts exact read/exists counts through real collaborators. Confirmed by hand that each new count assertion would have failed against the pre-PR code (3 reads → 1, 2 probes → 1), so these are genuine regression pins, not vacuous. Isolation is per-file (`pool: 'forks'`, default `isolate: true`), so the fs spy doesn't bleed into sibling suites.

**Verdict:** APPROVE

**Critical issues:** None

**Provenance:** Verified by reading the full diff, running the targeted architecture test suite (180/180 passed) and `tsc --noEmit` (clean) in this session, cross-checked with an independent fresh-context reviewer agent that reached the same conclusions. CI on this PR (Dogfood parity, lint, test on node 22 & 24) is green.

**Next:** Nothing blocking — the two inline suggestions below are optional cleanup, not required for merge.

*(Posted as a plain review comment, not an approval/changes-request — `gh` here authenticates as the PR author, so the GitHub review API blocks self-approval.)*

packages/cli/src/utils/architecture-document.ts
TheMostlyGreat: Resolved in 14616d9ab: `skeletonTarget` now requires the precomputed skeleton, so TypeScript rejects future callers that omit it. The existing project wiring tests and full verification suite are green.

packages/cli/src/utils/architecture-monorepo.ts
TheMostlyGreat: Resolved in 9e0d7f0d0. I expanded the cleanup to the complete test-only compatibility chain: `discoverLeafDirectories`, `extractMonorepoModel`, and `monorepoFingerprint` are removed from production, with transparent snapshot projections kept in the tests. Focused architecture coverage is 191/191; typecheck, Knip, and the full verification suite are green.

TheMostlyGreat (commented):
Reviewer: /quality-review (re-review after update)

Re-reviewed after the two fixes pushed in response to the last review (`14616d9ab` require precomputed skeletons, `9e0d7f0d0` remove legacy monorepo discovery wrappers).

**Correct:** ✓ Both fixes do exactly what they claim, nothing else changed — `git diff 2933038fd..20df06252 --stat` touches only `architecture-document.ts`, `architecture-monorepo.ts`, their test file, and ticket docs.
**Elegant:** ✓ `skeletonTarget`'s `skeleton` param is now required (`architecture-document.ts:144`) — the dead default from the last review is gone; both call sites already passed it explicitly, so this is a pure tightening with no behavior change.
**No-bloat:** ✓ `discoverLeafDirectories`, `extractMonorepoModel`, `monorepoFingerprint` removed from production (`architecture-monorepo.ts`); grepped `src/` for all three names — zero remaining references. Their tests now use local wrappers around `extractMonorepoArchitectureSnapshot`/`monorepoFingerprintOf` instead of importing dead production API — exactly the resolution suggested.
**Wiring:** ✓ unchanged from before, still solid.

**Verdict:** APPROVE

**Critical issues:** None
**Suggested improvements:** None — both prior ones resolved cleanly.

**Provenance:** (verified this session) 220/220 targeted architecture tests pass, `tsc --noEmit` clean, no stray references to removed functions, PR CI (Dogfood parity, lint, test node 22 & 24) all green on this head.

**Next:** Nothing outstanding.

*(Posted as a plain review comment, not an approval/changes-request — `gh` here authenticates as the PR author, so the GitHub review API blocks self-approval.)*

TheMostlyGreat (commented):
Reviewer: /quality-review + /refactor (follow-up pass)

**/quality-review:** No new commits since the last full review — this PR's head (`20df0625`) is exactly what I already reviewed and approved twice; the PR merged as `df31c884f` at that same head. Nothing new to check.

**/refactor scout** (scoped to what this PR actually introduced across `architecture-document.ts`, `architecture-fingerprint.ts`, `architecture-monorepo.ts`, `architecture-skeleton.ts` — not the pre-existing surrounding code in those files):

Ledger (1 entry, Tier 1 / optional):
1. `architecture-monorepo.ts` (`extractMonorepoArchitectureSnapshot`'s edge-building loop) — could be **Extract Function**'d into a small named helper (e.g. `buildPackageEdges(packages)`) for single-responsibility. Purely cosmetic; the function is ~38 lines with three clearly separated stages (discover leaves → build packages → build edges), so this is borderline, not a real "long function" smell. No behavior risk (Extract Function is the near-unconditionally-safe tier).

Nothing else rises to a real smell — no duplication, magic literals, deep nesting, or wrong-abstraction-level introduced by this diff. The two issues that would have been worth flagging (a dead default param, orphaned discovery wrappers) were already caught in the prior quality-review pass and fixed before merge.

Since the PR is merged, applying the one ledger entry would mean a new commit to `main`, not something foldable into this PR — and it's low enough value that I'd skip it unless requested as a follow-up. Report-only, no changes applied.

*(Posted as a plain review comment on the merged PR, for the record.)*

PR #1690, opened 2026-07-30
Show Linear users how to sync through portable transport

TheMostlyGreat:
## Summary

- replace stale Linear projection guidance with the supported `--plan` / `--apply-results` workflow
- add an end-to-end Linear example with precise connect, adoption, create, and update behavior
- make plain Linear live sync point directly to the supported portable path before credential or sidecar checks
- replace stale setup/2TK5AD and version-bound runtime guidance, with command and client regression coverage
- document the validation floor, relative Linear body pointer, and GitHub-only egress warning

## Why

The docs and runtime errors made the working portable Linear path look blocked. This update documents the actual workflow and ensures unsupported live-sync journeys lead directly to supported next steps.

## Impact

Users can configure Linear, adopt an existing issue, export plans for a trusted executor, and apply result artifacts without relying on live Linear projection. Plain Linear `sync-tracker` now exits non-zero with portable-path guidance instead of first asking for a credential the unwired adapter cannot use.

## Validation

- final independent quality review: **APPROVE**, no critical issues
- full Safeword verification: **5,640 Vitest tests passed, 5 skipped**
- Cucumber: **499 scenarios passed, 3 skipped; 15,444 steps passed, 4 skipped**
- focused post-review regression suite, lint, typecheck, CLI build, website build, Prettier, diff checks, and diff audit passed

Closes #1679

### DIFF
--- a/packages/cli/src/commands/sync-tracker.ts
+++ b/packages/cli/src/commands/sync-tracker.ts
@@ -116,7 +116,8 @@ function runApply(cwd: string, config: TicketBridgeConfig, filePath: string): vo
 
 /** An advisory on stderr — never stdout, which must stay a pure SyncPlan document. */
 function note(message: string): void {
-  process.stderr.write(`sync-tracker: ${message}.\n`);
+  const suffix = /[.!?]$/.test(message) ? '' : '.';
+  process.stderr.write(`sync-tracker: ${message}${suffix}\n`);
 }
 
 /**
@@ -171,7 +172,7 @@ async function runLiveSync(
   // Linear has no live adapter yet, so asking for a credential or sidecar first
   // is a dead end. Point directly to the two supported transport paths.
   if (provider === 'linear') {
-    console.log(LINEAR_LIVE_PROJECTION_GUIDANCE);
+    note(LINEAR_LIVE_PROJECTION_GUIDANCE);
     process.exitCode = 1;
     return;
   }
--- a/packages/cli/src/commands/sync-tracker.ts
+++ b/packages/cli/src/commands/sync-tracker.ts
@@ -9,7 +9,11 @@ import { readFileSync } from 'node:fs';
 import process from 'node:process';
 
 import { applyResults } from '../tracker-sync/apply-results.js';
-import { buildWriterRegistry, resolveRepoVisibility } from '../tracker-sync/clients.js';
+import {
+  buildWriterRegistry,
+  LINEAR_LIVE_PROJECTION_GUIDANCE,
+  resolveRepoVisibility,
+} from '../tracker-sync/clients.js';
 import { readTicketBridgeConfig } from '../tracker-sync/config.js';
 import { emptyPlan, parseResults } from '../tracker-sync/contract.js';
 import { readCorpus } from '../tracker-sync/corpus.js';
@@ -153,7 +157,8 @@ export async function syncTrackerCommand(options: SyncTrackerCommandOptions = {}
     return;
   }
 
-  // No flag → the live one-way projection (unchanged gh/Linear path).
+  // No flag → the live one-way projection. GitHub uses its live writer; Linear
+  // reports the portable path until its live adapter is wired.
   await runLiveSync(cwd, config, options);
 }
 
@@ -163,6 +168,13 @@ async function runLiveSync(
   options: SyncTrackerCommandOptions,
 ): Promise<void> {
   const provider = supportedProvider(config.provider);
+  // Linear has no live adapter yet, so asking for a credential or sidecar first
+  // is a dead end. Point directly to the two supported transport paths.
+  if (provider === 'linear') {
+    console.log(LINEAR_LIVE_PROJECTION_GUIDANCE);
+    process.exitCode = 1;
+    return;
+  }
   const dependencies: SyncTrackerDependencies = {
     config,
     tickets: provider === undefined ? [] : readCorpus(cwd, config.target?.repo),
--- a/packages/cli/tests/tracker-sync/clients.test.ts
+++ b/packages/cli/tests/tracker-sync/clients.test.ts
@@ -7,7 +7,7 @@ import { execFileSync } from 'node:child_process';
 import { buildWriterRegistry, resolveRepoVisibility } from '../../src/tracker-sync/clients.js';
 
 /**
- * resolveRepoVisibility (JS5K5G AC10) — locks the public/private collapse and the
+ * Client adapter coverage (JS5K5G) — locks the public/private collapse and the
  * catch→undefined fail-safe the egress warning depends on. child_process is
  * mocked so no real `gh` is invoked.
  */
@@ -106,18 +106,27 @@ describe('GitHub graph projection client', () => {
 });
 
 describe('Linear live projection client', () => {
-  it('points users to the supported adoption and portable-sync paths', async () => {
-    const writer = buildWriterRegistry('linear', undefined).linear;
+  const payload = {
+    title: 'Wire it up',
+    body: 'banner',
+    issueType: 'task',
+    labels: ['type:task'],
+    state: 'open' as const,
+  };
+  const ref = { provider: 'linear' as const, id: 'ENG-45' };
 
-    await expect(
-      writer.create({
-        title: 'Wire it up',
-        body: 'banner',
-        issueType: 'task',
-        labels: ['type:task'],
-        state: 'open',
-      }),
-    ).rejects.toThrow(
+  it.each([
+    ['create', () => buildWriterRegistry('linear', undefined).linear.create(payload)],
+    ['update', () => buildWriterRegistry('linear', undefined).linear.update(ref, payload)],
+    [
+      'projectGraph',
+      () =>
+        buildWriterRegistry('linear', undefined).linear.projectGraph(ref, payload, {
+          blockedBy: [],
+        }),
+    ],
+  ])('points %s callers to the supported adoption and portable-sync paths', async (_name, call) => {
+    await expect(call()).rejects.toThrow(
       'Adopt an existing Linear issue with `safeword ticket new <slug> --issue <key>`, or sync existing local tickets through `safeword sync-tracker --plan` and `--apply-results`.',
     );
   });
--- a/packages/cli/tests/tracker-sync/command-linear.test.ts
+++ b/packages/cli/tests/tracker-sync/command-linear.test.ts
@@ -1,0 +1,51 @@
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import nodePath from 'node:path';
+import process from 'node:process';
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { syncTrackerCommand } from '../../src/commands/sync-tracker.js';
+
+describe('sync-tracker Linear live-sync guidance', () => {
+  let cwd: string;
+  let logs: string[];
+
+  beforeEach(() => {
+    cwd = mkdtempSync(nodePath.join(tmpdir(), 'sync-linear-'));
+    mkdirSync(nodePath.join(cwd, '.safeword'), { recursive: true });
+    writeFileSync(
+      nodePath.join(cwd, '.safeword', 'config.json'),
+      JSON.stringify({ ticketBridge: { provider: 'linear', target: { team: 'ENG' } } }),
+    );
+    logs = [];
+    vi.spyOn(process, 'cwd').mockReturnValue(cwd);
+    vi.spyOn(console, 'log').mockImplementation(message => {
+      logs.push(String(message));
+    });
+    delete process.env.LINEAR_API_KEY;
+    process.exitCode = 0;
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    delete process.env.LINEAR_API_KEY;
+    process.exitCode = 0;
+    rmSync(cwd, { recursive: true, force: true });
+  });
+
+  it.each([
+    ['without a credential', undefined],
+    ['with a credential', 'dummy-linear-token'],
+  ])('points directly to portable sync %s', async (_condition, credential) => {
+    if (credential !== undefined) process.env.LINEAR_API_KEY = credential;
+
+    await syncTrackerCommand();
+
+    const output = logs.join('\n');
+    expect(process.exitCode).toBe(1);
+    expect(output).toContain('safeword sync-tracker --plan');
+    expect(output).toContain('--apply-results');
+    expect(output).not.toMatch(/credential|token/i);
+  });
+});
--- a/packages/cli/tests/tracker-sync/command-linear.test.ts
+++ b/packages/cli/tests/tracker-sync/command-linear.test.ts
@@ -9,7 +9,7 @@ import { syncTrackerCommand } from '../../src/commands/sync-tracker.js';
 
 describe('sync-tracker Linear live-sync guidance', () => {
   let cwd: string;
-  let logs: string[];
+  let stderr: string[];
 
   beforeEach(() => {
     cwd = mkdtempSync(nodePath.join(tmpdir(), 'sync-linear-'));
@@ -18,18 +18,19 @@ describe('sync-tracker Linear live-sync guidance', () => {
       nodePath.join(cwd, '.safeword', 'config.json'),
       JSON.stringify({ ticketBridge: { provider: 'linear', target: { team: 'ENG' } } }),
     );
-    logs = [];
+    stderr = [];
     vi.spyOn(process, 'cwd').mockReturnValue(cwd);
-    vi.spyOn(console, 'log').mockImplementation(message => {
-      logs.push(String(message));
+    vi.spyOn(process.stderr, 'write').mockImplementation(chunk => {
+      stderr.push(String(chunk));
+      return true;
     });
-    delete process.env.LINEAR_API_KEY;
+    vi.stubEnv('LINEAR_API_KEY', '');
     process.exitCode = 0;
   });
 
   afterEach(() => {
     vi.restoreAllMocks();
-    delete process.env.LINEAR_API_KEY;
+    vi.unstubAllEnvs();
     process.exitCode = 0;
     rmSync(cwd, { recursive: true, force: true });
   });
@@ -38,11 +39,11 @@ describe('sync-tracker Linear live-sync guidance', () => {
     ['without a credential', undefined],
     ['with a credential', 'dummy-linear-token'],
   ])('points directly to portable sync %s', async (_condition, credential) => {
-    if (credential !== undefined) process.env.LINEAR_API_KEY = credential;
+    if (credential !== undefined) vi.stubEnv('LINEAR_API_KEY', credential);
 
     await syncTrackerCommand();
 
-    const output = logs.join('\n');
+    const output = stderr.join('');
     expect(process.exitCode).toBe(1);
     expect(output).toContain('safeword sync-tracker --plan');
     expect(output).toContain('--apply-results');
--- a/packages/website/src/content/docs/reference/tracker-integration.mdx
+++ b/packages/website/src/content/docs/reference/tracker-integration.mdx
@@ -31,9 +31,11 @@ safeword connect linear --team ENG
 ```
 
 `connect` writes the `ticketBridge` block to `.safeword/config.json`. GitHub's live connection also
-verifies auth and seeds an empty sidecar. For Linear, the command records the provider/team identity;
-the portable executor owns the network call, and `--plan` can start without a sidecar. Credentials
-stay outside config and are never written to logs.
+verifies auth and seeds an empty sidecar. For Linear, the command records the provider/team identity
+before attempting live verification. Because live Linear verification is not wired yet, it then
+reports `Not connected` and exits 1; the printed Arcade OAuth handoff does not enable live sync in
+this release. The saved config is sufficient for the portable path, and `--plan` can start without a
+sidecar. Credentials stay outside config and are never written to logs.
 
 :::note[GitHub live sync; Linear portable sync]
 **GitHub** supports both the one-command live sync and the portable transport. **Linear** supports
@@ -44,7 +46,9 @@ GitHub-only; use the portable transport with a Linear-capable executor.
 
 With GitHub's live transport, `safeword ticket new <slug>` mints the tracker issue first and keys
 the local ticket to it. For either provider, `--issue <key>` adopts an existing issue instead of
-creating one. New Linear tickets become `create` intents when you generate a portable sync plan.
+creating one. For non-epic Linear tickets, live minting is not wired: `ticket new` without `--issue`
+fails, while `--issue ENG-45` adopts that existing issue and later plans an `update`. A local ticket
+with no recorded tracker ref — for example, one that predates the connect — plans a `create`.
 
 ## Sync status to the tracker
 
@@ -58,7 +62,7 @@ This GitHub path walks your active tickets and creates/updates their issues (and
 when a ticket is terminal). It is safe to run repeatedly. Run it whenever you want the tracker
 refreshed — or automate it in CI so nobody has to remember.
 
-### Automate in CI (GitHub Actions)
+#### Automate in CI (GitHub Actions)
 
 Safeword does not install a workflow for you — copy this into
 `.github/workflows/sync-tracker.yml`:
@@ -113,11 +117,27 @@ Running `sync-tracker` with neither flag is the usual `gh` path, unchanged.
 
 #### Linear example
 
-First record the Linear team identity and generate the provider-neutral plan. These steps do not
-need Safeword to construct a Linear client:
+First record the Linear team identity:
 
 ```bash
 safeword connect linear --team ENG
+```
+
+The command writes the Linear provider and team to `.safeword/config.json`, then reports the
+following result and exits 1 because live verification is not wired:
+
+```text
+Not connected — Linear verification needs the Arcade integration — not wired in v1.
+```
+
+That exit is expected for the portable path; the required config has already landed. The printed
+Arcade OAuth step does not enable live Linear sync in this release. Run `connect linear` once
+outside `set -e` automation, confirm the saved provider and team, then run only the planning step in
+automation.
+
+Now generate the provider-neutral plan; this does not construct a Linear client:
+
+```bash
 safeword sync-tracker --plan > linear-plan.json
 ```
 
@@ -152,9 +172,10 @@ invalid result leaves the map untouched.
 
 ## Egress
 
-The projected issue body defaults to **minimal** (title, status, labels, back-link to the
-canonical ticket) — never your spec or work-log. `body: full` is opt-in per project, and projecting
-`full` to a **public** GitHub repo emits a loud egress warning.
+The projected issue body defaults to **minimal** (title, status, labels, and a pointer to the
+canonical ticket) — never your spec or work-log. For GitHub, that pointer is a clickable repository
+URL; in a Linear plan it is the ticket's relative path inside the repository. `body: full` is opt-in
+per project, and projecting `full` to a **public** GitHub repo emits a loud egress warning.
 
 That automatic public-target warning is GitHub-only. Linear has
 [private teams](https://linear.app/docs/private-teams), but its access boundary is workspace/team
--- a/packages/website/src/content/docs/reference/tracker-integration.mdx
+++ b/packages/website/src/content/docs/reference/tracker-integration.mdx
@@ -87,10 +87,12 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 ```
 
-`sync-tracker` **fails loudly (non-zero exit)** when a provider is configured but no credential
-resolves — it never silently exits 0 and does nothing, so a misconfigured CI job turns red instead
-of quietly drifting. In a non-interactive run authenticated by an `ARCADE_USER_ID` (a user
-identity, not a service account), it also warns that a lapsed grant would otherwise fail silently.
+GitHub live sync **fails loudly (non-zero exit)** when no credential resolves — it never silently
+exits 0 and does nothing, so a misconfigured CI job turns red instead of quietly drifting. Plain
+Linear live sync also exits non-zero, but points directly to the supported portable path instead of
+asking for a credential that the unwired live adapter cannot use. In a non-interactive run
+authenticated by an `ARCADE_USER_ID` (a user identity, not a service account), Safeword also warns
+that a lapsed grant would otherwise fail silently.
 
 ### Portable transport (`--plan` / `--apply-results`)
 
@@ -123,12 +125,8 @@ First record the Linear team identity:
 safeword connect linear --team ENG
 ```
 
-The command writes the Linear provider and team to `.safeword/config.json`, then reports the
-following result and exits 1 because live verification is not wired:
-
-```text
-Not connected — Linear verification needs the Arcade integration — not wired in v1.
-```
+The command writes the Linear provider and team to `.safeword/config.json`, then reports
+`Not connected` and exits 1 because live Linear verification still needs the Arcade integration.
 
 That exit is expected for the portable path; the required config has already landed. The printed
 Arcade OAuth step does not enable live Linear sync in this release. Run `connect linear` once
--- a/packages/website/src/content/docs/reference/tracker-integration.mdx
+++ b/packages/website/src/content/docs/reference/tracker-integration.mdx
@@ -52,6 +52,13 @@ with no recorded tracker ref — for example, one that predates the connect — 
 
 ## Sync status to the tracker
 
+Plain live sync fails loudly (with a non-zero exit) when it cannot proceed. For GitHub, that means
+a missing credential turns CI red instead of silently allowing tracker state to drift. For Linear,
+plain live sync points directly to the supported portable path instead of asking for a credential
+that the unwired live adapter cannot use. In a non-interactive GitHub run authenticated by an
+`ARCADE_USER_ID` (a user identity, not a service account), Safeword also warns that a lapsed grant
+would otherwise fail silently.
+
 ### Live transport (GitHub)
 
 ```bash
@@ -86,13 +93,6 @@ jobs:
           # safeword reads the credential from GITHUB_TOKEN (not GH_TOKEN).
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 ```
-
-GitHub live sync **fails loudly (non-zero exit)** when no credential resolves — it never silently
-exits 0 and does nothing, so a misconfigured CI job turns red instead of quietly drifting. Plain
-Linear live sync also exits non-zero, but points directly to the supported portable path instead of
-asking for a credential that the unwired live adapter cannot use. In a non-interactive run
-authenticated by an `ARCADE_USER_ID` (a user identity, not a service account), Safeword also warns
-that a lapsed grant would otherwise fail silently.
 
 ### Portable transport (`--plan` / `--apply-results`)
 
--- a/packages/website/src/content/docs/reference/tracker-integration.mdx
+++ b/packages/website/src/content/docs/reference/tracker-integration.mdx
@@ -30,29 +30,33 @@ safeword connect github --repo owner/name
 safeword connect linear --team ENG
 ```
 
-`connect` writes the `ticketBridge` block to `.safeword/config.json`, verifies auth, and seeds an
-empty sidecar. Credentials are read from your OS keychain or an environment variable — **never**
-from config and never written to logs.
+`connect` writes the `ticketBridge` block to `.safeword/config.json`. GitHub's live connection also
+verifies auth and seeds an empty sidecar. For Linear, the command records the provider/team identity;
+the portable executor owns the network call, and `--plan` can start without a sidecar. Credentials
+stay outside config and are never written to logs.
 
-:::note[GitHub today; Linear projection is pending]
-**GitHub** is fully wired end-to-end. **Linear** identity works — `connect linear` and adopting an
-existing issue with `--issue` need no network — but `sync-tracker` to Linear is **not yet wired**
-(it errors with an actionable message, tracked by ticket 2TK5AD). Use GitHub for the live
-status mirror today.
+:::note[GitHub live sync; Linear portable sync]
+**GitHub** supports both the one-command live sync and the portable transport. **Linear** supports
+tracker identity (`connect linear` and adopting an existing issue with `--issue`) plus sync through
+`--plan` / `--apply-results`. Only the built-in live transport behind plain `sync-tracker` is
+GitHub-only; use the portable transport with a Linear-capable executor.
 :::
 
-Once connected, `safeword ticket new <slug>` mints the tracker issue first and keys the local
-ticket to it; `--issue <key>` adopts an existing issue instead of creating one.
+With GitHub's live transport, `safeword ticket new <slug>` mints the tracker issue first and keys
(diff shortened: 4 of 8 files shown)


packages/website/src/content/docs/reference/tracker-integration.mdx
```diff
@@ -30,29 +30,33 @@ safeword connect github --repo owner/name
...
-empty sidecar. Credentials are read from your OS keychain or an environment variable — **never**
-from config and never written to logs.
-
-:::note[GitHub today; Linear projection is pending]
-**GitHub** is fully wired end-to-end. **Linear** identity works — `connect linear` and adopting an
-existing issue with `--issue` need no network — but `sync-tracker` to Linear is **not yet wired**
-(it errors with an actionable message, tracked by ticket 2TK5AD). Use GitHub for the live
-status mirror today.
+`connect` writes the `ticketBridge` block to `.safeword/config.json`. GitHub's live connection also
+verifies auth and seeds an empty sidecar. For Linear, the command records the provider/team identity;
```
TheMostlyGreat: Reviewer: **CRITICAL** — "the command records the provider/team identity" is true but omits the part the reader will actually see. Live run on this base:

```
$ safeword connect linear --team ENG
...
Not connected — Linear verification needs the Arcade integration — not wired in v1.
EXIT CODE: 1
```

`verify.ts:27-31` hardcodes `ok: false` for Linear and `index.ts:82-85` makes that terminal. The config write at `index.ts:71` happens first, so the portable path does work — but the user has to be told to expect the failure, or they stop here (and `set -e` / CI stops for them).

Suggested: name the output and the exit code, and say plainly that it is expected and non-fatal for the portable path — the `ticketBridge` block is already written by then. Worth also noting the printed Arcade OAuth link is not yet a working step.

packages/website/src/content/docs/reference/tracker-integration.mdx
```diff
@@ -30,29 +30,33 @@ safeword connect github --repo owner/name
...
+tracker identity (`connect linear` and adopting an existing issue with `--issue`) plus sync through
+`--plan` / `--apply-results`. Only the built-in live transport behind plain `sync-tracker` is
+GitHub-only; use the portable transport with a Linear-capable executor.
 :::
 
-Once connected, `safeword ticket new <slug>` mints the tracker issue first and keys the local
-ticket to it; `--issue <key>` adopts an existing issue instead of creating one.
+With GitHub's live transport, `safeword ticket new <slug>` mints the tracker issue first and keys
+the local ticket to it. For either provider, `--issue <key>` adopts an existing issue instead of
+creating one. New Linear tickets become `create` intents when you generate a portable sync plan.
```
TheMostlyGreat: Reviewer: **CRITICAL** — this is inverted. Under a Linear config, `safeword ticket new <slug>` without `--issue` fails outright:

```
Failed to create ticket: Linear projection needs the Arcade integration — run `safeword setup` to wire it (tracked by 2TK5AD).
EXIT CODE: 1   # no folder written
```

`creation-mode.ts:13,34` routes non-epic Linear tickets to `create`, which calls the `linearNotWired()` throwing client (`clients.ts:170-175,191`). With `--issue` it adopts, and the recorded ref then plans as **`update`**, not `create` (verified in live plan output).

Where `create` intents actually come from on Linear: tickets that already exist with **no recorded ref** — e.g. a corpus predating the connect. That's the more useful sentence anyway, since it tells an adopting team their existing backlog all plans as `create` on the first run.

packages/website/src/content/docs/reference/tracker-integration.mdx
```diff
@@ -30,29 +30,33 @@ safeword connect github --repo owner/name
...
 ```
 
-This walks your active tickets and creates/updates their issues (and closes the issue when a
-ticket is terminal). It is safe to run repeatedly. Run it whenever you want the tracker refreshed —
-or automate it in CI so nobody has to remember.
+This GitHub path walks your active tickets and creates/updates their issues (and closes the issue
+when a ticket is terminal). It is safe to run repeatedly. Run it whenever you want the tracker
+refreshed — or automate it in CI so nobody has to remember.
 
 ### Automate in CI (GitHub Actions)
```
TheMostlyGreat: Reviewer: **Refactor nit** (non-blocking) — now that `### Live transport (GitHub)` exists, this CI recipe is a peer of it but is really *part of* it: the workflow runs plain `sync-tracker`. As `#### Automate in CI (GitHub Actions)` nested under Live transport, the two-transport split would read cleanly top to bottom. Everything else in the heading tree is right.

packages/website/src/content/docs/reference/tracker-integration.mdx
```diff
@@ -107,12 +111,58 @@ issue's number and back-link, and is idempotent — re-applying the same file ch
 flags are mutually exclusive, and a malformed results file is rejected without touching the map.
 Running `sync-tracker` with neither flag is the usual `gh` path, unchanged.
 
+#### Linear example
+
+First record the Linear team identity and generate the provider-neutral plan. These steps do not
+need Safeword to construct a Linear client:
+
+```bash
+safeword connect linear --team ENG
```
TheMostlyGreat, line 125: Reviewer: **CRITICAL** (same root cause as the note on line 34) — this block reads as a working two-command sequence, but the first command exits 1 with `Not connected`. In a script or CI step the second line never runs.

The fix is small: either show the expected output inline, or add one sentence after the block — "`connect linear` reports `Not connected` and exits 1 because live Linear verification isn't wired; the `ticketBridge` block is written regardless, which is all the portable transport needs."

The second command is fine — I ran `sync-tracker --plan` against a Linear config with no sidecar present and it produced a valid plan, exactly as documented.

packages/website/src/content/docs/reference/tracker-integration.mdx
```diff
@@ -107,12 +111,58 @@ issue's number and back-link, and is idempotent — re-applying the same file ch
...
+}
+```
+
+Save that document as `linear-results.json`, then record the Linear reference in the local map:
+
+```bash
+safeword sync-tracker --apply-results linear-results.json
+```
+
+Despite the field name `number`, Linear issue keys such as `ENG-45` are valid strings.
```
TheMostlyGreat, line 165: Reviewer: ✓ **Confirmed accurate, line by line** — this paragraph matches `apply-results.ts:37-66,82-84` exactly: the provider-neutral blank-key and non-HTTP(S) floor, the corpus-membership check, the GitHub-only numeric + URL-tail guards, and validate-before-mutate. I also ran the exact `ENG-45` results document above: exit 0, correct `recorded` ref written to the sidecar. Nothing to change here.

packages/website/src/content/docs/reference/tracker-integration.mdx
```diff
@@ -107,12 +111,58 @@ issue's number and back-link, and is idempotent — re-applying the same file ch
...
+numeric issue-number and URL-tail checks. Validation finishes before the map is changed, so one
+invalid result leaves the map untouched.
+
 ## Egress
 
 The projected issue body defaults to **minimal** (title, status, labels, back-link to the
 canonical ticket) — never your spec or work-log. `body: full` is opt-in per project, and projecting
 `full` to a **public** GitHub repo emits a loud egress warning.
 
+That automatic public-target warning is GitHub-only. Linear has
```
TheMostlyGreat, line 178: Reviewer: ✓ Correct and the best addition in this PR — `sync-tracker.ts:55` does gate the plan-time egress warning on `provider === 'github'`, so a Linear `body: full` plan warns about nothing. Flagging that the plan file itself carries the bodies is the right call.

**NOTE** worth folding into this paragraph: for Linear the minimal body's "back-link to the canonical ticket" (line 155) isn't a link at all. `backLinkUrl()` in `corpus.ts` only builds a URL from `target.repo`, and a Linear config has `team` and no `repo` — live Linear plan output is `safeword ticket ENG-45 → .project/tickets/ENG-45-login-bug`. Non-blocking, but this is the paragraph where a Linear reader would look for it.

TheMostlyGreat (commented):
Reviewer: **/quality-review + /refactor** — verified against the source **and against live CLI runs this session** (built `dist` from this base, ran the documented commands in a sandbox project).

**Verdict: REQUEST CHANGES** (posted as `COMMENT` — `gh` authenticates as the PR author, so this carries no merge block).

The portable-transport half of this PR is accurate and genuinely valuable — I confirmed every claim about `--plan` / `--apply-results` by running them. The problem is the two **entry-point** steps: both commands the guide opens with fail with exit code 1 on a Linear project today, and the guide now reads as if they succeed.

| | |
|---|---|
| **Currency** | ✓ Claims checked against this base commit's source, not training data |
| **Sources** | ⚠️ Portable-transport claims all trace to source + live runs; the two connect/create claims contradict them |
| **Correct** | ❌ Two verified doc-vs-behavior mismatches at the first two steps a Linear user takes |
| **Elegant** | ✓ Two-transport split is the right frame; heading tree is sound |
| **No-bloat** | ✓ Nothing here is padding — the Linear example earns its length |

---

### Critical

**1. `safeword connect linear --team ENG` reports failure and exits 1.**

```
$ safeword connect linear --team ENG
Authorize safeword in Arcade (OAuth) and approve in your browser:
  https://api.arcade.dev/oauth/authorize
Arcade holds the Linear token; safeword stores only the Arcade key.
Not connected — Linear verification needs the Arcade integration — not wired in v1.
EXIT CODE: 1
```

`createVerifyClient()` hardcodes `ok: false` for Linear (`packages/cli/src/tracker-connect/verify.ts:27-31`), and `connectTracker` treats a failed verify as terminal (`packages/cli/src/tracker-connect/index.ts:82-85`). The config **is** written first (`index.ts:71`) — I confirmed `.safeword/config.json` lands correctly — so the portable path really does work afterward. But a reader following this guide sees `Not connected`, and under `set -e` or in CI the next command never runs. That is the exact "working path looks blocked" confusion this PR set out to remove; it moves it from the callout into the terminal.

The printed handoff also sends them to `https://api.arcade.dev/oauth/authorize` for an integration that isn't wired — a dead end the guide doesn't warn about.

**2. "New Linear tickets become `create` intents" is inverted.** Under a Linear config, `safeword ticket new <slug>` without `--issue` hard-fails:

```
$ safeword ticket new login-bug --type task
Failed to create ticket: Linear projection needs the Arcade integration — run `safeword setup` to wire it (tracked by 2TK5AD).
EXIT CODE: 1        # no ticket folder written
```

`resolveCreationMode` routes a non-epic Linear ticket to `mode: 'create'` (`creation-mode.ts:13,34`), which calls the live writer — and `buildWriterRegistry` hands Linear a `linearNotWired()` throwing client (`clients.ts:170-175,191`). With `--issue ENG-45` it adopts, but then the ref is already recorded, so the plan emits **`update`**, not `create` (confirmed in the live plan output).

`create` intents do exist for Linear — just not from new tickets. They come from tickets that **already existed with no recorded ref**, e.g. a corpus that predates the connect. That's the sentence worth writing, and it's a more useful one: it tells an adopting team their whole backlog plans as `create` on the first run.

### Note

**3. Linear back-links are relative paths, not URLs.** `backLinkUrl()` only builds a URL when `target.repo` is set (`corpus.ts`), and a Linear config carries `team`, no `repo`. Live plan output for a Linear project:

```
safeword ticket ENG-45 → .project/tickets/ENG-45-login-bug
```

The Egress section describes the minimal body as carrying a "back-link to the canonical ticket" — for Linear it isn't clickable. One sentence in the new Egress paragraph would cover it.

### Follow-up (out of scope here — worth its own issue)

**4. The runtime error points at a closed ticket and a non-remedy.** `clients.ts:173` tells the user to "run `safeword setup` to wire it (tracked by 2TK5AD)" — but `2TK5AD` is `status: done`, and `setup` cannot wire Linear. This PR correctly stops the *docs* citing 2TK5AD as pending; the message a Linear user actually hits still does.

### Confirmed correct (checked, not assumed)

- The `--apply-results` validation paragraph matches `apply-results.ts:37-66,82-84` **exactly** — provider-neutral blank/non-HTTP(S) floor, corpus membership, GitHub-only numeric + URL-tail guards, validate-before-mutate. Precise and well-earned.
- `--plan` is genuinely provider-neutral and tolerates a missing sidecar (`sync-tracker.ts:62-72`) — verified with no `tracker-map.json` present.
- The doc's exact `ENG-45` results document round-trips: exit 0, correct `recorded` ref in the sidecar.
- `create` / `update` / `close` intent kinds are right (`plan.ts:80-84`).
- The GitHub-only egress-warning caveat is correct (`sync-tracker.ts:55` gates on `provider === 'github'`), and flagging that the plan file itself carries full bodies is the strongest addition in this PR. Keep it.

### Refactor

Structure is sound — the live/portable split is the right cut and the `####` nesting of the Linear example is correct. One nit inline on the CI heading.

**Next:** revise the two entry-point claims (state the expected `Not connected` output and exit 1 for `connect linear`, and re-aim the `create`-intent sentence at the pre-existing corpus), then this is an approve. Findings 3 and 4 are optional here.

packages/website/src/content/docs/reference/tracker-integration.mdx
TheMostlyGreat: Reviewer: **CRITICAL fix — committable.** Replaces the inverted `create`-intent claim.

Verified by live run on this base: under a Linear config, `ticket new` without `--issue` fails (`Failed to create ticket: Linear projection needs the Arcade integration…`, exit 1, no folder written) because `creation-mode.ts:13,34` routes it to the `linearNotWired()` throwing client (`clients.ts:170-175,191`). With `--issue` it adopts, and the recorded ref then plans as **`update`**, not `create`.

The rewrite also captures the more useful fact: an adopting team's pre-connect backlog is exactly where `create` intents come from.

```suggestion
With GitHub's live transport, `safeword ticket new <slug>` mints the tracker issue first and keys
the local ticket to it. On Linear, minting is not wired — `ticket new` without `--issue` fails, so
key new tickets to an existing issue with `--issue ENG-45` (this works on either provider). Tickets
that predate the connect carry no tracker ref, so your first portable plan projects them all as
`create` intents.
```

packages/website/src/content/docs/reference/tracker-integration.mdx
```diff
@@ -107,12 +111,58 @@ issue's number and back-link, and is idempotent — re-applying the same file ch
...
 
+#### Linear example
+
+First record the Linear team identity and generate the provider-neutral plan. These steps do not
+need Safeword to construct a Linear client:
+
+```bash
+safeword connect linear --team ENG
+safeword sync-tracker --plan > linear-plan.json
+```
```
TheMostlyGreat: Reviewer: **CRITICAL fix — committable.** Adds the missing caveat under the command block.

Verified by live run: `safeword connect linear --team ENG` prints `Not connected — Linear verification needs the Arcade integration — not wired in v1.` and **exits 1** (`verify.ts:27-31` hardcodes `ok: false` for Linear; `index.ts:82-85` makes it terminal). The `ticketBridge` block *is* written first at `index.ts:71` — I confirmed `.safeword/config.json` lands correctly — so the portable path works fine afterward. Without this sentence the reader stops at step 1, and a CI step stops for them.

````suggestion
First record the Linear team identity and generate the provider-neutral plan. These steps do not
need Safeword to construct a Linear client:

```bash
safeword connect linear --team ENG
safeword sync-tracker --plan > linear-plan.json
```

`connect linear` writes the `ticketBridge` block, then reports `Not connected` and exits 1 because
live Linear verification is not wired yet. That is expected — the config lands before the check,
which is all the portable transport needs. A `set -e` script must not treat it as fatal.
````

**Review by TheMostlyGreat — COMMENTED** [review]
Reviewer: **Follow-up to the review above — the two criticals as committable suggestions.**

Re-checked at head `b8b2d0c8` (unchanged since my last pass; `main` moved two commits, neither touching `tracker-sync` / `tracker-connect` / `ticket-create`, so every finding still stands).

Both suggestions below are ready to click **Commit suggestion**. I applied them locally and ran `npx prettier --check` on the result — **passes**, so the PR's own validation step stays green. `proseWrap` is unset in `.prettierrc` (defaults to `preserve`), so the rewrapped prose is safe.

Still open, unchanged from the review above:

- **NOTE** — Linear back-links render as bare relative paths, not URLs (`corpus.ts` `backLinkUrl()` needs `target.repo`; a Linear config carries `team`). Live output: `safeword ticket ENG-45 → .project/tickets/ENG-45-login-bug`. One sentence in the Egress paragraph would cover it — I left it as a suggestion rather than a patch since the wording depends on whether you'd rather fix the behavior.
- **Follow-up issue** — `clients.ts:173` still tells users to "run `safeword setup` to wire it (tracked by 2TK5AD)", but `2TK5AD` is `status: done` and `setup` cannot wire Linear. That's the message a Linear user actually hits. Worth its own issue rather than scope creep here — say the word and I'll open it.

Everything else I verified stays confirmed: the `--apply-results` validation paragraph matches `apply-results.ts` line for line, `--plan` is genuinely provider-neutral and sidecar-tolerant, the `ENG-45` example round-trips (exit 0, correct sidecar), and the GitHub-only egress caveat is correct and the best thing in this PR.

File: packages/website/src/content/docs/reference/tracker-integration.mdx (hunk 1 above)
Comment (TheMostlyGreat) [reply packages/website/src/content/docs/reference/tracker-integration.mdx]: Addressed in e300f1c68: the guide now states the exact `Not connected` result and exit 1, explains that config is written first, warns that the printed OAuth handoff does not enable live sync, and separates portable planning from connect-time failure handling.

File: packages/website/src/content/docs/reference/tracker-integration.mdx (hunk 2 above)
Comment (TheMostlyGreat) [reply packages/website/src/content/docs/reference/tracker-integration.mdx]: Addressed in e300f1c68. The guide now qualifies this as non-epic Linear creation, says adoption with `--issue` plans an `update`, and defines `create` by the real condition: a local ticket with no recorded tracker ref.

File: packages/website/src/content/docs/reference/tracker-integration.mdx (hunk 3 above)
Comment (TheMostlyGreat) [reply packages/website/src/content/docs/reference/tracker-integration.mdx]: Addressed in e300f1c68: the GitHub Actions recipe is now nested under the GitHub live-transport section.

File: packages/website/src/content/docs/reference/tracker-integration.mdx (hunk 4 above)
Comment (TheMostlyGreat) [reply packages/website/src/content/docs/reference/tracker-integration.mdx:125]: Addressed in e300f1c68. Connect and planning are now separate steps; the exact exit-1 output is shown, and `set -e` automation is told to run only planning after one-time connect/config confirmation.

File: packages/website/src/content/docs/reference/tracker-integration.mdx (hunk 5 above)
Comment (TheMostlyGreat) [reply packages/website/src/content/docs/reference/tracker-integration.mdx:165]: Re-verified during the follow-up quality review and live reproduction. The validation paragraph remains unchanged and accurate; resolving this informational thread.

File: packages/website/src/content/docs/reference/tracker-integration.mdx (hunk 6 above)
Comment (TheMostlyGreat) [reply packages/website/src/content/docs/reference/tracker-integration.mdx:178]: Addressed in e300f1c68: the egress section now distinguishes GitHub's clickable repository URL from Linear's in-repo relative ticket path, while retaining the GitHub-only warning and plan-file caveat.

File: packages/website/src/content/docs/reference/tracker-integration.mdx (hunk 2 above)
Comment (TheMostlyGreat) [reply packages/website/src/content/docs/reference/tracker-integration.mdx]: Addressed in e300f1c68 with two extra accuracy guards from independent review: the failure is qualified as non-epic, and `create` is tied to the absence of a recorded tracker ref rather than age alone.

File: packages/website/src/content/docs/reference/tracker-integration.mdx (hunk 7 above)
Comment (TheMostlyGreat) [reply packages/website/src/content/docs/reference/tracker-integration.mdx]: Addressed in e300f1c68. I did not apply the suggestion verbatim because it still left `set -e` behavior ambiguous; the final version separates the steps and tells automation to run only planning after one-time config setup.

File: packages/cli/tests/tracker-sync/clients.test.ts [hunk 8]
```diff
@@ -104,3 +104,21 @@ describe('GitHub graph projection client', () => {
...
+        body: 'banner',
+        issueType: 'task',
+        labels: ['type:task'],
+        state: 'open',
+      }),
+    ).rejects.toThrow(
+      'Adopt an existing Linear issue with `safeword ticket new <slug> --issue <key>`, or sync existing local tickets through `safeword sync-tracker --plan` and `--apply-results`.',
+    );
+  });
+});
```
TheMostlyGreat, line 133: Reviewer: **Refactor (optional, committable).** `linearNotWired()` returns a throwing client for all three methods, so `update` and `projectGraph` carry this same guidance — but only `create` is pinned. Now that the docs quote this string, locking the whole surface is cheap insurance.

**Compile-verified before posting**: applied on `e300f1c68`, then `prettier --write`, `tsc --noEmit` (pass), `eslint` (pass), `vitest` — **38/38 pass** (up from 36; the two new cases are the added coverage). Deliberately uses `ReturnType<typeof linearWriter>` and inline literals so it needs **no new imports** — the block is self-contained and commits green as-is.

```suggestion
describe('Linear live projection client', () => {
  const guidance =
    'Adopt an existing Linear issue with `safeword ticket new <slug> --issue <key>`, or sync existing local tickets through `safeword sync-tracker --plan` and `--apply-results`.';
  const linearWriter = () => buildWriterRegistry('linear', undefined).linear;
  const payload = {
    title: 'Wire it up',
    body: 'banner',
    issueType: 'task',
    labels: ['type:task'],
    state: 'open' as const,
  };
  const ref = { provider: 'linear' as const, id: 'ENG-45' };

  // Every method of the not-wired client carries this guidance and the docs quote
  // it, so lock the whole surface rather than `create` alone.
  it.each<[string, (writer: ReturnType<typeof linearWriter>) => Promise<unknown>]>([
    ['create', writer => writer.create(payload)],
    ['update', writer => writer.update(ref, payload)],
    ['projectGraph', writer => writer.projectGraph(ref, payload, { blockedBy: [] })],
  ])(
    'points %s callers to the supported adoption and portable-sync paths',
    async (_method, call) => {
      await expect(call(linearWriter())).rejects.toThrow(guidance);
    },
  );
});
```

packages/website/src/content/docs/reference/tracker-integration.mdx
```diff
@@ -107,11 +115,74 @@ issue's number and back-link, and is idempotent — re-applying the same file ch
...
+
+```bash
+safeword connect linear --team ENG
+```
+
+The command writes the Linear provider and team to `.safeword/config.json`, then reports the
+following result and exits 1 because live verification is not wired:
+
+```text
+Not connected — Linear verification needs the Arcade integration — not wired in v1.
```
TheMostlyGreat: Reviewer: **NOTE — new coupling with nothing holding it.** This reproduces `verify.ts:30` character for character. I confirmed the match against a live run, so it's correct *today*.

The problem is what holds it there. Grepping `tests/`, `src/`, and `website/src` for `not wired in v1` returns exactly **two** hits: this line and the source. No test asserts that message — `verify.ts`'s string has no coverage at all, unlike the `clients.ts` message you just locked in this PR.

So whoever rewords `verify.ts` gets a green suite and a silently stale doc — the same drift that made this PR necessary, one layer down.

Two ways out, both fine: pin the message in a `tests/tracker-connect/verify.test.ts` (mirroring the lock you added for `clients.ts`), or paraphrase here instead of quoting verbatim so there's nothing to drift. Non-blocking either way.

Minor and separate: `not wired in v1` bakes a version claim into a string that outlives the version.

TheMostlyGreat (commented):
Reviewer: **Full /quality-review + /refactor of `e300f1c68`** (the increment since `b8b2d0c8`, which is what I reviewed before).

**All five prior findings are resolved.** I re-verified each one by running the commands, not by reading the diff.

| Check | | |
|---|---|---|
| **Currency** | ✓ | Verified against `e300f1c68` source + live CLI runs today; `main` moved two commits, neither touching tracker code |
| **Sources** | ✓ | Every load-bearing claim traced to source and reproduced at runtime |
| **Correct** | ✓ | The two criticals are genuinely fixed, and the new prose is *more* precise than what I suggested (see below) |
| **Elegant** | ✓ | Heading nesting fixed; the error message now names both supported paths |
| **No-bloat** | ✓ | Nothing padded — the expected-output block earns its space |
| **Wiring** | ✓ | `clients-visibility.test.ts:108` builds the writer through the real `buildWriterRegistry` and asserts the thrown message — real collaborator, only `child_process` mocked at the boundary. No internal-seam mocking, no `provider: none` short circuit. |

**Verdict: APPROVE.** Critical issues: **None.** What remains is one NOTE and two refactor nits, none blocking.

### Verified fixed

1. **`connect linear` exit-1** — documented in both places. The `text` block at line 130 matches the runtime string **character for character**; I diffed it against a live run.
2. **`create`-intent claim** — corrected, and *better than my suggestion*: scoping it to "non-epic Linear tickets" is a precision I missed. `creation-mode.ts:25` does route epics to `local`, so they never touch the writer. Good catch.
3. **Back-link** — the GitHub-URL vs Linear-relative-path distinction is accurate (`corpus.ts` `backLinkUrl()`).
4. **Stale `2TK5AD` message** — replaced with actionable guidance. Verified both paths it surfaces on:
   - `ticket new` without `--issue` → `Failed to create ticket: Linear live projection is not wired. Adopt an existing…`
   - plain `sync-tracker` (credential present) → same guidance
5. **`###` → `####` for Automate in CI** — applied; the anchor is text-derived so it's unchanged, and I grepped for `automate-in-ci` cross-references (none).

### Objective checks (run on `e300f1c68`)

- `prettier --check` on all three changed files — **pass**
- `tsc --noEmit` (packages/cli) — **pass**
- `eslint` on both changed code files — **pass**
- `vitest tests/tracker-sync/clients-visibility.test.ts` — **36/36 pass**
- PR CI — **all green** (lint, test node 22.22.3, test node 24, Dogfood parity)

One transparency note: I could not reproduce the PR body's `bun run build` (packages/website). It fails locally with `The requested module 'neotraverse' does not provide an export named 'forEach'` — but **that reproduces identically on `origin/main`**, so it's a local dependency-resolution issue, **not a regression from this PR**. There's no website-build job in CI, so nothing gates on it. Flagging only so the green CI isn't over-read.

### NOTE — the docs now verbatim-quote a runtime string that no test pins

Line 130 reproduces `verify.ts:30`'s message exactly. I grepped `tests/`, `src/`, and `website/src` for `not wired in v1`: **two hits — the source and the doc, with nothing binding them.** No test asserts that message anywhere.

So editing `verify.ts`'s wording silently staleness the doc — which is the same drift class that produced #1679, now reintroduced one layer down. Cheap fix, either:

- assert the message in a `tests/tracker-connect/verify.test.ts` (mirrors the lock you just added for `clients.ts`), or
- paraphrase in the doc instead of quoting verbatim, so there's nothing to drift.

Worth noting `verify.ts`'s message has **zero** test coverage today, unlike its `clients.ts` sibling — so this is really a gap the PR half-closed.

### NOTE — the new signpost is unreachable on the likeliest journey

Plain `sync-tracker` on a Linear project with **no credential** hits the AC2 credential gate at `index.ts:196-200` first:

```
No linear credential resolved — set the token in your keychain or env. Aborting.
```

The user is sent to obtain a token that can never work; only after setting one (`LINEAR_API_KEY=dummy` — verified) do they reach the new "not wired → use `--plan`" guidance. The ordering is pre-existing and **not** introduced here, but the thing being hidden is this PR's own signpost, so it's the natural follow-up. Happy to open an issue.

### Refactor

**1. Test placement drift** (nit) — `clients-visibility.test.ts` now holds `resolveRepoVisibility`, `GitHub graph projection client`, **and** `Linear live projection client`. The filename stopped describing the contents two blocks ago. Rename to `clients.test.ts`, or split the Linear block out.

**2. The message contract is only half-locked** — see the inline suggestion. `linearNotWired()` throws the same guidance from all three methods, but only `create` is pinned. Now that the docs quote this message, locking the full surface is worth three lines.

**Next:** merge as-is — nothing here blocks. The inline suggestion is compile-verified and optional; the `verify.ts` message lock is the one I'd actually spend a follow-up on.

packages/website/src/content/docs/reference/tracker-integration.mdx
TheMostlyGreat: Addressed in d6384d883. The guide now paraphrases the `Not connected` result instead of duplicating the full runtime string, `verify.ts` uses the timeless `not wired yet` wording, and a direct semantic test pins the Arcade/unavailable behavior without making every copy edit a contract change.

packages/cli/tests/tracker-sync/clients.test.ts
TheMostlyGreat: Addressed in d6384d883. The adapter suite is now `clients.test.ts`, and an `it.each` table exercises `create`, `update`, and `projectGraph` through the real `buildWriterRegistry → createWriter → throwingClient` chain. The full verification suite passed (5,640 tests, 5 skipped), along with lint and typecheck.

packages/website/src/content/docs/reference/tracker-integration.mdx
```diff
@@ -79,22 +87,24 @@ jobs:
...
 
-`sync-tracker` **fails loudly (non-zero exit)** when a provider is configured but no credential
-resolves — it never silently exits 0 and does nothing, so a misconfigured CI job turns red instead
-of quietly drifting. In a non-interactive run authenticated by an `ARCADE_USER_ID` (a user
-identity, not a service account), it also warns that a lapsed grant would otherwise fail silently.
+GitHub live sync **fails loudly (non-zero exit)** when no credential resolves — it never silently
+exits 0 and does nothing, so a misconfigured CI job turns red instead of quietly drifting. Plain
+Linear live sync also exits non-zero, but points directly to the supported portable path instead of
+asking for a credential that the unwired live adapter cannot use. In a non-interactive run
+authenticated by an `ARCADE_USER_ID` (a user identity, not a service account), Safeword also warns
```
TheMostlyGreat: Reviewer: **NOTE — this sentence no longer holds for Linear, and it now sits right after a Linear sentence.**

I ran both providers against `d6384d883`:

```
Linear + CI=1 ARCADE_USER_ID=user_123 LINEAR_API_KEY=dummy
  → portable-path guidance only. No advisory.

GitHub + CI=1 ARCADE_USER_ID=user_123 GITHUB_TOKEN=dummy
  → "Warning: Arcade-User-ID is a user identity, not a service account…"
```

The new early return in `runLiveSync` lands before `syncTracker()`, so `emitAdvisories` never runs for Linear. The guard itself is provider-agnostic — but `ARCADE_USER_ID` is the Arcade credential, and `handoff.ts:24` says "Arcade holds the Linear token." So the advisory's only real audience was Linear, and it now fires only for GitHub-plus-Arcade, which nobody configures.

The code change is fine — an unwired adapter has no grant to lapse. The problem is reading order: this sentence follows "Plain Linear live sync also exits non-zero…" in the same paragraph, so it naturally scans as describing the Linear case, which is now exactly backwards.

Scoping it is enough — something like "In a non-interactive **GitHub** run authenticated by an `ARCADE_USER_ID`…". Non-blocking.

packages/website/src/content/docs/reference/tracker-integration.mdx
```diff
@@ -79,22 +87,24 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 ```
 
-`sync-tracker` **fails loudly (non-zero exit)** when a provider is configured but no credential
-resolves — it never silently exits 0 and does nothing, so a misconfigured CI job turns red instead
-of quietly drifting. In a non-interactive run authenticated by an `ARCADE_USER_ID` (a user
-identity, not a service account), it also warns that a lapsed grant would otherwise fail silently.
+GitHub live sync **fails loudly (non-zero exit)** when no credential resolves — it never silently
```
TheMostlyGreat: Reviewer: **Refactor nit — heading scope.** This paragraph now documents Linear failure semantics, but it lives under `#### Automate in CI (GitHub Actions)`, itself nested under `### Live transport (GitHub)`. Linear behavior is two levels deep inside GitHub-scoped headings, so a Linear reader scanning the ToC has no reason to look here.

Same family as the `###` → `####` nit from the last round, just inverted — that one was depth, this one is scope. The fix is cheap: the paragraph is really about *general* fail-loud exit semantics for both providers, so it could sit at the `## Sync status to the tracker` level, above the transport split, with only the `GITHUB_TOKEN` specifics staying in the CI subsection.

Non-blocking, and reasonable to leave if you'd rather not reshuffle.

packages/cli/src/commands/sync-tracker.ts
```diff
@@ -163,6 +168,13 @@ async function runLiveSync(
   options: SyncTrackerCommandOptions,
 ): Promise<void> {
   const provider = supportedProvider(config.provider);
+  // Linear has no live adapter yet, so asking for a credential or sidecar first
+  // is a dead end. Point directly to the two supported transport paths.
+  if (provider === 'linear') {
+    console.log(LINEAR_LIVE_PROJECTION_GUIDANCE);
```
TheMostlyGreat: Reviewer: **Refactor nit — this is an error path writing to stdout.** `console.log` plus `process.exitCode = 1`, while this same file has `note()` writing advisories to stderr, with the comment "stdout stays a pure SyncPlan document."

Not a regression: the credential message you replaced also reached stdout through the injected `log`, so this is consistent with its sibling. But it's new code, and the convention it matches is the weaker one. `note()` would keep the stream discipline intact and cost nothing.

If you leave it, it stays consistent with `runLiveSync`'s existing logging — genuinely defensible either way.

packages/cli/tests/tracker-sync/command-linear.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    writeFileSync(
+      nodePath.join(cwd, '.safeword', 'config.json'),
+      JSON.stringify({ ticketBridge: { provider: 'linear', target: { team: 'ENG' } } }),
+    );
+    logs = [];
+    vi.spyOn(process, 'cwd').mockReturnValue(cwd);
+    vi.spyOn(console, 'log').mockImplementation(message => {
+      logs.push(String(message));
+    });
+    delete process.env.LINEAR_API_KEY;
```
TheMostlyGreat: Reviewer: **Refactor nit — restore-safe env.** This deletes `LINEAR_API_KEY` without saving a pre-existing value, so a developer running the suite with a real key in their shell loses it for the rest of that worker process. Forked workers contain the blast radius, which is why this is a nit and not a finding.

`vi.stubEnv('LINEAR_API_KEY', …)` with `vi.unstubAllEnvs()` in `afterEach` restores whatever was there, and lets you drop both `delete` lines. Fits how the rest of the block already leans on `vi.restoreAllMocks()`.

TheMostlyGreat (commented):
Reviewer: **Full /quality-review + /refactor of `d6384d883`** (the increment since `e300f1c68`).

Both NOTEs and both refactor nits from my last pass are resolved — and #2 was fixed at the source rather than in prose, which is the better call.

| Check | | |
|---|---|---|
| **Currency** | ✓ | Verified against `d6384d883` with live CLI runs today |
| **Sources** | ✓ | Every claim reproduced at runtime, not read off the diff |
| **Correct** | ✓ | One behavior consequence worth a look (below), no correctness defect |
| **Elegant** | ✓ | Extracting `LINEAR_LIVE_PROJECTION_GUIDANCE` gives one owner for a string three call sites and the docs depend on |
| **No-bloat** | ✓ | Both new test files earn their place |
| **Wiring** | ✓ | `command-linear.test.ts` drives the real `syncTrackerCommand()` against a real temp-dir config, mocking only `process.cwd`/`console.log` — a genuine wiring test, not an internal-seam mock. Its `not.toMatch(/credential\|token/i)` negative assertion locks the actual property rather than the message text. |

**Verdict: APPROVE.** Critical issues: **None.**

### Verified fixed

1. **Untested verbatim quote** — solved from both ends: the `text` block is gone (paraphrased to `Not connected`), *and* `verify.test.ts` pins the message. Matching on `/Arcade integration.*not wired yet/i` rather than the exact string is the right call — it locks the concept and leaves wording free. `not wired in v1` → `not wired yet` also picks up the version-coupling aside.
2. **Credential gate ordering** — fixed in the code, not documented around. Verified live: plain `sync-tracker` on Linear now prints the portable-path guidance and exits 1 with no credential prompt, both with and without `LINEAR_API_KEY`.
3. **Test placement** — renamed to `clients.test.ts`, docblock updated to match.
4. **`it.each` over all three methods** — applied; inlining the registry call per case sidesteps the type import more simply than my version did.

### Objective checks (on `d6384d883`)

- `prettier --check` on all 7 changed files — **pass**
- `tsc --noEmit` — **pass**
- `eslint` on all changed source and test files — **pass**
- `vitest tests/tracker-sync/ tests/tracker-connect/` — **28 files, 193 tests, all pass**
- PR CI — **all green** (lint, node 22.22.3, node 24, Dogfood parity)
- `--plan` still returns a valid Linear plan (`update` intent for an adopted `ENG-45`) — the early return correctly sits in `runLiveSync` only

### NOTE — the early return makes the Arcade-User-ID advisory unreachable for Linear

Verified by running both providers:

```
Linear + CI=1 ARCADE_USER_ID=user_123 LINEAR_API_KEY=dummy
  → portable-path guidance only. No advisory.

GitHub + CI=1 ARCADE_USER_ID=user_123 GITHUB_TOKEN=dummy
  → "Warning: Arcade-User-ID is a user identity, not a service account…"
```

`emitAdvisories` (`index.ts:100-105`) is provider-agnostic, but `ARCADE_USER_ID` is the *Arcade* credential concept — and per `handoff.ts:24`, "Arcade holds the Linear token." Its only coherent audience was Linear. The new early return lands before `syncTracker()` is called, so Linear never reaches it, and the advisory now fires only for GitHub-plus-Arcade — a combination that doesn't occur.

**The honest counterpoint:** live Linear sync can't run at all, so a warning about *its* grant lapsing has nothing left to warn about. Skipping it is defensible. I'm not asking you to restore it.

The actionable half is the doc — see the inline comment on line 94.

Worth flagging separately: the AC12 test at `orchestrator.test.ts:270` still passes, because it calls `syncTracker()` directly with the default `provider: 'github'`. Green unit test over a path the CLI can no longer reach for the provider it was written for — the wiring-gap shape, on pre-existing coverage rather than anything you added.

### Refactor

**1. Linear semantics are documented two levels inside GitHub-scoped headings** — inline on line 90.

**2. The guidance goes to stdout, not stderr** — inline on `sync-tracker.ts:174`.

**3. Env hygiene in the new test** — inline on `command-linear.test.ts:26`.

**Next:** merge. Nothing here blocks; the line-94 doc scope is the only one I'd bother with, and it's a half-sentence.

packages/website/src/content/docs/reference/tracker-integration.mdx
TheMostlyGreat: Fixed in `3ed468d8b`. The advisory is now explicitly scoped to a non-interactive GitHub run, so it no longer reads as Linear behavior.

packages/website/src/content/docs/reference/tracker-integration.mdx
TheMostlyGreat: Fixed in `3ed468d8b`. I moved the provider-wide fail-loud semantics to the `Sync status to the tracker` level above the GitHub/portable transport split.

packages/cli/src/commands/sync-tracker.ts
TheMostlyGreat: Fixed in `3ed468d8b`. Linear live-path guidance now goes through the stderr `note()` path; the test was changed first to assert stderr and failed before the implementation update.

packages/cli/tests/tracker-sync/command-linear.test.ts
TheMostlyGreat: Fixed in `3ed468d8b`. The test now uses `vi.stubEnv` and `vi.unstubAllEnvs`, preserving any pre-existing `LINEAR_API_KEY` value.

PR #1652, opened 2026-07-29
Keep Stop reviews quiet until a user prompt

TheMostlyGreat:
## What changed

- Suppress a repeated generic Stop review until the next `UserPromptSubmit`.
- Persist and clear a session-scoped quality-state marker while preserving typecheck, phase, and done verification gates.
- Initialize missing session state for replayed/resumed sessions, then clean it up through the established quality-state lifecycle.
- Add installed-hook regressions for the complete Stop → prompt → Stop lifecycle and for a marker clear that must survive optional reminder failure.

## Why

Claude can invoke a Stop hook more than once without setting `stop_hook_active`, which previously produced synthetic idle review turns. Addresses #1492.

## Impact

Idle repeated Stops are silent; a real user prompt re-arms generic review; typecheck advice, phase reviews, and done-phase verification remain enforced.

## Validation

- Affected installed-hook coverage: idle review, prompt hook, typecheck gate, phase backstop, and frozen transcript format.
- Local gates: `bun run lint` (ESLint, Gherkin, TypeScript) and `packages/cli` `tsc --noEmit`.
- Template/dogfood synchronization and schema registration: `scripts/parity-check.ts --mode=all`.
- Ticket evidence: the TDD GREEN ledger and its reachability checks live with YBBGKB.
- Per-commit CI output and current status live on the [Checks tab](https://github.com/ArcadeAI/safeword/pull/1652/checks), not in this static description.

PR #1699, opened 2026-07-31
Ship a clean v0.70 release for safeword users

TheMostlyGreat:
## Summary

- audits the complete release range since `v0.69.0` across 93 pre-existing commits and 71 merged PRs
- fixes unsafe and non-portable tracker identities, hyphenated tracker-key parsing, GitHub CLI keychain authentication, graph-edge parity, generated-plugin drift, and package export ordering
- adds connected-provider process-boundary coverage and completes a bounded semantic refactor pass
- reconciles local tracking and closes completed GitHub issues #772, #773, and #1032 while retaining broader incomplete issues #644, #810, and #1166
- updates patch-level tooling dependencies and current GitHub Actions documentation
- includes all reviewer-authored refactors, current `main`, and resolution of every review thread

## Why

The upcoming release combines a large set of changes across the CLI, hooks, templates, tracker synchronization, Codex plugin, documentation, and release packaging. This PR provides a single evidence-backed quality gate and fixes the release blockers found during that review.

## Root causes addressed

- tracker-provided ticket identities reached path construction without a strict portable-filesystem boundary
- consumers reconstructed connected tracker keys by splitting folder names instead of reading authoritative ticket metadata
- live sync preflight required environment credentials even though its `gh` transport supports stored keychain authentication
- plan and live tracker projections duplicated graph resolution and diverged on self-edges
- generated/package validation was not part of the initial fast verification lane, allowing documentation and export-condition drift

## Review follow-up

- fast-forwarded the three reviewer-authored refactor/documentation commits
- merged current `main` without conflict
- researched the Windows filename note against Microsoft and Git primary sources
- rejected reserved device basenames followed by a period before callbacks or filesystem writes, while retaining valid slug-suffixed lookalikes
- resolved the sole inline review thread with negative and positive regression evidence
- independent refreshed-diff and focused re-reviews: **APPROVE**, no critical issues

## Verification

- Vitest: 376 files passed; 5,663 tests passed, 5 skipped
- BDD: 499 scenarios passed, 3 skipped; 15,444 steps passed, 4 skipped
- release tests: 26/26
- lint, TypeScript, formatting, CLI/package build, website typecheck/build, schema/template parity, dependency-cruiser, Knip, publint, and production dependency audit passed
- all 200 generated pairs and 8 contracts are in sync

Detailed evidence is recorded in `.project/tickets/P2JDY5-release-readiness-v0-70/ticket.md`, `.project/tickets/P2JDY5-release-readiness-v0-70/verify.md`, and `.safeword/logs/ticket-P2JDY5-release-readiness-v0-70.md`.


packages/cli/src/utils/ticket-writer.ts
```diff
@@ -75,6 +75,22 @@ export interface MintedIdentity {
 /** Mints identity from the tracker (the network boundary). Injected for tests. */
 export type IdentitySource = () => Promise<MintedIdentity>;
 
+const SAFE_TRACKER_IDENTITY = /^[A-Z0-9][\w.-]*$/i;
```
TheMostlyGreat, line 78: Reviewer: `SAFE_TRACKER_IDENTITY` (`/^[A-Z0-9][\w.-]*$/i`) correctly blocks every path separator, traversal segment, and control character verified in the parametrized suite below (`ticket-writer-issue-first.test.ts`). One case it does not block: a Windows-reserved device name (`CON`, `PRN`, `AUX`, `NUL`, `COM1`-`COM9`, `LPT1`-`LPT9`) or a trailing-dot value — both are legal per this regex but are known-problematic Windows filenames (silently stripped or reserved by the OS). Not fixed here: it is unverified whether untrusted tracker responses on a Windows host are actually in this project's threat model, so I am flagging it as a NOTE rather than guessing at a fix. If it matters, the narrowest fix is an explicit reserved-name blocklist check alongside this regex.

TheMostlyGreat (commented):
Reviewer: Independent `/quality-review` + `/refactor` pass on this PR's full diff against `main` (44 files — tracker-identity safety, hyphenated-ID metadata, GitHub CLI keychain auth, graph-edge unification, typecheck-gate path containment, release packaging).

**Verdict: APPROVE** — no critical or high-severity issues.

**Pushed 3 follow-up refactor commits on top of this PR** (`0dbdae4a5`, `d96f2c11a`, `a0a8f6fb1`), each independently tested and scoped:

1. **`0dbdae4a5`** — extract `utils/frontmatter.ts`. `feature-source.ts`'s `frontmatterScalar` and `health.ts`'s `coverageTicketScalar` were byte-identical parsers introduced independently within this same PR (both split on `\r?\n`, check the `---` fence, scan for a `field:` prefix). Unified into one shared reader; both call sites now import it.
2. **`d96f2c11a`** — extract `utils/gh-cli.ts`. `github-rest.ts`'s `ghAuthToken`/`isBearerCredentialSyntax` and `sync-tracker.ts`'s new `keychainCredential` both re-implemented "shell out to `gh auth token`, strip the trailing newline, validate RFC 6750 Bearer syntax." Unified into one function. Side effect worth knowing about: `sync-tracker.ts`'s keychain check now also strips `GITHUB_TOKEN` (every casing) from the child env before invoking `gh`, matching its own doc comment that it reads the *keychain* credential rather than an inherited env var — previously it would silently reflect back whatever `GITHUB_TOKEN` happened to be set to. This doesn't change `syncTracker`'s actual AC2 exit code (the same env var is checked again immediately after via `resolveCredential`'s fallback) — it's a fidelity fix, not a behavior change.
3. **`a0a8f6fb1`** — doc-only. Restored a comment on `createdThisRun` in `github-rest.ts` that the "isolate retro issue snapshot state" extraction dropped. It explains *why* the same-run-created-issues list exists: `prepareEncounters` (pipeline.ts) does not dedupe by signature, so two same-signature findings in one triage batch would double-file without it. That invariant isn't documented anywhere else in the codebase.

All three ran through their existing targeted suites (150 tests / 8 files) plus a full `tsc --noEmit` and a diff-scoped `/audit` (dependency-cruiser clean, config in sync, no dead code left behind) before being pushed.

**One item found but deliberately not actioned** — flagged inline on `ticket-writer.ts`: `SAFE_TRACKER_IDENTITY` would accept a Windows-reserved filename or a trailing-dot value as "safe." Not fixed since I couldn't verify whether untrusted tracker responses on a Windows host are actually in this project's threat model — your call whether it's worth a follow-up ticket.

**Everything else confirmed correct, no action needed:**
- The tracker-identity path-injection fix has solid parametrized regression coverage (8 malicious-input cases: `../`, `..\`, `.`, `..`, absolute POSIX/Windows paths, NUL byte, control char) exercised against the real `createIssueFirstTicket`.
- `loadTrackerMapOrEmpty`'s sidecar-corrupt-vs-missing unification incidentally fixes a real latent bug: `ticket-create/index.ts`'s `writeReference` used to silently discard a *corrupt* sidecar and overwrite it with a fresh one (losing every other ticket's recorded ref) — it now throws instead.
- `typecheck-gate.ts`'s `findUpWithin` unification fixes a real sibling-directory string-prefix bug (`/repo-sibling` used to match root `/repo` via `.startsWith`) — has a dedicated regression test.
- The new keychain wiring test (`wiring.test.ts`) exercises the real `syncTrackerCommand` against a fake `gh` executable on `PATH` — only the process boundary is mocked, so it's real coverage, not a mock-only check.
- The three additional smells found during your own release review (lint-hook orchestration, github-rest test-suite size, workspace-detection duplication) were correctly deferred to their own tickets (`5EX2NF`, `84GNG0`, `KRSFP6`) instead of widening this diff — good call.
- Dogfood/template/Codex mirror byte-parity holds for every hook/script/skill this PR touches.
- Dependency bumps are trivial devDependency patches (`@types/node`, `markdownlint-cli2` and transitives) — no production dependency changes.

packages/cli/src/utils/ticket-writer.ts
TheMostlyGreat: Resolved in 760afdb76. Microsoft confirms the reserved-device rule applies when the device basename is immediately followed by an extension. Because safeword always composes `${id}-${slug}`, bare `CON` and a raw ID ending in `.` become portable (`CON-login-bug`, `ENG-45.-login-bug`); `NUL.json-login-bug` remains reserved. The validator now rejects exactly the reserved-basename-plus-period family before `onMinted` or filesystem writes. Regression coverage includes `nul.json`, `CON.ticket`, and `lpt9.backup`, plus positive cases for `CON`, `COM1`, and `ENG-45.`. Focused tests pass 16/16; the refreshed full suite passes 5,663 tests and 499 BDD scenarios.

TheMostlyGreat (commented):
Reviewer: Follow-up on the updated PR (merged `main` + the Windows-reserved-filename fix).

Verified the fix against Microsoft's own [Naming Files, Paths, and Namespaces](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file) doc rather than taking the commit's reasoning at face value:

- Confirms the exact rule `WINDOWS_RESERVED_DEVICE_WITH_EXTENSION` codifies: a reserved name "followed immediately by an extension" (e.g. `NUL.txt`) is equivalent to the bare reserved name. The `[1-9]` bound on `COM`/`LPT` is also correct (no `COM0`/`LPT0`).
- The doc additionally reserves Unicode superscript-digit variants (`COM¹`, `LPT²`, `LPT³`). Checked whether that's a gap here — it isn't: `SAFE_TRACKER_IDENTITY` already rejects any identity containing those characters at the first gate (`\w` is ASCII-only), confirmed empirically with `node -e`. The fix is airtight as shipped.
- The 3 negative + 3 positive parametrized cases land exactly on the real boundary (bare device name → safe once slug-suffixed; device name + dot → unsafe regardless of suffix).

Ran the combined suite after fast-forwarding to your latest (241 tests / 11 files, spanning `ticket-writer`, the merged-in architecture changes, and my earlier 3 refactor commits) plus a full `tsc --noEmit` — all green. My earlier 3 commits survived the `main` merge with no duplication reintroduced.

Nothing further to refactor in this delta — the fix is minimal and self-contained. This closes out my only open thread; **APPROVE** (recorded as COMMENT — self-review blocks the actual approve state).

TheMostlyGreat (commented):
Reviewer: Full `/quality-review` + `/refactor` pass covering everything since my last full review (`a0a8f6fb1`) — 3 commits (`760afdb76`, `015f2fd7e`, `d1b10c121`) plus the `main` merge.

**Verdict: APPROVE** — no critical or high-severity issues.

### `760afdb76` — Windows-reserved-name fix (already confirmed in my earlier follow-up, restated here for the record)
Verified against Microsoft's own [Naming Files, Paths, and Namespaces](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file) doc: the "reserved name followed immediately by an extension" rule and the `COM1`-`COM9`/`LPT1`-`LPT9` bound (no `COM0`/`LPT0`) both match exactly. Checked the one thing the doc adds that the fix doesn't explicitly handle — Unicode superscript-digit device names (`COM¹`, `LPT²`) — and confirmed empirically that `SAFE_TRACKER_IDENTITY` already rejects those at the first gate (`\w` is ASCII-only), so there's no residual gap. Test coverage (3 negative + 3 positive parametrized cases) lands exactly on the real boundary.

### `d1b10c121` — new since my last pass, reviewed fresh
Two independent changes bundled here:

1. **Keychain-only sync now has real end-to-end wiring proof.** The previous wiring test only proved the *preflight* accepted a keychain credential; this extends the same fake-`gh`-on-`PATH` test through an actual `issue create` call, verifies the sidecar (`tracker-map.json`) records the right `{provider, id, url}`, asserts a clean exit code, and asserts empty stderr. This is a real gap-close, not a duplicate of the existing test — the old assertion (`callLog` trimmed to exactly `'auth token'`) is gone because the call log now legitimately contains multiple `gh` invocations (`auth token`, `repo view`, `issue create`), and the new assertions check for each by content rather than exact-string-matching the whole log. Good test; only the process boundary (`gh`) is mocked.
2. **Pinned `actions/checkout` and `actions/setup-node` in the copyable docs workflow to immutable commit SHAs**, replacing floating `@v7` tags. I verified both SHAs directly against GitHub's API rather than trusting the commit message:
   - `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1` — confirmed via `gh api repos/actions/checkout/git/refs/tags/v7.0.1`, exact match. `v7.0.1` is also the current latest release (published 2026-07-20).
   - `actions/setup-node@820762786026740c76f36085b0efc47a31fe5020` — confirmed via `gh api repos/actions/setup-node/git/refs/tags/v7.0.0`, exact match. `v7.0.0` is also the current latest release (published 2026-07-14).
   - Both correct and current — no drift, no typo, no wrong-commit risk. This is a legitimate supply-chain hardening (SHA-pinning defends against a retagged/mutated release).
3. Also cleared a trailing-whitespace diff-hygiene defect in `verify.md` — confirmed with `git diff --check` across the PR's full current diff against `main`: clean.

### Main-merge (`df31c884f` "Reuse monorepo discovery during architecture healing", `ee8473d56` docs) — out of scope, noted for transparency
These are already-shipped, independently-reviewed commits on `main`; they show up in the branch history from the merge but contribute nothing to this PR's actual diff against current `main` (verified via the three-dot diff). I did check one thing: whether `df31c884f` incidentally resolved the deferred follow-up ticket `KRSFP6` (duplicated workspace-discovery policy between `architecture-skeleton.ts` and `architecture-monorepo.ts`) — it doesn't. `architecture-skeleton.ts`'s `declaresWorkspaces` is still a separate, narrower reimplementation (it has no Cargo-workspace check, unlike `architecture-monorepo.ts`'s full `discoverWorkspaces`). `KRSFP6` remains open and correctly scoped as its own follow-up — not a regression, not something this PR needs to absorb.

### Refactor pass: nothing new to do
Checked for new duplication introduced since my last pass — none. I did look at whether the "write a fake `gh`/tool script, `chmod` it, prepend `PATH`" pattern in the new wiring test is now duplicated enough to extract a shared test helper: it appears in 12 test files repo-wide, but only 2 of them (`wiring.test.ts`, `ticket-new-wiring.test.ts`) belong to this PR, and the pattern predates this PR across the other 10 — each script's actual behavior differs per test, so this reads as idiomatic per-test setup, not a clean duplicate. Extracting a shared helper would be a repository-wide test-infrastructure change unrelated to this PR's scope; not proposing it here.

### Verification run against the current head (`d1b10c121`)
- Combined targeted suite: **259 tests / 14 files, all green** (wiring, ticket-writer, feature-source, health/phase-anchor, github-rest, sync-tracker, architecture-monorepo, architecture-project, ticket-new-wiring, tracker-map, plan-gh-parity).
- Full `tsc --noEmit`: clean.
- Diff-scoped `/audit`: config in sync, dependency-cruiser clean (272 modules, 386 dependencies, 0 violations), `git diff --check` clean.
- My earlier 3 refactor commits (`frontmatter.ts`, `gh-cli.ts` extractions, restored dedup-rationale comment) remain intact through both the `main` merge and these new commits.

Nothing outstanding on my end.

TheMostlyGreat (commented):
Reviewer: Full `/quality-review` + `/refactor` pass covering everything since my last full review (`d1b10c121`) — 3 new commits on the PR's own branch: `35aa34a13`, `4a52599fa`, `b92f1a846` (plus another `main` merge, confirmed out of scope — see below).

**Verdict: APPROVE** — no critical, high, or medium issues.

### `35aa34a13` — reconciles a stale ticket index, sharpens credential-resolution naming
Two independent fixes bundled here:

1. **`.project/tickets/INDEX.md` was stale** after the previous `main` merge (missing entries, some stale statuses). Regenerated from canonical ticket files. I verified the claimed count directly rather than trusting the commit message: `find .project/tickets -maxdepth 1 -mindepth 1 -type d | wc -l` → **467**, matching `## Tickets (467)` in the regenerated index exactly.
2. **Renamed `keychainCredential` → `ghCliCredential`** in `sync-tracker.ts`, and tightened the doc comments on it and on `utils/gh-cli.ts`'s `resolveGhCliToken` (which I extracted in an earlier pass). This is a direct, correct follow-up to my own first review: the old name/comment claimed this reads "the OS credential store," which overstated it — `gh auth token` can resolve from `GH_TOKEN` *or* the stored credential, and the function never distinguished which. The new name and comment ("what `gh` has on its own... `GH_TOKEN` or the OS credential store") are accurate to what the code actually does. Pure rename + doc fix, no behavior change — confirmed by grep (`ghCliCredential` has exactly one definition and one call site, both updated consistently).
3. `packages/website/.../tracker-integration.mdx` updated to match: now says safeword's preflight "probes `GH_TOKEN` or stored authentication before falling back to `GITHUB_TOKEN`," and separately documents `gh`'s own precedence for live writes. I checked this against the actual code path (`resolveCredential` in `secrets.ts`: keychain-arg first, then raw `GITHUB_TOKEN` env fallback) — accurate.

### `4a52599fa`, `b92f1a846` — bookkeeping only
Test-count refresh in `verify.md`/ticket log after merging `main` again (5,663→5,670 tests, 376→377 files — consistent with `main` having brought in new Stop-hook test files), and a single "PR Scope: ✅" line added to the already-merged `YBBGKB` ticket's `verify.md`. Low-risk, accurate, no findings.

### Another `main` merge — confirmed out of scope again
`main` advanced further (PR #1652/#1723, Stop-hook quiet-idle-review work) and the branch merged it. Checked the three-dot diff against current `main` before and after: the Stop-hook/`quality-state.ts`/`ARCHITECTURE.md` changes contribute **zero** lines to this PR's actual diff (they cancel out — already on both sides). Confirmed with `git diff --stat origin/main...HEAD -- <those paths>` → empty. Not reviewing someone else's already-merged, already-reviewed work as if it were this PR's.

### Refactor pass: nothing new
The only code touched is the `ghCliCredential` rename, which is itself the clean-up. No new duplication introduced.

### Verification against current head (`b92f1a846`)
- Targeted suite: **184 tests / 12 files, all green** (wiring, ticket-writer, feature-source, phase-anchor, check, github-rest, sync-tracker, ticket-new-wiring, tracker-map, plan-gh-parity).
- Full `tsc --noEmit`: clean.
- `/audit`: config in sync, dependency-cruiser clean (274 modules, 390 dependencies, 0 violations), `git diff --check` against `main` clean.
- Dogfood/template byte-parity re-verified across every touched hook/script/skill pair: `typecheck-gate.ts`, `phase-provenance.ts`, `cleanup-zombies.sh`, `ticket-system/SKILL.md` — all match.

Nothing outstanding on my end.

PR #1700, opened 2026-07-31
Keep BDD scenario evidence honest

TheMostlyGreat:
## What changed

- Require BDD primary proof to exercise the actor-facing action and observe the actor-visible result.
- Add RED and REFACTOR review checks plus regression coverage for that boundary.
- Make queued package-test runs report lock owner PID, checkout root, and elapsed wait periodically.

## Why

A lower-level store call or internal-state assertion can pass without proving the user-visible scenario it claims to cover. Separately, a legitimate cross-worktree package-test lock previously looked like an idle test hang after one silent notice.

## Impact

Maintainers get more trustworthy BDD completion evidence and clear progress during serialized test waits. Existing lock serialization, stale-lock handling, and the 20-minute wait cap are unchanged.

## Validation

- `bun run test tests/hooks/scenario-proof-fidelity-documentation.test.ts`
- `bun run test tests/test-runner-lock.test.ts` (8/8)
- `bun run test` (5,647 passed, 5 skipped across 377 files)
- `bun run lint`
- Prettier and `git diff --check`

### DIFF
--- a/.project/tickets/1698-keep-bdd-evidence-honest/verify.md
+++ b/.project/tickets/1698-keep-bdd-evidence-honest/verify.md
@@ -16,8 +16,13 @@ tests passed 44/44.
 **Template Parity:** ✅ Canonical BDD/TDD and TDD-review templates exactly match
 their installed `.claude/skills/` dogfood copies.
 
+**Codex Plugin Release Gate:** ✅ `bun run --cwd packages/cli test:release`
+passed 26/26 tests across 8 files after `bun run --cwd packages/cli
+generate:codex-plugin` regenerated the BDD and TDD-review skill assets.
+
 **PR Scope:** ✅ Changes are limited to issue #1698's planning artifacts,
-scenario-proof guidance, installed dogfood copies, and the regression test.
+scenario-proof guidance, installed and generated agent-surface copies, and
+regression tests.
 
 **Lock Investigation:** ✅ The earlier apparent idle run was waiting for a
 legitimate test owner in another worktree, not hanging inside Vitest. The
--- a/.project/tickets/1698-keep-bdd-evidence-honest/verify.md
+++ b/.project/tickets/1698-keep-bdd-evidence-honest/verify.md
@@ -1,0 +1,24 @@
+# Verification: Keep BDD evidence honest
+
+**Test Suite:** ✅ `bun run test:done` passed 1,244/1,244 tests across 81 files.
+
+**Full Suite:** ✅ `bun run test` passed 5,645 tests with 5 skips across 377
+files. Vitest completed in 379.17 seconds after the global package-test lock
+became available.
+
+**Focused Proof:** ✅ The scenario-fidelity and adjacent BDD documentation
+tests passed 44/44.
+
+**Lint:** ✅ ESLint and Gherkin validation passed.
+
+**Typecheck:** ✅ `tsc --noEmit` passed.
+
+**Template Parity:** ✅ Canonical BDD/TDD and TDD-review templates exactly match
+their installed `.claude/skills/` dogfood copies.
+
+**PR Scope:** ✅ Changes are limited to issue #1698's planning artifacts,
+scenario-proof guidance, installed dogfood copies, and the regression test.
+
+**Lock Investigation:** ✅ The earlier apparent idle run was waiting for a
+legitimate test owner in another worktree, not hanging inside Vitest. The
+global wrapper allows up to 20 minutes of lock waiting.
--- a/.project/tickets/1698-keep-bdd-evidence-honest/verify.md
+++ b/.project/tickets/1698-keep-bdd-evidence-honest/verify.md
@@ -2,9 +2,10 @@
 
 **Test Suite:** ✅ `bun run test:done` passed 1,244/1,244 tests across 81 files.
 
-**Full Suite:** ✅ `bun run test` passed 5,645 tests with 5 skips across 377
-files. Vitest completed in 379.17 seconds after the global package-test lock
-became available.
+**Full Suite:** ✅ At the issue #1698 snapshot, `bun run test` passed 5,645
+tests with 5 skips across 377 files. Vitest completed in 379.17 seconds after
+the global package-test lock became available. The later CCYD5S snapshot records
+its then-current 5,647-test suite after additional lock-runner coverage.
 
 **Focused Proof:** ✅ The scenario-fidelity and adjacent BDD documentation
 tests passed 44/44.
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -57,7 +57,7 @@ When lock creation is extracted from the wait loop
 Then successful acquisition, serialization, stale-lock recovery, and wait-cap
 behavior remain unchanged
 
-- [x] REFACTOR skip: focused lock-runner suite passed 9/9 after the extract
+- [x] REFACTOR f4195b63e
 
 ## Patch-level refactor
 
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -39,9 +39,9 @@ When the runner waits for its configured wait cap
 Then it reports status no more often than the supported minimum interval
 And malformed interval settings fall back to the default interval
 
-- [x] RED 09ac816ab
+- [x] RED skip: unsafe-interval test failed before the minimum clamp
 - [x] GREEN 09ac816ab
-- [x] REFACTOR 09ac816ab
+- [x] REFACTOR skip: shared helpers are the completed cleanup
 
 ## Patch-level refactor
 
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -21,10 +21,12 @@ And it emits increasing elapsed-wait status more than once
 
 ### Scenario: CCYD5S.SM1.AC3.tolerates_incomplete_owner_metadata
 
-Given a live lock has readable but incomplete or non-object owner metadata
+Given a live lock has readable but incomplete owner metadata, or stale non-object
+owner metadata
 When another runner reports its wait
 Then it prints the fields that are available, or marks unavailable fields clearly
 And it continues waiting or reaches the configured wait cap without crashing
+And stale non-object metadata is reaped through the normal mtime fallback
 
 - [x] RED skip: focused runner suite failed because incomplete metadata emitted no details
 - [x] GREEN fa16069cb
@@ -38,6 +40,13 @@ Given a live package-test lock and a one-millisecond status interval setting
 When the runner waits for its configured wait cap
 Then it reports status no more often than the supported minimum interval
 And malformed and zero interval settings fall back to the default interval
+
+### Scenario: CCYD5S.SM1.AC6.negative_maximum_wait_uses_the_default
+
+Given another runner owns a live package-test lock
+When a waiting runner sets its maximum wait to a negative value
+Then it waits for the owner under the default maximum wait
+And it does not proceed without the lock or overlap the owner's build and test
 
 - [x] RED skip: unsafe-interval test failed before the minimum clamp
 - [x] GREEN 09ac816ab
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -15,9 +15,9 @@ Then the queued runner reports the owner's process ID
 And it reports the owner's checkout root
 And it emits increasing elapsed-wait status more than once
 
-- [x] RED: focused runner suite failed because no periodic owner status existed
-- [x] GREEN: owner checkout and increasing elapsed statuses passed
-- [x] REFACTOR: shared status formatting and scheduling remained local to the runner
+- [x] RED skip: focused runner suite failed because no periodic owner status existed
+- [x] GREEN fa16069cb # owner checkout and increasing elapsed statuses passed
+- [x] REFACTOR skip: shared status formatting and scheduling remained local to the runner
 
 ### Scenario: CCYD5S.SM1.AC3.tolerates_incomplete_owner_metadata
 
@@ -26,10 +26,10 @@ When another runner reports its wait
 Then it prints the fields that are available
 And it continues waiting or reaches the configured wait cap without crashing
 
-- [x] RED: focused runner suite failed because incomplete metadata emitted no details
-- [x] GREEN: available owner fields and fallback labels passed
-- [x] REFACTOR: no additional structure was needed
+- [x] RED skip: focused runner suite failed because incomplete metadata emitted no details
+- [x] GREEN fa16069cb # available owner fields and fallback labels passed
+- [x] REFACTOR skip: no additional structure was needed
 
 ## Patch-level refactor
 
-- [x] cross-scenario: both scenarios use the same metadata reader and wait reporter
+- [x] cross-scenario skip: both scenarios use the same metadata reader and wait reporter
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -32,8 +32,14 @@ timestamp, is reaped through the normal recovery paths
 - [x] RED skip: focused runner suite failed because incomplete metadata emitted no details
 - [x] GREEN fa16069cb
 - [x] REFACTOR skip: no additional structure was needed
-- [ ] RED: reproduce stale recovery for a syntactically valid owner object
-  with no usable process ID or timestamp, including an epoch timestamp.
+- [x] RED 8a5d2d35e
+  Reproduced stale recovery for a syntactically valid owner object with no
+  usable process ID or timestamp, including an epoch timestamp.
+- [x] GREEN 8e5ebbf9e
+  Focused runner suite passed 11/11 with mtime recovery for `{}` and age
+  recovery for the Unix epoch.
+- [x] REFACTOR skip: retained the existing recovery paths and named their
+  shared metadata validity checks.
 
 ## Rule: Wait diagnostics remain safely rate-limited
 
@@ -64,7 +70,11 @@ When a waiting runner sets its maximum wait to a negative or blank value
 Then it waits for the owner under the default maximum wait
 And it does not proceed without the lock or overlap the owner's build and test
 
-- [ ] RED: reproduce blank maximum-wait fallback while a real owner holds the lock.
+- [x] RED 8a5d2d35e
+  Reproduced blank maximum-wait fallback while a real owner holds the lock.
+- [x] GREEN 8e5ebbf9e
+  Focused runner suite passed 11/11 with blank and negative fallback.
+- [x] REFACTOR skip: reused the existing safe-integer environment resolver.
 
 ## Refactoring ledger
 
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -43,6 +43,22 @@ And malformed interval settings fall back to the default interval
 - [x] GREEN 09ac816ab
 - [x] REFACTOR skip: shared helpers are the completed cleanup
 
+## Refactoring ledger
+
+- [x] CCYD5S.RF1 skip: Extracted lock creation from `acquireLock()` in
+  `packages/cli/scripts/run-vitest-with-build-lock.mjs`; existing integration
+  scenarios cover successful creation, contention, stale-lock recovery, and
+  wait-cap behavior.
+
+### Scenario: CCYD5S.SM1.AC5.preserves_lock_behavior_after_extracting_creation
+
+Given the existing lock-runner integration scenarios
+When lock creation is extracted from the wait loop
+Then successful acquisition, serialization, stale-lock recovery, and wait-cap
+behavior remain unchanged
+
+- [x] REFACTOR skip: focused lock-runner suite passed 9/9 after the extract
+
 ## Patch-level refactor
 
 - [x] cross-scenario skip: both scenarios use the same metadata reader and wait reporter
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -21,16 +21,19 @@ And it emits increasing elapsed-wait status more than once
 
 ### Scenario: CCYD5S.SM1.AC3.tolerates_incomplete_owner_metadata
 
-Given a live lock has readable but incomplete owner metadata, or stale non-object
-owner metadata
+Given a live lock has readable but incomplete owner metadata, or stale metadata
+that is non-object or lacks both a usable process ID and timestamp
 When another runner reports its wait
 Then it prints the fields that are available, or marks unavailable fields clearly
 And it continues waiting or reaches the configured wait cap without crashing
-And stale non-object metadata is reaped through the normal mtime fallback
+And stale metadata without a usable process ID or timestamp, or with an expired
+timestamp, is reaped through the normal recovery paths
 
 - [x] RED skip: focused runner suite failed because incomplete metadata emitted no details
 - [x] GREEN fa16069cb
 - [x] REFACTOR skip: no additional structure was needed
+- [ ] RED: reproduce stale recovery for a syntactically valid owner object
+  with no usable process ID or timestamp, including an epoch timestamp.
 
 ## Rule: Wait diagnostics remain safely rate-limited
 
@@ -41,23 +44,9 @@ When the runner waits for its configured wait cap
 Then it reports status no more often than the supported minimum interval
 And malformed and zero interval settings fall back to the default interval
 
-### Scenario: CCYD5S.SM1.AC6.negative_maximum_wait_uses_the_default
-
-Given another runner owns a live package-test lock
-When a waiting runner sets its maximum wait to a negative value
-Then it waits for the owner under the default maximum wait
-And it does not proceed without the lock or overlap the owner's build and test
-
 - [x] RED skip: unsafe-interval test failed before the minimum clamp
 - [x] GREEN 09ac816ab
 - [x] REFACTOR skip: shared helpers are the completed cleanup
-
-## Refactoring ledger
-
-- [x] CCYD5S.RF1 skip: Extracted lock creation from `acquireLock()` in
-  `packages/cli/scripts/run-vitest-with-build-lock.mjs`; existing integration
-  scenarios cover successful creation, contention, stale-lock recovery, and
-  wait-cap behavior.
 
 ### Scenario: CCYD5S.SM1.AC5.preserves_lock_behavior_after_extracting_creation
 
@@ -67,6 +56,22 @@ Then successful acquisition, serialization, stale-lock recovery, and wait-cap
 behavior remain unchanged
 
 - [x] REFACTOR f4195b63e
+
+### Scenario: CCYD5S.SM1.AC6.negative_maximum_wait_uses_the_default
+
+Given another runner owns a live package-test lock
+When a waiting runner sets its maximum wait to a negative or blank value
+Then it waits for the owner under the default maximum wait
+And it does not proceed without the lock or overlap the owner's build and test
+
+- [ ] RED: reproduce blank maximum-wait fallback while a real owner holds the lock.
+
+## Refactoring ledger
+
+- [x] CCYD5S.RF1 skip: Extracted lock creation from `acquireLock()` in
+  `packages/cli/scripts/run-vitest-with-build-lock.mjs`; existing integration
+  scenarios cover successful creation, contention, stale-lock recovery, and
+  wait-cap behavior.
 
 ## Patch-level refactor
 
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -16,7 +16,7 @@ And it reports the owner's checkout root
 And it emits increasing elapsed-wait status more than once
 
 - [x] RED skip: focused runner suite failed because no periodic owner status existed
-- [x] GREEN fa16069cb # owner checkout and increasing elapsed statuses passed
+- [x] GREEN fa16069cb
 - [x] REFACTOR skip: shared status formatting and scheduling remained local to the runner
 
 ### Scenario: CCYD5S.SM1.AC3.tolerates_incomplete_owner_metadata
@@ -27,7 +27,7 @@ Then it prints the fields that are available
 And it continues waiting or reaches the configured wait cap without crashing
 
 - [x] RED skip: focused runner suite failed because incomplete metadata emitted no details
-- [x] GREEN fa16069cb # available owner fields and fallback labels passed
+- [x] GREEN fa16069cb
 - [x] REFACTOR skip: no additional structure was needed
 
 ## Patch-level refactor
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -1,0 +1,35 @@
+# Test Definitions: Explain test-lock waits
+
+The executable proof is the existing Vitest integration coverage around
+`run-vitest-with-build-lock.mjs`. Tests use isolated temporary locks and fake
+build/test binaries; they do not contend for the repository's real global lock.
+
+## Rule: A queued package-test command explains its wait
+
+### Scenario: CCYD5S.SM1.AC1-2.reports_owner_and_elapsed_wait_periodically
+
+Given one runner owns a package-test lock from a known checkout
+And a second runner is queued behind it
+When the wait lasts across multiple status intervals
+Then the queued runner reports the owner's process ID
+And it reports the owner's checkout root
+And it emits increasing elapsed-wait status more than once
+
+- [x] RED: focused runner suite failed because no periodic owner status existed
+- [x] GREEN: owner checkout and increasing elapsed statuses passed
+- [x] REFACTOR: shared status formatting and scheduling remained local to the runner
+
+### Scenario: CCYD5S.SM1.AC3.tolerates_incomplete_owner_metadata
+
+Given a live lock has readable but incomplete owner metadata
+When another runner reports its wait
+Then it prints the fields that are available
+And it continues waiting or reaches the configured wait cap without crashing
+
+- [x] RED: focused runner suite failed because incomplete metadata emitted no details
+- [x] GREEN: available owner fields and fallback labels passed
+- [x] REFACTOR: no additional structure was needed
+
+## Patch-level refactor
+
+- [x] cross-scenario: both scenarios use the same metadata reader and wait reporter
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -30,6 +30,19 @@ And it continues waiting or reaches the configured wait cap without crashing
 - [x] GREEN fa16069cb
 - [x] REFACTOR skip: no additional structure was needed
 
+## Rule: Wait diagnostics remain safely rate-limited
+
+### Scenario: CCYD5S.SM1.AC4.clamps_an_unsafe_status_interval
+
+Given a live package-test lock and a one-millisecond status interval setting
+When the runner waits for its configured wait cap
+Then it reports status no more often than the supported minimum interval
+And malformed interval settings fall back to the default interval
+
+- [x] RED: the runner honored a one-millisecond interval and had no malformed-setting coverage
+- [x] GREEN: shared configuration parsing clamps unsafe values and tests pass
+- [x] REFACTOR: one parser owns integer environment-variable validation
+
 ## Patch-level refactor
 
 - [x] cross-scenario skip: both scenarios use the same metadata reader and wait reporter
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -39,9 +39,9 @@ When the runner waits for its configured wait cap
 Then it reports status no more often than the supported minimum interval
 And malformed interval settings fall back to the default interval
 
-- [x] RED: the runner honored a one-millisecond interval and had no malformed-setting coverage
-- [x] GREEN: shared configuration parsing clamps unsafe values and tests pass
-- [x] REFACTOR: one parser owns integer environment-variable validation
+- [x] RED 09ac816ab
+- [x] GREEN 09ac816ab
+- [x] REFACTOR 09ac816ab
 
 ## Patch-level refactor
 
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
@@ -21,9 +21,9 @@ And it emits increasing elapsed-wait status more than once
 
 ### Scenario: CCYD5S.SM1.AC3.tolerates_incomplete_owner_metadata
 
-Given a live lock has readable but incomplete owner metadata
+Given a live lock has readable but incomplete or non-object owner metadata
 When another runner reports its wait
-Then it prints the fields that are available
+Then it prints the fields that are available, or marks unavailable fields clearly
 And it continues waiting or reaches the configured wait cap without crashing
 
 - [x] RED skip: focused runner suite failed because incomplete metadata emitted no details
@@ -37,7 +37,7 @@ And it continues waiting or reaches the configured wait cap without crashing
 Given a live package-test lock and a one-millisecond status interval setting
 When the runner waits for its configured wait cap
 Then it reports status no more often than the supported minimum interval
-And malformed interval settings fall back to the default interval
+And malformed and zero interval settings fall back to the default interval
 
 - [x] RED skip: unsafe-interval test failed before the minimum clamp
 - [x] GREEN 09ac816ab
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/user-stories.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/user-stories.md
@@ -35,3 +35,10 @@ lock serialization.
 Given a maintainer supplies an invalid or excessively small diagnostic interval
 When a command waits for the package-test lock
 Then the runner uses a safe minimum interval and keeps wait output readable.
+
+#### CCYD5S.SM1.AC5 - Lock coordination stays readable
+
+Given package-test locking needs both lock creation and wait coordination
+When a maintainer reads or changes the runner
+Then lock creation is named separately from the wait loop without changing
+serialization, stale-lock recovery, or diagnostic behavior.
--- a/.project/tickets/CCYD5S-explain-test-lock-waits/user-stories.md
+++ b/.project/tickets/CCYD5S-explain-test-lock-waits/user-stories.md
@@ -29,6 +29,8 @@ Given the lock metadata is incomplete, malformed, or changes while my command re
 When the wait status is rendered
 Then the command reports the available details without crashing or weakening
 lock serialization.
+And malformed stale metadata remains eligible for the existing stale-lock
+recovery path.
 
 #### CCYD5S.SM1.AC4 - Diagnostic configuration stays safe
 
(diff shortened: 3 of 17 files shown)


packages/cli/templates/skills/bdd/TDD.md
```diff
@@ -24,6 +24,32 @@ Degradation is the intended path — no gate blocks on harness absence.
 
 Start with the most constraining test — usually E2E or integration. Prefer the highest scope that covers the behavior with acceptable feedback speed.
 
+### Scenario proof fidelity
```
TheMostlyGreat, line 27: Reviewer: CRITICAL — this new "Scenario proof fidelity" section (and the matching addition in `tdd-review/SKILL.md`) isn't reflected in `packages/cli/codex-plugin/skills/bdd/references/TDD.md`, the derived Codex-plugin copy generated from this file via `bun run generate:codex-plugin`. `packages/cli/tests/codex-plugin-catalogue.release.test.ts` byte-compares `codex-plugin/skills/**` against the canonical transformation of `templates/skills/**`, so `bun run test:release` (CI's "Release-gate tests" step, `.github/workflows/ci.yml:120`) will fail on this exact diff. Fix: run `bun run generate:codex-plugin` from `packages/cli` and commit the regenerated files under `codex-plugin/skills/**`. (verified: diffed both file pairs directly — the new content is present here and absent from the codex-plugin copy; traced the release-gate test and CI step that would catch it)

packages/cli/templates/skills/tdd-review/SKILL.md
```diff
@@ -33,6 +33,11 @@ Focused review (~1 minute). Check the test that was just written:
 - **Atomic?** Tests ONE behavior. Red flag: multiple When/Then pairs.
 - **Right assertions?** Meaningful expectations, not `.toBeTruthy()` or `.not.toThrow()`.
 - **Behavior, not implementation?** Tests observable outcomes. Red flag: mocking internals, checking call counts.
+- **Scenario fidelity?** For a BDD scenario, does the primary test exercise the
```
TheMostlyGreat, line 36: Reviewer: CRITICAL — same generator-drift issue as `TDD.md`: `packages/cli/codex-plugin/skills/tdd-review/SKILL.md` is missing this "Scenario fidelity?" bullet (and the second one added near line 65 below). Same fix — see the comment on `templates/skills/bdd/TDD.md:27` for the full explanation.

.project/tickets/1698-keep-bdd-evidence-honest/verify.md
```diff
@@ -0,0 +1,24 @@
...
+became available.
+
+**Focused Proof:** ✅ The scenario-fidelity and adjacent BDD documentation
+tests passed 44/44.
+
+**Lint:** ✅ ESLint and Gherkin validation passed.
+
+**Typecheck:** ✅ `tsc --noEmit` passed.
+
+**Template Parity:** ✅ Canonical BDD/TDD and TDD-review templates exactly match
```
TheMostlyGreat, line 17: Reviewer: WARN — this parity claim is true for `.claude/skills/` vs `packages/cli/templates/skills/`, but doesn't cover the derived `packages/cli/codex-plugin/skills/**` copies, which are stale after this PR (see the codex-plugin comment on `templates/skills/bdd/TDD.md`). `bun run test:release` — the step that would actually catch this — isn't part of `bun run test:done` or `bun run test`, both reported green above, so this verify.md doesn't reflect the real gap. Worth adding a `test:release` line to this checklist going forward.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -55,6 +58,20 @@ function resolveMaximumLockWaitMilliseconds() {
 
 const maximumLockWaitMilliseconds = resolveMaximumLockWaitMilliseconds();
 
+function resolveLockStatusIntervalMilliseconds() {
```
TheMostlyGreat: Reviewer: WARN (refactor) — `resolveLockStatusIntervalMilliseconds` duplicates `resolveMaximumLockWaitMilliseconds` (line 49) almost exactly: read an env var, return the default if unset, `Number()`-parse, validate with `Number.isSafeInteger` plus a sign check, fall back to default if invalid. Only the env-var name, default constant, and `> 0` vs `>= 0` comparator differ. A shared `resolvePositiveIntegerEnv(name, defaultValue, { allowZero })` would replace both call sites and head off a third near-copy the next time a tunable is added. Low risk, easy win.

TheMostlyGreat (commented):
Combined `/quality-review` + `/refactor` pass over this PR's full diff (two independent subagent reviews, findings cross-checked against the actual code/CI config before posting). **1 blocker, 3 warns, 5 notes** — inline below.

Since `gh` here authenticates as this PR's own author, GitHub blocks APPROVE/REQUEST_CHANGES on your own PR, so this posts as `COMMENT` and carries no merge block — treat the blocker as a real one anyway.

**Verified sound, not just asserted:**
- Lock-wait reporting logic (`run-vitest-with-build-lock.mjs`): no timer/interval leaks (synchronous polling, not `setInterval`), strictly increasing/non-duplicated elapsed reports, no spam under default config. The 20-minute wait cap, stale-lock reaping, and mkdir-based lock exclusion are genuinely untouched, matching the PR description's "unchanged" claim.
- Malformed/missing owner metadata is handled gracefully everywhere it's read (try/catch + `typeof` guards), consistent with the pre-existing `removeStaleLock` pattern.
- Wiring gate: the two new `test-runner-lock.test.ts` cases spawn real Node subprocesses running the actual unmocked script and race two genuine processes for the lock — real integration coverage, not mocked.
- Mirror parity between `.claude/skills/**` and `packages/cli/templates/skills/**` (the pair this PR touches) is byte-identical — confirmed via direct diff.
- No deprecated Node APIs; `import.meta.dirname`, `Atomics.wait`+`SharedArrayBuffer`, `Array.prototype.toSorted` are all current and already used elsewhere in this codebase.
- Diff scope matches the PR description's three stated changes with no creep; both tickets' scenario docs map cleanly to the real code/test changes.

**The blocker:** `packages/cli/codex-plugin/skills/bdd/references/TDD.md` and `packages/cli/codex-plugin/skills/tdd-review/SKILL.md` are derived, checked-in copies generated from `packages/cli/templates/skills/**` via `bun run generate:codex-plugin`. This PR added content to the templates but never regenerated the Codex-plugin copies, so they're now missing the new "Scenario proof fidelity" section and "Scenario fidelity?" bullets (confirmed by diffing both pairs directly — the new content is present in the templates and entirely absent from the codex-plugin/skills copies).

`packages/cli/tests/codex-plugin-catalogue.release.test.ts` and `codex-plugin-cache.release.test.ts` byte-compare `codex-plugin/skills/**` against the canonical transformation of `templates/skills/**` and will throw `Codex plugin asset differs from the canonical transformation` for both files. That's `test:release`, run by `.github/workflows/ci.yml`'s "Release-gate tests" step (explicitly commented as unbypassable) — separate from `bun run test`/`test:done`, which is why neither ticket's `verify.md` (both report green full-suite runs) caught it. **Fix:** run `bun run generate:codex-plugin` from `packages/cli` and commit the regenerated files.

packages/cli/tests/test-runner-lock.test.ts
```diff
@@ -167,6 +171,75 @@ describe('package test runner lock (379)', () => {
     expectSerializedByRunner(readEvents(logPath));
```
TheMostlyGreat, line 189: Reviewer: WARN — `expectSuccessfulSerializedRun` (defined line 121-126, called for both `first`/`second` above) still allowlists only `['', 'Waiting for another safeword package test run to finish...\n']` as valid stderr, but this PR replaced that message entirely with the new `Waiting for safeword package test lock (...)` format (`run-vitest-with-build-lock.mjs:141-143`) — the old literal can never appear in real output again. Under normal conditions here the 120ms fake-binary delay stays under the 1s report threshold so stderr is empty and this passes, but under CI/host contention that pushes the wait past 1s, the new message appears, isn't allowlisted, and the test fails with a confusing mismatch rather than a clear signal. This repo hit exactly this flake class before (the allowlist itself was added in #890 after an observed failure with this same shape). Suggest: `result.stderr === '' || result.stderr.startsWith('Waiting for safeword package test lock')`, mirroring the `startsWith` check the PR's own new test already uses at line 201.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -135,12 +187,18 @@ function acquireLock() {
       return false;
```
TheMostlyGreat, line 208: Reviewer: NOTE — the "Proceeding without safeword package test lock after waiting Nms" message just above (unchanged by this PR) still prints raw milliseconds (`waiting 1200000ms` at the 20-minute default) instead of routing through the new `formatElapsedWait` helper introduced a few lines up — inconsistent with the new periodic-status message right above it now that a formatter exists. Not a regression, just a missed consistency pass; changing it would also mean updating the raw-ms assertions in `test-runner-lock.test.ts:239` and `:337`.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -135,12 +187,18 @@ function acquireLock() {
...
     }
 
-    if (!showedWaitNotice && waitedMilliseconds >= 1000) {
-      console.error('Waiting for another safeword package test run to finish...');
-      showedWaitNotice = true;
+    if (waitedMilliseconds >= nextStatusAtMilliseconds) {
+      reportLockWait(waitedMilliseconds);
+      do {
```
TheMostlyGreat: Reviewer: NOTE — since `nextSleepMilliseconds` (line 197) is always capped at `nextStatusAtMilliseconds - waitedMilliseconds`, `waitedMilliseconds` can only ever land exactly on `nextStatusAtMilliseconds`, never overshoot it — so this `do...while` is provably a single-iteration increment. Harmless as written, but a reader has to trace the whole sleep-bound computation to confirm the loop can't repeat. A plain `nextStatusAtMilliseconds += lockStatusIntervalMilliseconds;` (or a one-line comment noting the invariant) would remove that verification burden. Not blocking.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -38,10 +38,13 @@ const childEnvironment = {
 const lockParent = nodePath.join(tmpdir(), 'safeword-test-locks');
 const lockName = 'safeword-package-test';
 const defaultMaximumLockWaitMilliseconds = 20 * 60 * 1000;
+const defaultLockStatusIntervalMilliseconds = 30_000;
+const initialLockStatusDelayMilliseconds = 1000;
 const lockDirectory = process.env.SAFEWORD_TEST_LOCK_DIR
   ? nodePath.resolve(process.env.SAFEWORD_TEST_LOCK_DIR)
   : nodePath.join(lockParent, `${lockName}.lock`);
 const ownerPath = nodePath.join(lockDirectory, 'owner.json');
+const checkoutRoot = nodePath.resolve(cliRoot, '..', '..');
```
TheMostlyGreat, line 47: Reviewer: NOTE — `checkoutRoot` assumes `packages/cli` sits exactly two directories below the repo root (hardcoded `'..', '..'`). Consistent with this file's existing `cliRoot`/`localBinDirectory` conventions, so not a new pattern, but a future repo restructuring would silently report a wrong-but-plausible checkout path in the lock-wait diagnostics rather than erroring. Informational only.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -95,11 +112,45 @@ function removeStaleLock() {
...
+    return `${milliseconds}ms`;
+  }
+
+  const totalSeconds = Math.floor(milliseconds / 1000);
+  const minutes = Math.floor(totalSeconds / 60);
+  const seconds = totalSeconds % 60;
+  return minutes === 0 ? `${seconds}s` : `${minutes}m ${seconds}s`;
+}
+
+function reportLockWait(waitedMilliseconds) {
```
TheMostlyGreat, line 148: Reviewer: NOTE — `reportLockWait` re-reads and re-parses `ownerPath` (`JSON.parse(readFileSync(...))`) independently of `removeStaleLock` (lines 90-92), which does the same read+parse with different failure handling (stat-based staleness vs. an empty placeholder object). A small shared `readOwner()` returning `undefined` on failure would remove the duplication, but this is a one-line duplication rather than a structural one — low value given the size.

packages/cli/tests/test-runner-lock.test.ts
```diff
@@ -167,6 +171,75 @@ describe('package test runner lock (379)', () => {
...
+    );
+    expect(new Set(elapsedMilliseconds).size).toBe(elapsedMilliseconds.length);
+  });
+
+  it('reports available fields when owner metadata is incomplete', async () => {
+    const temporaryDirectory = makeTemporaryDirectory();
+    const { binaryDirectory } = await createFakeTestBinaries(temporaryDirectory);
+    const lockDirectory = nodePath.join(temporaryDirectory, 'lock');
+    await mkdir(lockDirectory, { recursive: true });
+    writeFileSync(
```
TheMostlyGreat: Reviewer: NOTE — this is now the 3rd/4th near-identical `mkdir` lock dir + hand-built `writeFileSync(owner.json, ...)` + assert-on-stderr pattern in this file (also present in the pre-existing "reaps dead-owner stale locks" and "proceeds with a warning after the configured wait cap" tests). A `seedOwnerFile(lockDirectory, fields)` helper would consolidate them. This is a pre-existing pattern the PR extends rather than introduces — non-blocking, but this PR was a natural point to consolidate it.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -55,6 +58,20 @@ function resolveMaximumLockWaitMilliseconds() {
 
 const maximumLockWaitMilliseconds = resolveMaximumLockWaitMilliseconds();
 
+function resolveLockStatusIntervalMilliseconds() {
+  const raw = process.env.SAFEWORD_TEST_LOCK_STATUS_INTERVAL_MS;
+  if (raw === undefined) {
+    return defaultLockStatusIntervalMilliseconds;
```
TheMostlyGreat: Reviewer: NOTE — `resolveLockStatusIntervalMilliseconds` falls back to the 30s default on non-numeric or non-positive `SAFEWORD_TEST_LOCK_STATUS_INTERVAL_MS`, but no test exercises a malformed value (only valid `'50'`/`'100'` are used in `test-runner-lock.test.ts`). Mirrors the equally-untested existing pattern in `resolveMaximumLockWaitMilliseconds`, so this isn't a new double standard, just an existing gap this PR extends rather than closes.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -55,6 +58,20 @@ function resolveMaximumLockWaitMilliseconds() {
...
+  }
+
+  const parsed = Number(raw);
+  return Number.isSafeInteger(parsed) && parsed > 0
```
TheMostlyGreat: Reviewer: NOTE — only `parsed > 0` is enforced here, no minimum clamp. Setting `SAFEWORD_TEST_LOCK_STATUS_INTERVAL_MS=1` would print a status line roughly every millisecond of wait. Requires deliberate misconfiguration (default is 30s) and isn't reachable via normal usage — informational only.

TheMostlyGreat (commented):
Follow-up to my previous review — self-audit turned up two NOTE-level findings from the quality-review pass that didn't make it into the first batch. Adding them for completeness.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -38,22 +38,36 @@ const childEnvironment = {
...
+function resolveSafeIntegerEnvironmentVariable(name, fallback, minimum) {
+  const raw = process.env[name];
   if (raw === undefined) {
-    return defaultMaximumLockWaitMilliseconds;
+    return fallback;
   }
 
   const parsed = Number(raw);
-  return Number.isSafeInteger(parsed) && parsed >= 0 ? parsed : defaultMaximumLockWaitMilliseconds;
+  return Number.isSafeInteger(parsed) && parsed >= 0 ? Math.max(parsed, minimum) : fallback;
```
TheMostlyGreat: Reviewer: WARN — this consolidation changed what `SAFEWORD_TEST_LOCK_STATUS_INTERVAL_MS=0` means. Old `resolveLockStatusIntervalMilliseconds` required `parsed > 0` (strict), so `0` was invalid and fell back to the 30s default. Here, `parsed >= 0` accepts `0`, then `Math.max(0, minimum)` clamps it to 50ms — so `0` now means "spam every 50ms" instead of "use the default". Negative values still fall back to default (fail `parsed >= 0`), so `0` and `-1` now diverge where they used to behave identically. Confirmed by trace and by reproducing it live: `SAFEWORD_TEST_LOCK_STATUS_INTERVAL_MS=0` with a 300ms wait cap produced 5 status lines; the pre-refactor code would have produced 0. Untested at every level (no test, scenario, or AC names the exact-zero case) — worth either an explicit test asserting the new intended behavior, or reconsidering whether `0` should still mean "use default" for symmetry with negative input. (Separately: passing `minimum: 0` to this same function for `maximumLockWaitMilliseconds` above is a behavioral no-op given the `parsed >= 0` guard already in place — harmless, just worth knowing it's not doing anything.)

packages/cli/tests/test-runner-lock.test.ts
```diff
@@ -116,9 +120,33 @@ function expectSerializedByRunner(events: string[]) {
...
 function expectSuccessfulSerializedRun(result: { stderr: string; status: number | null }) {
   expect(result.status).toBe(0);
-  expect(['', 'Waiting for another safeword package test run to finish...\n']).toContain(
-    result.stderr,
+  expect(
+    result.stderr === '' || result.stderr.startsWith('Waiting for safeword package test lock'),
+  ).toBe(true);
+}
+
+function parseElapsedWaitMilliseconds(line: string): number {
```
TheMostlyGreat: Reviewer: WARN — this re-implements `formatElapsedWait`'s output grammar (ms / "Xs" / "Ym Zs") in reverse to parse it back into a number. Two issues: (1) it's duplicated inverse logic that could silently round-trip through a production formatting regression rather than catching it; (2) the 1000ms-and-above branch (lines ~139-144) has zero coverage from this suite — every calling test uses sub-1000ms elapsed windows. More importantly, `formatElapsedWait` floors to whole seconds past 1000ms, so two distinct real elapsed times in the same one-second window parse back to the *same* number — which would break the uniqueness assertion (`new Set(elapsedMilliseconds).size === elapsedMilliseconds.length`, a few lines below) if this test ever runs under enough contention/slowness to push a report past the 1s boundary. Not failing today given the fixed 250ms/50ms parameters, but it's incidental, not structural. Consider comparing raw stderr substrings positionally instead of re-deriving the format grammar.

.project/tickets/CCYD5S-explain-test-lock-waits/verify.md
```diff
@@ -0,0 +1,18 @@
+# Verification: Explain test-lock waits
+
+**Focused Proof:** ✅ `bun run test tests/test-runner-lock.test.ts` passed 8/8
```
TheMostlyGreat: Reviewer: WARN — this still says `test-runner-lock.test.ts` "passed 8/8", but the `f4195b63e` refactor commit added a 9th test (`clamps unsafe status intervals and ignores malformed settings`) — running the file now gives 9/9. `ticket.md`'s own later work-log entries already say 9/9, so this file just wasn't refreshed after the last commit. Worth a quick correction given this PR's own stated purpose.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -68,11 +82,21 @@ function isProcessAlive(pid) {
...
 }
 
-function removeStaleLock() {
-  let owner;
+function readOwner() {
   try {
-    owner = JSON.parse(readFileSync(ownerPath, 'utf8'));
+    const parsed = JSON.parse(readFileSync(ownerPath, 'utf8'));
+    return {
+      owner: typeof parsed === 'object' && parsed !== null ? parsed : {},
```
TheMostlyGreat: Reviewer: NOTE — this non-object-JSON normalization (falling back to `{}` if `owner.json` parses to a non-object) is a safe defensive addition, not a regression, but no test fixture seeds a non-object `owner.json` to actually exercise this branch. Low priority.

TheMostlyGreat (commented):
**Round 2 review** (head moved from `6a862344b` to `35c7893698f` — commits `09ac816ab` "Resolve PR review feedback", `f4195b63e` "refactor: isolate lock creation", plus evidence/doc updates). Two fresh independent subagents re-reviewed the current diff from scratch, then I cross-checked every claim below against the actual code myself (including executing the release-gate suite and the full `test-runner-lock.test.ts` file, not just reading diffs).

**Consolidated ledger — every finding raised across both rounds, current status:**

| # | Finding | Status |
|---|---|---|
| 1 | CRITICAL: codex-plugin skill copies stale, would fail `test:release` | **RESOLVED** — `bun run generate:codex-plugin` was rerun; ran the actual release gate myself: 8 files / 26 tests, all pass. `1698-keep-bdd-evidence-honest/verify.md` now documents this. |
| 2 | WARN: duplicate `resolveMaximumLockWaitMilliseconds`/`resolveLockStatusIntervalMilliseconds` | **RESOLVED** — consolidated into `resolveSafeIntegerEnvironmentVariable(name, fallback, minimum)`. See new finding below — the consolidation itself introduced an untested boundary-behavior change. |
| 3 | WARN: stale stderr-literal allowlist risked a flake under contention | **RESOLVED** — now `startsWith('Waiting for safeword package test lock')`. |
| 4 | WARN: `1698.../verify.md`'s parity claim didn't cover the codex-plugin copies | **RESOLVED** — a "Codex Plugin Release Gate" line was added specifically closing this gap. |
| 5 | NOTE: "Proceeding without lock..." message didn't use the new `formatElapsedWait` | **RESOLVED**. |
| 6 | NOTE: `do...while` catch-up loop was provably always a single iteration | **RESOLVED** — replaced with a plain increment; confirmed the invariant still holds under the new code too. |
| 7 | NOTE: `checkoutRoot` is a hardcoded, unvalidated two-levels-up path | **NOT ADDRESSED** — unchanged, still informational/non-blocking, no action expected. |
| 8 | NOTE: `reportLockWait` duplicated `removeStaleLock`'s owner-file read+parse | **RESOLVED** — both now go through a shared `readOwner()`. |
| 9 | NOTE: repeated hand-built `mkdir`+`writeFileSync(owner.json)` fixture across tests | **RESOLVED** — extracted into `seedOwnerFile()`, used consistently. |
| 10 | NOTE (follow-up): interval env var's invalid-value fallback was untested | **RESOLVED** — new test `'clamps unsafe status intervals and ignores malformed settings'` covers both the too-small and non-numeric cases (though see new finding below re: the exact-zero case). |
| 11 | NOTE (follow-up): no minimum clamp on the status interval (spam risk) | **PARTIALLY RESOLVED** — a 50ms floor now exists, but the mechanism used to add it changed behavior at a boundary nobody tested (see below). |

**New findings this round** (inline below):

- **WARN** — `resolveSafeIntegerEnvironmentVariable`'s consolidation silently changed what `SAFEWORD_TEST_LOCK_STATUS_INTERVAL_MS=0` means: previously ignored (fell back to the 30s default, since the old check was `parsed > 0`); now honored and clamped to the 50ms floor. Confirmed by code trace and by one subagent's live reproduction (5 status lines in a 300ms window with `STATUS_INTERVAL_MS=0`, vs. 0 lines pre-refactor). Untested at every level — no test, no scenario doc, no acceptance criterion names the zero case.
- **WARN** — the new `parseElapsedWaitMilliseconds` test helper re-implements `formatElapsedWait`'s output grammar in reverse. Its ≥1-second branch is currently uncovered (all exercising tests use sub-1000ms elapsed windows), and once real elapsed time crosses 1000ms, `formatElapsedWait` floors to whole seconds — so two distinct wait times in the same one-second window would parse to the *same* number, which could break the new test's uniqueness assertion under CI contention/slower runs. Not a current failure; a latent flakiness risk.
- **WARN** — `.project/tickets/CCYD5S-explain-test-lock-waits/verify.md` still claims `test-runner-lock.test.ts` "passed 8/8"; running it now gives 9/9 (the `f4195b63e` refactor commit added a 9th test). The ticket's own later work-log entries in `ticket.md` already say 9/9 — `verify.md` just wasn't refreshed after the last commit. Notable given this PR's own stated purpose is keeping evidence honest.
- **NOTE** — `readOwner()`'s new non-object-JSON normalization (falls back to `{}` unless `typeof parsed === 'object'` and `parsed !== null`) is a safe defensive addition, not a regression, but no fixture exercises a non-object `owner.json` to confirm it.

**What's still solid:** `createLock()` extraction is behavior-preserving (same try/catch, same EEXIST handling). `readOwner()` sharing between `removeStaleLock` and `reportLockWait` is correct — each keeps its own fallback semantics. The three new/changed test cases were executed (9/9 pass) and their arithmetic (e.g. the 50ms-clamp test expecting exactly 2 status lines in a 120ms window) checks out by hand-trace. Template parity and the codex-plugin fix are both independently re-verified, including by actually running the release gate rather than trusting the diff alone.

packages/cli/tests/test-runner-lock.test.ts
```diff
@@ -167,15 +195,113 @@ describe('package test runner lock (379)', () => {
...
+      pid: number;
+    };
+    const waiterRun = runNodeScript(secondRunner, ['tests/second.test.ts'], env);
+    const [ownerResult, waiterResult] = await Promise.all([ownerRun, waiterRun]);
+
+    expect(ownerResult.status).toBe(0);
+    expect(waiterResult.status).toBe(0);
+    const waitStatuses = waiterResult.stderr
+      .split('\n')
+      .filter(line => line.startsWith('Waiting for safeword package test lock'));
```
TheMostlyGreat: Reviewer: NOTE — `result.stderr.split('\n').filter(line => line.startsWith('Waiting for safeword package test lock'))` is repeated verbatim here and at line ~276. A small `waitStatusLines(stderr)` helper would remove the duplication. Low priority at just two call sites.

TheMostlyGreat (commented):
Self-audit follow-up to round 2 — one NOTE from the refactor pass didn't make it into that review. Adding it now for completeness.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -38,22 +38,38 @@ const childEnvironment = {
...
+function resolveSafeIntegerEnvironmentVariable(name, fallback, minimum, allowZero = true) {
+  const raw = process.env[name];
   if (raw === undefined) {
-    return defaultMaximumLockWaitMilliseconds;
+    return fallback;
   }
 
   const parsed = Number(raw);
-  return Number.isSafeInteger(parsed) && parsed >= 0 ? parsed : defaultMaximumLockWaitMilliseconds;
+  const isValid = Number.isSafeInteger(parsed) && (allowZero || parsed > 0);
```
TheMostlyGreat: Reviewer: WARN — `(allowZero || parsed > 0)` doesn't express "allowZero permits the zero boundary" — because `||` short-circuits, whenever `allowZero` is `true` the entire sign check is skipped, not just the zero case. Concretely verified (both an isolated repro and a full subprocess run against a real held lock): `SAFEWORD_TEST_LOCK_MAX_WAIT_MS=-5` now resolves to `0` via `Math.max(-5, 0)`, and the wrapper proceeds without the lock immediately instead of falling back to the 20-minute default — a real regression versus `main`, where this call site had an explicit `parsed >= 0` gate. The interval call site (`allowZero=false`) is unaffected and correctly still requires `parsed > 0`. No test exercises a negative value for either env var. Fix: `Number.isSafeInteger(parsed) && (allowZero ? parsed >= 0 : parsed > 0)` — this both restores the pre-regression behavior at the max-wait site and is honestly clearer than the current `||` form (a one-line comment stating the per-call-site invariant would also have caught this at write-time, since the boolean-flag parameter itself invites this kind of misreading).

.project/tickets/CCYD5S-explain-test-lock-waits/verify.md
```diff
@@ -0,0 +1,20 @@
...
+files in 422.63 seconds.
+
+**Static Checks:** ✅ ESLint, Gherkin validation, `tsc --noEmit`, Prettier, and
+`git diff --check` passed.
+
+**PR Scope:** ✅ Changes are limited to lock-owner metadata, periodic waiting
+diagnostics, the lock-creation refactor, their isolated integration coverage,
+and CCYD5S planning evidence.
+
+**Behavior Boundary:** ✅ Lock serialization, stale-owner removal, and the
```
TheMostlyGreat: Reviewer: WARN — "the 20-minute maximum wait are unchanged" is only true for the *default value*; the *validation boundary* around it changed this commit (see the inline comment on `run-vitest-with-build-lock.mjs:57` — negative `SAFEWORD_TEST_LOCK_MAX_WAIT_MS` used to safely fall back to the default and no longer does). Given this PR's whole theme is keeping evidence honest, worth tightening this claim once the underlying issue is fixed.

packages/cli/tests/test-runner-lock.test.ts
```diff
@@ -167,15 +182,137 @@ describe('package test runner lock (379)', () => {
     expectSerializedByRunner(readEvents(logPath));
   });
 
+  it('reports the owning checkout and increasing elapsed wait periodically', async () => {
```
TheMostlyGreat: Reviewer: WARN — deleting `parseElapsedWaitMilliseconds` in favor of `waitStatuses.slice(0, 2)).toEqual([...])` is a real improvement for the two elements it checks (exact string match including PID/checkout, not just `toContain`), but it's also a coverage narrowing: the prior version asserted monotonic-increase and uniqueness across the *entire* `waitStatuses` array. A realistic run of this exact scenario emits ~16 status lines (verified by reproducing it standalone), so elements 3 through ~16 are no longer checked for ordering, uniqueness, or content at all. A future off-by-one in `nextStatusAtMilliseconds` accumulation that only manifests after the first two reports would go uncaught. Consider keeping a lightweight check across the full array (e.g. `expect(waitStatuses.length).toBeGreaterThan(2)` plus a cheap monotonicity check on the raw lines, without re-deriving the elapsed-time parser) alongside the new exact-match assertions on the first two.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -68,11 +84,21 @@ function isProcessAlive(pid) {
...
+    return {
+      owner: typeof parsed === 'object' && parsed !== null ? parsed : {},
+      readable: true,
+    };
   } catch {
+    return { owner: {}, readable: false };
+  }
+}
+
+function removeStaleLock() {
```
TheMostlyGreat, line 110: Reviewer: WARN — pre-existing (unchanged since `09ac816ab`, not introduced this round), surfaced while reviewing the newly-added non-object-owner test: when `owner.json` parses to valid-but-non-object JSON (e.g. a bare string or number), `readOwner()` returns `{ owner: {}, readable: true }`. Here, `readable: true` skips the `mtimeMs`-based staleness fallback, and the synthesized `{}` has no `.pid`/`.createdAt`, so the pid-liveness and 6-hour checks are also no-ops. A lock left behind with a syntactically-valid-but-wrong-shaped owner file can never be reaped by any of the three staleness mechanisms — only the wait-cap eventually frees a waiter. A genuinely unparseable file is handled better (falls to the mtime path). Worth having `readOwner()` treat non-object JSON the same as unparseable JSON (`readable: false`) rather than treating it as a valid-but-empty owner.

packages/cli/tests/hooks/scenario-proof-fidelity-documentation.test.ts
```diff
@@ -0,0 +1,75 @@
+import { readFileSync } from 'node:fs';
+import nodePath from 'node:path';
+
+import { describe, expect, it } from 'vitest';
+
+const repoRoot = nodePath.resolve(import.meta.dirname, '../../../..');
+
+const bddTddCopies = [
```
TheMostlyGreat, line 8: Reviewer: NOTE — `bddTddCopies`/`tddReviewCopies` only list the `templates/` and `.claude/skills/` copies, not the `packages/cli/codex-plugin/skills/**` copies also touched by this PR. Not a functional gap today (directly diffed all three — the codex-plugin copies do contain the new content, differing only by the expected `$safeword:` command-syntax substitution), and the separate release-catalogue tests already guard future drift there — just noting this new test relies on that other suite rather than covering the third copy itself.

TheMostlyGreat (commented):
**Round 3 review** — new commit `26ee17e6c` "Fix lock runner review findings" addresses round 2's items. Two fresh independent subagents re-reviewed from scratch (no memory of prior rounds), converged independently on the same headline issue via their own standalone reproductions, and I verified it myself with a direct trace before posting.

**Prior-round status recap:** all of round 1's 10 items and round 2's 4 new items are RESOLVED, except `checkoutRoot`'s hardcoded two-levels-up path (still open, informational, no action expected). Specifically this commit correctly: deleted the `parseElapsedWaitMilliseconds` round-trip parser in favor of deterministic exact-string assertions (verified non-flaky — the underlying `waitedMilliseconds` is a simulated counter, not wall-clock, so `50ms`/`100ms` are exact by construction); extracted `waitStatusLines()` for the duplicated filter (confirmed used at all 4 remaining call sites, no leftovers); added a real subprocess-backed test for non-object `owner.json`; added a `'0'`-interval test case confirming the interval call site's fallback is correct; and updated `CCYD5S-explain-test-lock-waits/verify.md` to the accurate "10/10" count.

**New finding this round (the one that matters):** fixing the interval call site's zero-boundary added a 4th `allowZero` parameter to the now-shared `resolveSafeIntegerEnvironmentVariable`, and the boolean logic used — `allowZero || parsed > 0` — doesn't do what the commit message says. When `allowZero` is `true` (the default, used at the *other* call site, `SAFEWORD_TEST_LOCK_MAX_WAIT_MS`), that `||` short-circuits to `true` unconditionally, meaning **the sign check is dropped entirely for that call site, not just the zero-boundary.** Both subagents independently reproduced this end-to-end (spawning the real script against a real held lock) and confirmed: `SAFEWORD_TEST_LOCK_MAX_WAIT_MS=-5` now resolves `maximumLockWaitMilliseconds` to `0` and the wrapper proceeds without the lock on the very first contention — where, before this exact commit, a negative value correctly fell back to the 20-minute default (confirmed via `git show main:...` and the immediate parent commit, both of which had an explicit `parsed >= 0` gate). No test anywhere exercises a negative value for either env var, so this shipped untested. Full details and a corrected expression are in the inline comment.

Three more real findings below, all independently verified by execution or direct trace, not just read.

TheMostlyGreat (commented):
Self-audit follow-up to round 3 — two low-priority items the agents flagged as hedged/low-value didn't make it into the main review. Posting them for completeness rather than silently dropping them.

.project/tickets/1698-keep-bdd-evidence-honest/verify.md
```diff
@@ -0,0 +1,29 @@
+# Verification: Keep BDD evidence honest
+
+**Test Suite:** ✅ `bun run test:done` passed 1,244/1,244 tests across 81 files.
+
+**Full Suite:** ✅ `bun run test` passed 5,645 tests with 5 skips across 377
```
TheMostlyGreat: Reviewer: NOTE (low confidence, unverified) — this ticket's verify.md reports "5,645 tests... across 377 files" while `CCYD5S-explain-test-lock-waits/verify.md` reports "5,647 tests... across 377 files" for what should be the same full-suite run at a later point in the same PR. `test-runner-lock.test.ts` grew from 6 to 10 `it()` blocks across the PR's commits, which doesn't cleanly reconcile with a +2 delta between these two snapshots — though this is plausibly just snapshot-ordering across an interleaved commit history rather than a real inconsistency. Not independently re-verified (would mean re-running the full ~5,647-test suite, which risks contending with this PR's own global test lock) — flagging as an unverified low-confidence gap rather than asserting it's wrong.

packages/cli/tests/test-runner-lock.test.ts
```diff
@@ -167,15 +182,137 @@ describe('package test runner lock (379)', () => {
...
+
+    expect(result.status).toBe(0);
+    expect(result.stderr).toContain(`owner PID ${process.pid}`);
+    expect(result.stderr).toContain('checkout unavailable');
+    expect(result.stderr).toContain(
+      'Proceeding without safeword package test lock after waiting 220ms.',
+    );
+  });
+
+  it('reports unavailable fields when owner metadata is not an object', async () => {
```
TheMostlyGreat: Reviewer: NOTE (low priority) — this new test and the existing 'reports available fields when owner metadata is incomplete' test repeat the same five-step shape (temp dir + fake binaries + lock dir + seed owner file + run + assert on stderr), differing only in the seeded payload, a couple of env values, and the final assertions. A shared `runWithSeededOwner(owner, args, envOverrides)` helper could cut the duplication, but this same shape recurs across nearly every test in this file already — it's consistent with, not worse than, the file's prevailing style, so not a regression introduced by this PR specifically.

TheMostlyGreat (commented):
**Round 4 review.** Note: this branch was rebased onto a newer `main` between round 3 and now (confirmed via merge-base ancestor check — round 3's reviewed commit is no longer in this branch's history), which pulled in ~35 unrelated main commits into the raw commit log. I diffed against the actual current merge-base (`git diff origin/main...HEAD`) and cross-checked against `gh pr view --json files` to make sure both fresh reviewers were looking at the PR's real 17-file scope, not rebase noise.

On top of the rebase, one new content commit landed: `3c79e9941` "Fix lock runner review findings" — and it's thorough. Two fresh independent subagents verified all of round 3's items by executing code, not just reading:

- The `allowZero ? parsed >= 0 : parsed > 0` fix is exactly correct — traced and executed against both call sites across a wide boundary sweep (negative, zero, positive, NaN, Infinity, strings). No remaining asymmetry.
- `readOwner()` now correctly treats non-object JSON (including `null` and arrays) as unreadable, routing it through the existing mtime stale-lock recovery — verified this actually changes real reap behavior, not just the reported diagnostic strings.
- The negative-max-wait regression now has a real test (two genuine racing processes against a real lock) — confirmed it would have failed under the pre-fix bug and passes now.
- The periodic-status test was restored to asserting the *full* 4-line sequence (not just the first 2) — confirmed this isn't flaky since the underlying counter is a simulated tick count, not wall-clock time.
- The non-object-owner test now proves an actual reap happened (build/vitest events fire) rather than just asserting diagnostic strings — confirmed this would catch a regression the old version couldn't.
- The 5,645-vs-5,647 test-count question from round 3 is now explicitly reconciled in both tickets' verify.md as two different snapshots.

**All of round 1-3's items are now resolved**, except `checkoutRoot`'s hardcoded two-levels-up path (still open, informational only, no action expected).

**Two real findings survived this round**, both independently confirmed by both reviewers:

packages/cli/tests/test-runner-lock.test.ts
```diff
@@ -102,10 +113,9 @@ function expectSerializedByRunner(events: string[]) {
     const parentEvents = events.filter(event => event.includes(`:${parentId}`));
     expect(parentEvents[0]).toBe(`build:start:${parentId}`);
     expect(parentEvents[1]).toBe(`build:end:${parentId}`);
-    expect([
-      `vitest:start:${parentId}:run,tests/first.test.ts`,
-      `vitest:start:${parentId}:run,tests/second.test.ts`,
-    ]).toContain(parentEvents[2]);
+    const vitestStartEvent = parentEvents[2] ?? '';
```
TheMostlyGreat: Reviewer: WARN — this fix commit loosened the shared `expectSerializedByRunner` helper's vitest-start assertion from an exact two-string match to `startsWith(prefix) + endsWith('.test.ts')`, apparently just to accommodate the new negative-max-wait test's `tests/owner.test.ts`/`tests/waiter.test.ts` filenames. That test's filenames are purely symbolic (consumed by a stub binary that just logs argv) — it could have reused the existing `tests/first.test.ts`/`tests/second.test.ts` convention like the other two call sites still do, avoiding the need to touch a helper shared across all 3 tests that use it. Both reviewers independently confirmed by simulation that this is a real loss of discriminating power, not just style: a bug that passed a wrong or mangled filename to the child process (e.g. a future off-by-one in path-rebasing, or an argument mixup between the two racing processes) would fail the old exact-match assertion but now silently passes, since it only needs to start with the right prefix and end in `.test.ts`. Suggest either reverting the new test to the `first`/`second` naming convention, or parameterizing `expectSerializedByRunner` with the two expected filenames per call site (it already has them as literal args to `runNodeScript`) instead of loosening to prefix/suffix for everyone.

.project/tickets/CCYD5S-explain-test-lock-waits/test-definitions.md
```diff
@@ -0,0 +1,73 @@
...
+And malformed and zero interval settings fall back to the default interval
+
+### Scenario: CCYD5S.SM1.AC6.negative_maximum_wait_uses_the_default
+
+Given another runner owns a live package-test lock
+When a waiting runner sets its maximum wait to a negative value
+Then it waits for the owner under the default maximum wait
+And it does not proceed without the lock or overlap the owner's build and test
+
+- [x] RED skip: unsafe-interval test failed before the minimum clamp
```
TheMostlyGreat: Reviewer: WARN — this RED/GREEN/REFACTOR checklist ("unsafe-interval test failed before the minimum clamp" / `GREEN 09ac816ab`) is describing AC4's interval-clamping test, but it's now positioned directly under the newly-inserted AC6 scenario (line 44) instead of under AC4 (line 37) where it belongs. Net effect: AC4 has no evidence checklist of its own anymore, and AC6 has inherited AC4's by proximity. Given this PR's whole point is keeping scenario evidence honest, this specific kind of misattribution is exactly what it's meant to prevent — worth moving AC4's checklist back directly under its own scenario and giving AC6 its own (referencing whichever commit actually GREENs the negative-max-wait test).

.project/tickets/CCYD5S-explain-test-lock-waits/user-stories.md
```diff
@@ -0,0 +1,55 @@
...
+
+#### CCYD5S.SM1.AC4 - Diagnostic configuration stays safe
+
+Given a maintainer supplies an invalid, zero, or excessively small diagnostic interval
+When a command waits for the package-test lock
+Then invalid and zero values use the default interval
+And a small positive value uses the safe minimum interval to keep wait output
+readable.
+
+#### CCYD5S.SM1.AC6 - Maximum-wait configuration remains safe
```
TheMostlyGreat: Reviewer: NOTE (cosmetic) — AC6 is inserted between AC4 and the pre-existing AC5, breaking numeric order (AC4, AC6, AC5) in both this file and the corresponding `test-definitions.md`. No functional impact, but appending AC6 after AC5 would read more naturally for future maintainers.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -38,22 +38,38 @@ const childEnvironment = {
...
+function resolveSafeIntegerEnvironmentVariable(name, fallback, minimum, allowZero = true) {
+  const raw = process.env[name];
   if (raw === undefined) {
-    return defaultMaximumLockWaitMilliseconds;
+    return fallback;
   }
 
   const parsed = Number(raw);
-  return Number.isSafeInteger(parsed) && parsed >= 0 ? parsed : defaultMaximumLockWaitMilliseconds;
+  const isValid = Number.isSafeInteger(parsed) && (allowZero ? parsed >= 0 : parsed > 0);
```
TheMostlyGreat, line 57: Reviewer: NOTE (pre-existing, not a round-4 regression) — `SAFEWORD_TEST_LOCK_MAX_WAIT_MS=''` (empty string) still resolves to `0` at this call site (`allowZero=true`), silently meaning "don't wait," same as before this PR. Low real-world likelihood, just noting it wasn't addressed by this round's fix.

packages/cli/scripts/run-vitest-with-build-lock.mjs
```diff
@@ -68,11 +84,22 @@ function isProcessAlive(pid) {
   }
 }
 
-function removeStaleLock() {
-  let owner;
+function readOwner() {
```
TheMostlyGreat, line 95: Reviewer: NOTE (pre-existing gap, not new) — a syntactically-valid owner.json that parses to `{}` (or any object missing `pid`/`createdAt`) still gets `readable:true` with no fields to act on, so none of the three staleness mechanisms (mtime, pid-liveness, 6-hour age) can ever reap it — only the wait-cap eventually frees a waiter. This existed identically before this PR (old code was `readable:true` for any valid JSON). Worth a future ticket, not a blocker here.

packages/cli/tests/test-runner-lock.test.ts
```diff
@@ -167,15 +188,172 @@ describe('package test runner lock (379)', () => {
...
-  it('reaps dead-owner stale locks before acquiring', async () => {
+  it('uses the default maximum wait when configured with a negative value', async () => {
     const temporaryDirectory = makeTemporaryDirectory();
     const { binaryDirectory, logPath } = await createFakeTestBinaries(temporaryDirectory);
     const lockDirectory = nodePath.join(temporaryDirectory, 'lock');
-    await mkdir(lockDirectory, { recursive: true });
-    writeFileSync(
-      nodePath.join(lockDirectory, 'owner.json'),
-      `${JSON.stringify({ createdAt: new Date().toISOString(), pid: 2_147_483_647 })}\n`,
+    const ownerPath = nodePath.join(lockDirectory, 'owner.json');
```
TheMostlyGreat: Reviewer: NOTE (low priority) — `nodePath.join(lockDirectory, 'owner.json')` is repeated here and at line 222; a trivial helper would remove the duplication but it's one line each, not worth forcing. Separately, this test and the one at line ~222 share an identical 4-line "spawn owner -> waitForPath -> spawn waiter -> Promise.all" shape that could be extracted into a `raceOwnerAndWaiter(...)` helper — but the inline form keeps each test's differing env vars visible, so leaving it as-is is a reasonable call.

TheMostlyGreat (commented):
Self-audit follow-up to round 4 — three low-priority items both agents surfaced but explicitly hedged as pre-existing/not-worth-forcing didn't make it into the main review. Posting for completeness; none of these block anything.

Issue #1910, opened 2026-08-04
Validate PR review with asymmetric retrospective evidence


TheMostlyGreat:
Parent: #1908

## User value

Before builders trust Safeword's ready-vs-human routing, prove that retrospective evidence ranks reviewers the same way recorded human judgments did—without consuming new engineer time.

## Label contract

- A finding acted on later is a **strong positive**.
- A cleanly merged PR is a **weak negative** and cannot, by itself, penalize a finding.
- PR 2118 is the calibration warning: it merged clean but an Arcade engineer called its finding “a good comment.”
- Eight of 42 sampled Arcade PRs had zero human review.

## Falsification-first experiment

1. Inspect the real local Claude JSONL structure.
2. Mine approximately 20–30 `diff → finding → acted on?` pairs from local sessions and git history.
3. Run two deliberately different PR-review prompts against the frozen pairs.
4. Compare the retrospective ranking with Nate's recorded 10-PR triage plus the PR 2118 exchange.
5. Stop cheaply if the rankings disagree. Scale mining only if they agree.

## Contamination guards

- Post-model-cutoff bugs only.
- Never provide a log's outcome to the review being graded.
- Hold the corpus out; do not tune prompts on it.
- Spot-check SZZ links.
- An LLM jury may pre-filter but is never the gate.
- No new engineer review time.

## Acceptance

- Frozen examples, provenance, prompt versions, raw outputs, and scoring code are reproducible.
- The asymmetric metric cannot lower a finding's score solely because its PR merged cleanly.
- Both prompt rankings and the recorded-human ranking are reported, including uncertainty.
- SZZ links receive documented manual spot checks using already available evidence.
- A clear go/kill decision is recorded before any large miner is built.

TheMostlyGreat:
SAFEWORD_TERRA_CANARY_AUTHORIZATION_V1
[...]

PR #1917, opened 2026-08-04
Review substantive PRs without repository-specific setup

TheMostlyGreat:
## Outcome

Ships the smallest trustworthy advisory PR-review release: ready pull requests are inspected at an exact revision, unfamiliar artifacts receive a technology-neutral integrity review, consequential uncertainty routes to a human, and Safeword never merges or modifies customer code.

Part of #1909. Parent epic: #1908.

This remains a draft while repository CI runs and the protected release environment is configured.

## What changed

- Added the `review-pr` inspection and publication commands with deterministic evidence, bounded OpenAI review, conservative routing, freshness, and one marker-owned receipt.
- Added ready/material-update and scheduled reconciliation workflows with per-PR serialization.
- Kept fork inspection read-only and non-executing. Fork results cross a JSON artifact into a trusted `workflow_run` publisher that checks out no PR code and receives no model secret.
- Made advisory review strictly opt-in: missing, malformed, or false configuration installs no workflows; literal `true` installs all three managed workflows.
- Added pinned `actionlint` validation and a release-blocking disposable-repository compatibility smoke.

## Safety boundary

- Model text is evidence, never proof that a remedy or test ran.
- Missing, incomplete, stale, failed, over-budget, or consequentially unfamiliar evidence cannot produce `looks ready`.
- Publication is constrained to ordinary marker-owned issue comments; the live smoke proves it does not create reviews, statuses, checks, or mergeability changes.
- Model credentials exist only in the inspection job and are absent from every write-capable job.

## Verification

- Live OpenAI Flux regression: **1 scenario / 39 steps passed** and routed the access-control change to a human.
- Live disposable GitHub proof: event run `31116176245`, trusted publisher `31116192147`, and scheduled projection `31116229231` all succeeded.
- Concurrency proof: one active and one pending `pr-review-1` lease; exactly one current receipt remained.
- Merge-neutrality proof: reviews, statuses, checks, and mergeability were unchanged after publication.
- Workflow and smoke contracts: **6/6 passed**.
- Schema contracts: **36/36 passed**.
- Release contracts: **4/4 passed**.
- Push-time schema drift gate: **753/753 passed**.
- ESLint, TypeScript, Prettier, actionlint, architecture boundaries, and the final diff-scoped audit passed.
- Both disposable repositories were permanently deleted after the proof.

## Release setup

Before release, configure the protected GitHub environment `pr-review-smoke` with `SAFEWORD_PR_REVIEW_SMOKE_TOKEN` and the documented owner variables. npm publication now fails closed unless that live compatibility proof succeeds.

## Out of scope

Automatic merge, automatic fixes, executing untrusted fork code, retrospective calibration (#1910), independent-vendor selection (#1911), deeper BDD/TDD assessment (#1912), and budget/noise policy (#1913) remain separate work.


TheMostlyGreat (commented):
Reviewer: `/review-spec` + `/quality-review` + `/refactor` pass over `042fc6d`. Advisory only — this is a `COMMENT` review and carries no merge block.

**Verdict: REQUEST CHANGES.** The safety architecture is sound and the workflow split is right. Two things need a decision before this ships, and four scenarios are green against something other than what customers will actually see.

---

## 1. Disabling PR review after a version bump leaves the workflows installed

`prReviewWorkflowFile.removeIfUnmodified` returns the template with the **current** `VERSION` already interpolated, and `planOmittedManagedRemoval` removes only on exact byte equality (`reconcile.ts:333-337`).

Sequence that breaks:

1. Customer installs at `0.70.0` — three workflows land, pinned `safeword@0.70.0`.
2. Customer upgrades to `0.71.0` **and** sets `prReview.enabled: false` in the same `setup`.
3. Step 3 (`planManagedFiles`) skips — the generator returns `undefined`.
4. Step 3b compares on-disk `0.70.0` bytes against `0.71.0` expected bytes → mismatch → **file kept**.

The customer is left with three `pull_request_target` workflows carrying `secrets: inherit`, running `safeword@0.70.0` on every PR, after opting out. `workflow-contract.test.ts:264-277` only exercises the same-version disable, so nothing catches it.

Fix: normalize `__SAFEWORD_VERSION__` before comparing (or match any version pin), and add the upgrade-and-disable case to the contract test.

## 2. `npm publish` now depends on a live two-org GitHub proof

`release.yml` makes `publish` need `advisory-pr-review-smoke`. That job creates repos under two different owners, forks one into the other, sets an environment secret, drives Actions end-to-end, polls a dated preview API (`X-GitHub-Api-Version: 2026-03-10`, `actions/concurrency_groups`), and deletes both repos — inside a 15-minute budget with a 300s `waitFor` default.

Every future release — including unrelated patches — now fails on PAT expiry, Actions queue latency, rate limiting, or that preview API changing shape. This repo already has a history of releases silently drifting from main; adding a live external dependency to the publish path raises that risk rather than lowering it. The PAT also needs `delete_repo`, which is a meaningful blast radius to keep in the release pipeline.

Worth a deliberate call: keep it blocking only for minor/major cuts, or run it scheduled and require a recent green, rather than gating every publish.

---

## Scenario findings (`/review-spec`)

The suite is genuinely strong on the routing lattice — the precedence table, the three-way stale scenario, and the empty/binary-only split are well built and would each fail under a real mutation. Four scenarios, though, are green against an artifact the customer never reads.

The spec's vocabulary uses "current receipt" for two different things: the JSON `PublishedReceipt` and the rendered comment body. The scenarios assert against the first; humans read the second.

**must-fix — `safe-advisory-core.NTB1.R1` "Available evidence is reported with its actual values" cannot be true in production.**
`reviewedReceiptBody` hardcodes `checks: []`, `skippedChecks: []`, `tokenUsage: {}` (`review-pr-publication.ts:212-226`). The scenario passes because the step calls `renderReceipt` directly with a hand-built view (`steps/...:1027-1043`). Every real receipt says `Checks: none` and `Token usage: unknown input, unknown output`. Either wire usage and check results through the handoff, or narrow the AC to what the MVP publishes.

**must-fix — `safe-advisory-core.TBU1.R2` coverage assertions never reach the receipt.**
`renderReceipt` emits no `coverage`, `missingEvidence`, or `reviewableTextArtifacts` lines. "The current receipt marks `<path>` as integrity-reviewed" is true of the JSON and false of the comment — which is exactly the evidence R2 exists to make visible. `review-pr-publication.test.ts:37` even passes coverage in and asserts nothing about it appearing.

**must-fix — `safe-advisory-core.NTB1.R2` "any model-proposed remedy is labeled unverified" is vacuous.**
`unverifiedRemedy` is never set on the production path; the model's `nextAction` prose renders as a plain `Next action:` line with no unverified labeling. The step supplies the field itself. Given the PR's own stated boundary ("model text is evidence, never proof that a remedy ran"), the shipped receipt doesn't carry that label.

**must-fix — `safe-advisory-core.SWM1.R1`'s fork scenario proves a module production never runs.**
`runSplitPrivilegeReview` is referenced only by the steps file. On the shipping path both audits are constant literals (`review-pr.ts:43`, `review-pr-publication.ts:282`), so `publishValidatedSplitPrivilegeEvidence` validates a constant the same process just built. The real enforcement — `workflow-contract.test.ts` plus the YAML — is good and sufficient; the scenario just shouldn't claim audit evidence it doesn't have.

**should-strengthen — `safe-advisory-core.TBU1.R4` stale-write ordering isn't observable through the real entry point.**
`reviewPullRequest` publishes a 3-key stale receipt before the fresh one, but `inspectPullRequestCommand`'s `publish` callback just overwrites `published`, so only the last receipt reaches the handoff — and that 3-key stale shape would actually be **rejected** by the publisher's `hasValidReceiptShape` if it ever escaped. The `review-pr invalidate` command already owns that write. Consider dropping the intra-run stale publish rather than keeping an unreachable branch.

## Notes (non-blocking)

- Both public handlers swallow the cause in `catch {}` (`public-handlers.ts`). A failing inspect surfaces only `PR_REVIEW_INSPECT_FAILED` with no detail, in a job whose logs are the only diagnostic a customer has.
- The 5-minute sweep re-selects any PR whose receipt lacks a `Route:` line. A PR stuck at `prerequisites pending` re-runs 2-3 jobs every five minutes indefinitely. No model spend, but permanent Actions churn.
- Marker semantics live in three places under two rules: line-exact in TypeScript (`hasExactReceiptMarker`) vs `startswith(...)` in the jq at `pr-review.yml:44` and `pr-review-worker.yml:69`. They agree today only because the marker is always written first.
- `PullRequestReviewState.requiredPrerequisites` is set only by tests; production resolves prerequisite state before calling, so the `resolvePrerequisites` empty-list branch is dead on the real path.

## Verified as correct (`/quality-review`, sources fetched this session)

- `actions/upload-artifact@043fb46…` resolves to `v7.0.1` and `actions/download-artifact@3e5f45b…` to `v8.0.1` — both the current latest releases, SHA-pinned. `actionlint v1.7.12` is current, pinned by tarball SHA256.
- The `pull_request_target` → read-only inspect → JSON artifact → `workflow_run` publisher split matches GitHub's documented secure-use guidance. No fork checkout anywhere; `.safeword/config.json` is read from the default branch via the contents API, not the PR head. Model credentials appear only in the inspect job.
- The OpenAI call is `store: false`, `tools: []`, strict JSON schema, with findings rejected unless their path was in the supplied evidence — solid injection containment for untrusted diff text.
- `check-pr-review-workflows.ts` pairs the actionlint run with a deliberately-invalid control fixture. That's the right shape for a lint gate.

## Refactors pushed to this branch (`/refactor`, behavior-preserving)

Three commits, no semantic change — the bugs above are deliberately left for you to decide on:

- `refactor: reuse the exact-marker predicate in invalidate` — `invalidatePullRequestCommand` had an inline copy of `hasExactReceiptMarker`'s line-split logic.
- `refactor: extract the shared receipt evidence projection` — the artifact→evidence mapping was duplicated verbatim across the try and catch arms of `inspectPullRequestCommand`.
- `refactor: name the inspect-side receipt parser for its contract` — two module-local functions were both called `parseReviewedReceipt` with very different validation strength; the inspect-side one is now `parseOwnReceipt`, with a comment saying why the publisher's copy is stricter.

**Next:** decide on the release-gating question (2) first — it blocks every future publish, not just this one — then fix the version-skew removal (1) with a test, and reconcile the four scenarios against the rendered receipt.

PR #1938, opened 2026-08-04
Make temporary workarounds fail CI when they become removable

TheMostlyGreat:
Closes #1907.

Codifies the upstream-workaround tripwire pattern, then applies it to a real workaround already in this repo.

## The pattern

A workaround for someone else's bug is temporary by intent and permanent in practice. A comment only reaches someone already reading that file; a ticket ages out of triage; a reminder is tied to something that ends. The workaround outlives its cause, and the next person to touch it sees pointless-looking indirection and "cleans it up," reintroducing the original bug.

The fix is a test that fails when the workaround becomes **removable** — asserting the dependency is still pinned at the last known-bad version, so the next upgrade turns CI red with a failure message naming what to delete.

- `testing-guide.md` — when a tripwire is warranted, the assert-the-pin-not-the-bug inversion, the header runbook contract, the lane it runs in
- `tripwire-template.md` — fillable header + test scaffold, registered in `schema.ts`
- `testing` and `debug` skills — trigger at the moment the decision is live; `debug` Phase 4 now asks for a tripwire when the root cause is upstream and the fix is a workaround

## First real use

Safeword's Codex plugin is profile-scoped, so its lifecycle hooks would fire in every repository on the machine. The workaround is a synthetic enrollment marker (`.safeword/SAFEWORD.md`) checked by every Codex project gate; its removal depends on openai/codex#18115. That dependency was recorded only in an ADR "Reassess when" row and a few ticket triggers — nothing anyone reads on a dependency bump.

Two tripwires now cover it from opposite directions, because each alone can sit silently green:

- **The pin** (`codex-project-scope-tripwire.test.ts`) fires when *we* upgrade Codex. Verified to fail on a version bump.
- **The issue watcher** fires when *upstream* moves, via the existing weekly monitor — one source entry, no new infrastructure. Verified in both directions against the committed snapshot.

The pin's header states its own limitation outright: it guards a devDependency, not the Codex build users run. openai/codex#18115 was confirmed still open while preparing this PR, so the seeded snapshot state is verified rather than assumed.

## Fixes found while building

- **Source isolation.** `runUpstreamMonitor` had no per-source error handling, so one failing source aborted the run. The watched-issue source is checked third, behind two network fetches — either one throwing meant the tripwire silently never ran. Failures are now logged distinctly from "no change" and the runner exits non-zero.
- **Duplicate issues.** Existing-issue lookup used the search API, which is eventually consistent; a manual re-run after a scheduled one filed a duplicate. It now lists open issues and excludes PRs. Exhausting the page cap **throws** — see the correction below.
- **Parser fragility.** A `>` inside a quoted attribute spilled markup into page text; CDATA-wrapped Atom titles leaked their markers into the snapshot; invalid numeric character references crashed the scraper. All three would corrupt or disable a watch.
- **Timeouts.** Every request now carries a 30s ceiling.
- **Dead code.** `SourceChange.hash`, `MonitorDependencies.now()`, and `SourceChangeInput.currentFetchedAt` were all computed or declared and never read.

Every fix has a test verified to fail without it.

## Correction — a bug review caught in this PR's own code

An earlier revision of this description claimed the capped issue lookup "reports hitting the page cap rather than treating it as 'no such issue'." That was true of the log line and **false of the behavior**: it logged, then returned `undefined`, and `reportSourceChange` reads `undefined` as "no existing issue" — so it created the duplicate the cap existed to prevent. A log line is not control flow.

Fixed in `4a5b67d` by throwing; per-source isolation turns that into a recorded failed check rather than a lost run. Left here rather than quietly edited away, because it is the third instance in this PR of logging where the code needed to refuse — the same failure mode the shipped pattern is about.

## The sealed-review test (#1918)

CI requires `plugin/` to be regenerated and committed; ticket 93C14D's sealed manifest appeared to forbid it. That looked like a design conflict. It was a test bug.

`git show <sealed-commit>:plugin/runtime/cli.js` failed with a spawn-level buffer error — the file is ~1.6 MB against `spawnSync`'s 1 MiB default — and `reviewedInput`'s fallback then read the working tree. For any sealed input over that limit, the check silently stopped comparing the bytes the reviewer attested to and started demanding an unchanged working tree. The seal was intact the whole time.

Raising `maxBuffer` restores the intended comparison. **No sealed evidence is edited and no reviewer attestation is re-minted** — the tempting fix was to update the recorded hash, which would have voided a real attestation to work around a buffer size. Spawn-level failures now throw in both `reviewedInput` and `sealedCommit` rather than silently comparing different bytes.

## Inherited CI repair — and the debt it carries (#1973)

`1d980a5` fixes two Codex-migration scenarios that were failing identically on `main` (`fba74dd`) and therefore on every PR branched from it. The cause was fixture non-determinism, not product behavior: the scenario named *"Unavailable profile enrollment…"* never made enrollment unavailable, so wherever Codex was on `PATH` it measured the **available** path and exited `0` instead of the asserted `2`. Setting `PATH: ''` makes the boundary real. **Acceptance setup only — no product code changed.** See #1974 for why this should not be treated as a product bug.

That fix has a known cost, tracked in **#1973**: `PATH: ''` sits on the *shared* `When the plugin migration upgrade runs` step, which serves three scenarios. The two that are not about unavailability now run a migration that fails immediately and then assert user data is intact — passing because nothing ran. Fixing that properly needs a deterministic *available*-Codex fixture, which is Codex-migration work rather than tripwire work, so it is filed rather than widened into this PR.

## Note for anyone regenerating the plugin locally

`plugin/runtime/cli.js` is produced by `Bun.build`, so its bytes depend on the bun version running the generator. The repo pins `bun@1.3.14`; an older bun on `PATH` emits different code for identical source and can never match the freshness gate. `bun run generate:claude-plugin` re-invokes `bun` from `PATH`, so the pinned version must be *first on PATH* — passing a different bun binary to `bun run` has no effect. The generator now carries this warning inline.

## Verification

`/verify`, `/audit`, `/quality-review`, and `/refactor` were run three times: at the first complete draft, again over the commits that landed after it, and again at the current head. Full suite 6,561 passing; Gherkin 910 scenarios; build, typecheck, lint, `bun audit`, depcruise clean. All four CI checks green at the head.

Snapshot wiring is table-driven across all four monitor sources: the declared path must resolve, the header must name that source, and the body must survive the real parser. Mutation-verified — deleting a snapshot fails on ENOENT, and swapping in another source's snapshot fails on the header assertion.

Two pre-existing local failures are unrelated — `chmod`-based failure-injection tests that don't fail as root, confirmed identical on a clean tree (#1920). They pass in CI.

GitHub API claims were verified against current docs rather than memory: `state_reason` includes `null`, `X-GitHub-Api-Version: 2022-11-28` is supported until 2028-03-10, and unauthenticated reads are 60/hour per IP.

**Independent review.** An earlier revision of this description said the coordinator "could not run (`independence: none`)". That is no longer accurate. With extended timeouts it completes: the author-side run returned APPROVE at `independence=cross-agent`, and a later Claude-authored pass returned `request_changes` at `independence=degraded` (Claude reviewing Claude, since no system-level Codex is installed here — the coordinator correctly refuses the repo-local one). Under the provenance rule a degraded verdict informs rather than blocks. Its two warnings were triaged: the fence-injection finding does not reproduce — every diff line carries a `+`/`-` prefix, so upstream content cannot close the fence, and `issues.test.ts:10` pins exactly that. The Cursor-scraper scope warning is real but pre-existing and out of scope here.

https://claude.ai/code/session_01UPpTdcLAj7JkXzSpigWk1v


TheMostlyGreat (commented):
## Quality Review

**Currency:** ✓ Current as of 2026-08-04. The watched Codex issue remains open; `@openai/codex` 0.146.0 is current while the deliberate 0.145.0 pin remains the last known-bad tripwire.
**Sources:** ✓ GitHub REST behavior, rate limits, API-version support, the upstream Codex issue, and the npm version were checked against current primary sources.
**Correct:** ✓ after one blocking fix. The capped open-issue lookup used to log uncertainty and return `undefined`, which made `reportSourceChange` immediately create the duplicate the cap was meant to prevent. It now throws and the per-source monitor records a failed check instead.
**Elegant:** ✓ The monitor keeps one source-isolation path and one issue-client boundary.
**No-bloat:** ✓ Three dead/repeated mechanics were removed without adding abstractions.
**Wiring:** ✓ `packages/cli/tests/upstream-monitor/wiring.test.ts` uses the real source definition, real committed snapshot, real file reader, and real monitor flow; only network and GitHub issue mutation are boundary fakes.

**Verdict:** NEEDS DISCUSSION

**Critical issues:** None after `4a5b67d97`.

**Suggested improvements:** A table-driven committed-snapshot wiring assertion for the other three monitor sources would close the same path/shape gap, but it is non-blocking because missing snapshots already fail the scheduled run red.

**Changes pushed:**

- `4a5b67d97` — refuse duplicate issue creation after a capped lookup.
- `94d0529f2` — remove the unused monitor clock dependency.
- `08d713425` — remove the unused fetch timestamp field.
- `1dd320062` — lowercase Atom input once instead of reallocating it per entry.

**Verification:** 31/31 touched tests pass; 910/913 acceptance scenarios pass with 3 intentional skips (34,443 passing steps); lint and typecheck are clean; diff audit reports no dependency violations or config drift; `bun audit` reports no vulnerabilities. The repository-wide Vitest run became non-diagnostic under local contention, with unrelated subprocess-heavy tests timing out across review, stop/retro, host adapters, setup, and override suites; it was stopped after the distribution clearly exceeded this PR scope. The pushed CI matrix is the full-suite signal.

**Provenance:**

- verified: https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues
- verified: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api
- verified: https://docs.github.com/en/rest/about-the-rest-api/api-versions
- verified: https://github.com/openai/codex/issues/18115
- verified: https://www.npmjs.com/package/@openai/codex/v/0.146.0

The first independent cross-agent pass approved the bounded authored diff and surfaced the dead clock dependency. The required post-change re-review timed out twice (`preferred_failure=timed_out`, fallback `invalid_output`), so I am preserving that evidence limitation instead of self-issuing an approval.

**Next:** Let the pushed CI matrix finish, then rerun the independent quality-review coordinator or have a human reviewer confirm the four follow-up commits.

TheMostlyGreat (commented):
## Quality Review — latest head

**Currency:** ✓ Current as of 2026-08-04 (Hawaii). `openai/codex#18115` remains open; npm still reports `@openai/codex` 0.146.0 as latest while this PR deliberately pins the last known-bad 0.145.0 tripwire.
**Sources:** ✓ GitHub issue/API behavior, npm package state, WHATWG numeric-character-reference handling, and JavaScript `String.fromCodePoint` bounds were checked against current primary sources.
**Correct:** ✓ The latest snapshot-wiring update covers all four committed monitor snapshots. A fresh cross-agent pass found one additional robustness edge: an out-of-range numeric HTML entity could throw and permanently leave the Cursor watch red. `ef4a1539c` now follows the HTML parser contract and emits U+FFFD for invalid scalar references.
**Elegant:** ✓ `2d399c65c` reuses the existing quote-aware tag scanner for Atom `<entry>` tags instead of maintaining a second naive scanner.
**No-bloat:** ✓ The refactor ledger is complete. Frontmatter and whitespace notes were deferred because generated/committed snapshots already round-trip and normalization is deliberately symmetric. The fenced-diff concern is already pinned by a test proving every diff line is prefixed, so upstream backticks cannot close the fence at line start.
**Wiring:** ✓ `packages/cli/tests/upstream-monitor/wiring.test.ts` uses real source definitions, real committed snapshots, the real parser, and real file reads; only network and GitHub issue mutation are boundary fakes.

**Verdict:** APPROVE

**Critical issues:** None.

**Changes pushed:**

- `ef4a1539c` — tolerate invalid HTML numeric character references without disabling the Cursor watch; includes out-of-range hex/decimal and surrogate coverage.
- `2d399c65c` — reuse the quote-aware tag scanner for Atom entry tags.

**Independent review:** The bounded latest-update review completed cross-agent (`author_agent=codex`, `actual_reviewer=claude`, `independence=cross-agent`) and approved. The first post-fix attempt timed out safely; its exact retry then completed cross-agent and approved the corrected source plus source/wiring tests.

**Verification:** 432/432 Vitest files pass (6,573 tests, 5 intentional skips); 910/913 acceptance scenarios pass with 3 intentional skips (34,443 passing steps); full lint/format/typecheck clean; diff audit reports no dependency violations or config drift; `bun audit` reports no vulnerabilities; pre-push schema drift gate passes 753/753 tests.

**Provenance:**

- verified: https://github.com/openai/codex/issues/18115
- verified: https://registry.npmjs.org/@openai/codex/latest
- verified: https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues
- verified: https://html.spec.whatwg.org/multipage/parsing.html#parse-error-character-reference-outside-unicode-range
- verified: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint#exceptions

**Next:** Human reviewer can confirm and merge once the fresh CI run at `2d399c65c` is green.

TheMostlyGreat (commented):
## Quality Review — updated head

**Currency:** ✓ Current as of 2026-08-05. `openai/codex#18115` remains open; npm reports `@openai/codex` 0.146.0 as latest while the tripwire deliberately pins 0.145.0 as the last known-bad version.
**Sources:** ✓ GitHub issue/API behavior, supported API versions, rate limits, npm state, and WHATWG numeric-character-reference behavior were checked against current primary sources.
**Correct:** ✓ No known code defect remains. The branch now incorporates current `main`; the conflict resolution preserves both fail-closed Git subprocess handling and merge-history search for the manifest commit whose inputs match the seal.
**Elegant:** ✓ `608c67e33` removes the test-only dead parser created by the merge and makes production consume the candidate parser its unit tests exercise.
**No-bloat:** ✓ The refactor ledger is complete; no other cleanup justified widening the diff.
**Wiring:** ✓ `packages/cli/tests/upstream-monitor/wiring.test.ts` uses real source definitions, committed snapshots, real parsing, and real filesystem reads; only network and GitHub issue mutation are boundary fakes.

**Verdict:** NEEDS DISCUSSION

**Critical issues:** None.

**Suggested improvements:** None in the reviewed code. A human independent confirmation remains required because the coordinator could not produce independent evidence.

**Changes pushed:**

- `e5936bbbf` — merge current `main`, resolve the sealed-review test composition, and regenerate Claude plugin identity/inventory at 0.73.0.
- `608c67e33` — keep sealed-commit candidate parsing on the production path instead of testing dead code.

**Verification:** 436/436 Vitest files pass (6,598 tests, 5 intentional skips); 41 merge-focused tests and 11 refactor-focused tests pass; ESLint, Prettier, typecheck, plugin release contract, dependency architecture (275 modules / 395 dependencies), config drift, and pre-push schema gate (753/753) are clean. The diff-scoped audit deliberately skips repository-wide Knip, clone, and outdated-dependency discovery.

**Independent review limitation:** The prescribed full-packet attempt, exact retry, and a narrowed post-refactor attempt all failed closed with `assigned_reviewer=claude`, `preferred_failure=timed_out`, fallback `invalid_output`, and `independence=none`. I am not substituting a self-approval.

**Provenance:**

- verified: https://github.com/openai/codex/issues/18115
- verified: https://registry.npmjs.org/@openai/codex/latest
- verified: https://docs.github.com/en/rest/issues/issues#list-repository-issues
- verified: https://docs.github.com/en/rest/about-the-rest-api/api-versions
- verified: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api
- verified: https://html.spec.whatwg.org/multipage/parsing.html#numeric-character-reference-end-state
- verified: https://git-scm.com/docs/git-merge
- verified: https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/resolving-a-merge-conflict-using-the-command-line

**Next:** Let CI finish at `608c67e33`, then have a human reviewer provide the independent confirmation the coordinator could not produce.

PR #1972, opened 2026-08-05
Give new users one Safe Word install command

TheMostlyGreat:
## What changed

- Make `safeword install` configure the project and install Claude Code and Codex support by default.
- Keep Cursor opt-in through `--agents=cursor` or an explicit agent list.
- Retain `setup`, `upgrade`, `claude install`, and `codex install` indefinitely as deprecated compatibility aliases that route through the canonical lifecycle.
- Align status, plans, recovery guidance, templates, the website, packaged plugin documentation, and test fixtures with the unified lifecycle.
- Preserve scoped installs such as `safeword install --agents=claude`, `--agents=codex`, and `--agents=none` for deliberate host or project-only workflows.

## Why

First-time users were taught multiple overlapping commands: configure the project with `setup`, then install individual host integrations. The CLI also defaulted to Cursor-oriented project assets even when users only needed Claude Code and Codex. This made installation harder to explain and allowed documentation, status checks, and recovery actions to disagree about the canonical path.

The unified lifecycle gives users one default command while retaining every existing alias for compatibility. Cursor remains untouched unless explicitly selected.

## Validation

- Full non-live BDD suite: 375 scenarios, 5,438 steps passed.
- Full Vitest suite: 429 files passed; 6,550 tests passed and 5 skipped.
- TypeScript typecheck passed.
- ESLint and Gherkin lint passed.
- Schema-drift push gate: 753 tests passed.
- Claude plugin release contract aligned at 0.72.0.

The push gate also reported a non-blocking historical phase-provenance warning for the local ZE5RRG ticket ledger; it did not indicate a code, schema, or test failure.

Closes #1925


packages/cli/src/lifecycle/status.ts
```diff
@@ -79,12 +83,98 @@ function projectVersionFinding(
...
+
+  if (agents.includes('claude')) {
+    const { observeClaudeStatus } = await import('../claude-plugin/status.js');
+    surfaces.push({ name: 'claude', result: observeClaudeStatus(cwd) });
+  }
+  if (agents.includes('codex')) {
+    const { observeCodexMigration } = await import('../codex-plugin/installer.js');
+    surfaces.push({ name: 'codex', result: observeCodexMigration(cwd, environment) });
+  }
+  if (agents.includes('cursor')) {
```
TheMostlyGreat, line 118: **Must fix (fixed on branch): the Cursor surface is synthesised, not observed.**

`createResult({ state: project.state, ... })` copies the project verdict, so nothing about Cursor is ever measured — a Cursor-only gap renders as `Cursor: ready`. That contradicts NTB1.R1 ("results identify each surface's outcome"), and the `coverage: 'project-owned Cursor assets'` label asserts a coverage claim the code doesn't make.

Replaced with `observeCursorProject`, which reads back the Safe Word-owned Cursor files the schema declares. Same fix applied to the install path in `lifecycle/commands.ts`.

packages/cli/src/schema.ts
```diff
@@ -1425,3 +1425,49 @@ export const SAFEWORD_SCHEMA: SafewordSchema = {
...
+    path.startsWith('.safeword/skills/') ||
+    CURSOR_PROJECT_PATHS.has(path)
+  );
+}
+
+function withoutCursorEntries<T>(values: Record<string, T>): Record<string, T> {
+  return Object.fromEntries(Object.entries(values).filter(([path]) => !isCursorProjectPath(path)));
+}
+
+/** Select project-owned surfaces without scattering Cursor path filters through commands. */
```
TheMostlyGreat, line 1530: **Must fix (fixed on branch): existing Cursor users silently freeze on upgrade.**

This filter is right, but combined with `--agents` not being persisted it opens a hole: a project that installed Cursor assets gets the `claude,codex` default on its next plain `safeword install`, every Cursor path is filtered out, and those files are then **neither upgraded nor reported**. No finding, no next action, no drift signal.

I read "Cursor stays untouched unless selected" as intentional — but *untouched and unmentioned* is the failure mode. Status now emits `CURSOR_NOT_SELECTED` (info) naming `--agents=cursor` when the project carries Cursor assets the current selection excludes. `isCursorProjectPath` is exported so `lifecycle/cursor.ts` can reuse this exact predicate rather than duplicating the path list.

TheMostlyGreat (commented):
Reviewer: inline findings from `/review-spec`, `/quality-review`, and `/refactor`. All of these are fixed in the commits pushed to this branch — anchored here so the reasoning sits next to the code. Posted as COMMENT (gh authenticates as the PR author, so this carries no merge block).

packages/cli/src/cli-protocol/result.ts
```diff
@@ -78,6 +78,16 @@ const EMPTY_EFFECTS: Effects = {
   destructive: [],
 };
 
+export function combineEffects(groups: readonly Partial<Effects>[]): Effects {
```
TheMostlyGreat, line 90: **Fixed on branch:** `Object.fromEntries(...) as unknown as Effects` — a double cast that turns off the checker on the one function whose whole job is shape correctness. It was also a near-duplicate of `mergeEffects` in `project-install.ts`, which built the same five categories from the same literal list.

Now an explicit typed literal, with `mergeEffects` reusing it and applying its dedupe on top.

packages/cli/src/cli-protocol/result.ts
```diff
@@ -100,10 +110,27 @@ export function withDeprecation(
   replacement: string,
```
TheMostlyGreat, line 127: **Not a defect — flagging for the release notes.**

Swapping `retainedThrough`/`removalEligibleAfter` for `retention: 'indefinite'` is the right model, but it removes `retained_through` and `removal_eligible_after` from published JSON findings metadata. That's a machine-contract change for anyone parsing deprecation metadata.

Pre-1.0, so a 0.x minor is consistent with existing practice. It just isn't mentioned in the PR description, which leads with compatibility.

packages/cli/src/lifecycle/commands.ts
```diff
@@ -0,0 +1,467 @@
...
+    uninstalling ? uninstallOperation : 'upgrade',
+    projectSchema,
+  );
+  const agentEffects = uninstalling ? agentUninstallEffects : agentInstallEffects;
+  const surfaces = [
+    { name: 'project', effects: project.plan.effects },
+    ...agents.map(agent => ({ name: agent, effects: agentEffects(agent) })),
+  ];
+  const observations = await profilePreconditions(cwd, agents);
+  const preconditionDigest = createHash('sha256')
```
TheMostlyGreat, line 377: **Fixed on branch:** `prepareInstall` returned `PreparedUninstall`.

Beyond the name, the two functions differed only in reconciliation operation, per-agent effects, and `requiresConfirmation` — the ~20 lines of digest construction and result assembly were duplicated verbatim. Folded into `prepareLifecycle(cwd, operation, agents, full)` and renamed the shape `PreparedLifecycle`.

The cursor branch also used `createResult({ state: 'healthy' }).effects` to obtain an empty `Effects` — now a named `EMPTY_SURFACE_EFFECTS` constant.

packages/cli/src/cli-protocol/public-handlers.ts
```diff
@@ -69,23 +78,24 @@ function configCheckResult(inspection: ConfigInspection): CliResult {
...
-        : undefined,
-    progress: invocation.progress,
+async function doctorHandler(invocation: CommandInvocation): Promise<CliResult> {
+  const parsed = parseAgentSelection(invocation.options.agents);
+  if (!parsed.ok) return invalidAgentSelection('doctor', parsed.error);
+  const { diagnoseLifecycle } = await import('../lifecycle/doctor.js');
+  return diagnoseLifecycle(invocation.cwd, parsed.selection.agents);
+}
+
+async function installHandler(invocation: CommandInvocation): Promise<CliResult> {
```
TheMostlyGreat, line 94: **Fixed on branch: `setupHandler` is unreachable.**

`alias('setup', 'install')` in `catalog.ts` sets `handler: canonical.handler`, so the `setup` route dispatches `installHandler`. Nothing ever calls `publicHandler('setup')`, making this function and its `HANDLERS.setup` entry dead — and misleading, since it implies `setup` still runs project-only convergence with its own offline gate.

Also removed in the same pass: `src/commands/plan.ts` (`observePlan`, zero importers after `planHandler` moved to `planLifecycle`) and the four one-line re-export shims this file imported through (`commands/{converge-setup,status,doctor,lifecycle}.ts`).

packages/cli/src/lifecycle/doctor.ts
```diff
@@ -0,0 +1,68 @@
...
+    })),
+  ]);
+}
+
+export async function diagnoseLifecycle(
+  cwd: string,
+  agents: readonly AgentIntegration[],
+  environment: NodeJS.ProcessEnv = process.env,
+): Promise<CliResult> {
+  const surfaces = await observeLifecycleSurfaces(cwd, agents, environment);
```
TheMostlyGreat, line 46: **Fixed on branch:** `(summary.data as { surfaces: readonly object[] }).surfaces` casts blindly into another module's payload — if `summarizeLifecycleStatus` changes shape, this breaks silently at runtime with no type error.

Both commands now build surfaces from a shared, typed `lifecycleSurfaceSummaries()` in `status.ts`. `diagnostics()` also returned `readonly object[]`; it now has a named `LifecycleDiagnostic` interface.

packages/cli/features/steps/unified-first-time-install.steps.ts
```diff
@@ -0,0 +1,2239 @@
...
+    createLegacyGuidanceFixture(this);
+  },
+);
+
+When('its catalogue and human plan are rendered', function (this: UnifiedInstallWorld) {
+  runRawCommand(this, ['codex', 'clean-guidance'], false);
+});
+
+Then(
+  'both call the operation destructive deactivation rather than only a backup',
```
TheMostlyGreat, line 1473: **Must fix (fixed on branch): the architecture Rule TBU2.R4 executed nothing.**

The `When` stored the flag string and the `Then` re-derived the expected value from that same string:

```ts
this.architectureFlags = flags;                                              // When
assert.equal(flags.includes('--from-index') ? 'index' : 'worktree', input);  // Then
```

That's a tautology over the Examples table — it passes with the architecture command deleted. Worst case was the scenario named *"differential-tested against legacy behavior"*, whose `Then` was `assert.equal(normalizedLegacy, canonical)`: no fixture, no run, no comparison. Three scenarios, eight rows.

They now build a git fixture whose worktree and index describe different module graphs (`shipping` worktree-only, `billing` index-only), run the real CLI, and assert generated content plus actual `git diff --cached` staging. The differential scenario runs each spelling in its own fresh fixture, and guards its own vacuity: the two rows must differ in staging, so a constant implementation fails.

packages/cli/features/steps/unified-first-time-install.steps.ts
```diff
@@ -0,0 +1,2239 @@
...
+  },
+);
+
+Given(
+  'setup is a retained compatibility route for non-destructive install',
+  function (this: UnifiedInstallWorld) {
+    initializeHosts(this);
+  },
+);
+
```
TheMostlyGreat, line 1074: **Must fix (fixed on branch): the alias outline invoked 3 of its 25 rows.**

Rule TBU2.R6 claims *"Every shipped alias remains executable"*, but the `When` ran only `bare safeword`, `claude install`, and `codex install`. The other 22 rows asserted catalogue metadata only — never proving the spelling still dispatches, which is the entire claim.

Every command alias now runs for real and must report its canonical route via the `CLI_ALIAS_DEPRECATED` finding. The two option aliases can't be invoked standalone, so they name the suite that proves them instead of passing silently.

**Product observation, not fixed — your call:** `codify` and `connect` needed operands supplied, because Commander's required-argument check runs *before* the handler. So a user typing a retained alias with a missing argument gets `missing required argument` and **no compatibility guidance at all** — exactly the user this Rule exists for. Worth a follow-up issue.

packages/cli/features/steps/unified-first-time-install.steps.ts
```diff
@@ -0,0 +1,2239 @@
...
+    stdout: completed.stdout,
+    stderr: completed.stderr,
+    exitCode: completed.status ?? 1,
+  };
+}
+
+/**
+ * A lifecycle command completed without failing. Exit 0 is healthy or changed;
+ * exit 2 is `action_required` — a real outcome (a plan awaiting confirmation, a
+ * pending host activation), not an error. Exit 1 and anything else are failures.
```
TheMostlyGreat, line 454: **Must fix (fixed on branch): this helper asserted only `retention`.**

The original built its expected object out of the actual one:

```ts
assert.deepEqual(world.compatibilityRoute, {
  route: world.compatibilityRoute?.route === alias ? alias : world.compatibilityRoute?.route,
  ...
});
```

Both ternaries self-cancel, so the only real assertion was `retention: 'indefinite'` — which the sibling `metadata schedules no deletion date` step already made. The alias↔canonical binding was actually enforced by the `Given`'s `.find()`, not here; the helper read as an assertion and proved nothing.

Now asserts route, replacement, and retention directly, with messages naming the alias.

packages/cli/features/steps/unified-first-time-install.steps.ts
```diff
@@ -0,0 +1,2239 @@
...
+    const statusData = this.statusEnvelope?.data as Record<string, unknown> | undefined;
+    const doctorData = this.doctorEnvelope?.data as Record<string, unknown> | undefined;
+    assert.equal(statusData?.diagnostics, undefined);
+    assert.equal(statusData?.coverage, undefined);
+    assert.ok(Array.isArray(doctorData?.diagnostics));
+    assert.ok(Array.isArray(doctorData?.coverage));
+  },
+);
+
+Given('the public command catalogue and handlers', () => {
```
TheMostlyGreat, line 681: **Must fix (fixed on branch): title/behaviour mismatch.**

The scenario is *"The catalogue forbids status and doctor from sharing one handler contract"*, but the `Given` was an empty comment, the catalogue was never inspected, and the `Then` was `assert.notDeepEqual(statusEnvelope, doctorEnvelope)` — which passes if doctor is literally status with a different `command` label. The Rule it proves is exactly the one it fails to check.

The `Given` now asserts distinct handlers and fixtures in `commandCatalog`; the `Then` asserts doctor carries `diagnostics` and `coverage` and status carries neither.

packages/cli/features/steps/unified-first-time-install.steps.ts
```diff
@@ -0,0 +1,2239 @@
...
+    stderr: completed.stderr,
+    exitCode: completed.status ?? 1,
+  };
+}
+
+/**
+ * A lifecycle command completed without failing. Exit 0 is healthy or changed;
+ * exit 2 is `action_required` — a real outcome (a plan awaiting confirmation, a
+ * pending host activation), not an error. Exit 1 and anything else are failures.
+ */
```
TheMostlyGreat, line 455: **Fixed on branch: 18 success paths asserted `notEqual(exitCode, 1)`.**

Exit 2 is a real code in this CLI — the offline install refusal and the stale-plan refusal both use it — so a refusal satisfied a success assertion. Two scenarios asserting opposite outcomes could both pass on the same exit code.

Resolved per site rather than by a blanket rule, because the distinction is genuine: most now assert `0`, while the sites where `action_required` is the correct outcome (a default install with host activation pending, a plan that declares effects, status/doctor observations) use the named `assertCommandDidNotFail` helper that accepts 0 or 2 and rejects everything else. Getting this wrong in the first pass is what surfaced the distinction — the blanket `=== 0` broke 25 scenarios that were legitimately action-required.

packages/cli/features/steps/unified-first-time-install.steps.ts
```diff
@@ -0,0 +1,2239 @@
...
+
+Then(
+  'core project configuration and Safe Word-owned Cursor assets are installed',
+  function (this: UnifiedInstallWorld) {
+    const project = requiredPath(this.projectRoot, 'project root');
+    assert.equal(existsSync(nodePath.join(project, '.safeword/SAFEWORD.md')), true);
+    assert.equal(existsSync(nodePath.join(project, '.cursor')), true);
+  },
+);
+
```
TheMostlyGreat, line 1029: **Must fix (fixed on branch): only the preservation half of the claim was asserted.**

The scenario is *"Safe Word Cursor entries are **reconciled** without replacing unrelated content"*, but the step checked only `customer.json` and `third-party.json` — both untouched files. An install that does nothing at all to `.cursor/` passes.

Now also requires the schema-declared Safe Word-owned Cursor files to exist, so the positive half is proven too.

packages/cli/features/steps/unified-first-time-install.steps.ts
```diff
@@ -0,0 +1,2239 @@
...
+    }
+    runInstall(this, ['--agents', 'none']);
+    assert.equal(this.result.exitCode, 0, this.result.stderr || this.result.stdout);
+    this.fixtureBefore = fixtureEffectDigest(this);
+    runRawCommand(this, [alias]);
+  },
+);
+
+Then(
+  'the observable contract remains {string}',
```
TheMostlyGreat, line 1329: **Should strengthen (fixed on branch):** the `When` spawned `--help` and `capabilities` into `this.referenceHelp` / `this.referenceCapabilities`, then the `Then` ignored both and asserted `commandCatalog` instead.

So *"CLI reference and capability fixtures are validated"* validated neither: the reference could drop a command entirely and the scenario would stay green. The `Then` now asserts against the gathered help text and the capabilities payload.

Also in this pass: `mixedInstallSummary` was set by a `Given` and never read anywhere.

PR #1991, opened 2026-08-05
Prove user data survives a Codex handoff that succeeds (#1989)

TheMostlyGreat:
Fixes #1989.

## Gap

All three `TB1.R4` acceptance scenarios covered a handoff that **declines** — non-vacuously, since #1973. Nothing covered user-owned tickets, learnings, or authored `.agents/skills` surviving a handoff that **succeeds**. That is the path where files actually move and get rewritten, so it is where survival is least obvious.

## Change

Two proof lanes, no production change:

- The focused Vitest regression exercises the real automatic-migration and transactional-finalization collaborators against a temporary filesystem.
- The executable acceptance scenario drives the packaged CLI through a successful handoff, proving the plugin install crosses the fake `codex` process boundary, authored project data remains byte-identical, and managed fallback assets are removed.
- A shared hermetic fake Codex runtime isolates `CODEX_HOME` and fakes only the external `codex`/`bun` boundary; no network or ambient profile is used.
- The feature, spec, ticket done criteria, and scenario ledger now agree on the successful-handoff behavior.

## On the marketplace-source override

#1989 proposed threading a `marketplaceSource` override through `automaticallyMigrateLegacyCodex` so the acceptance lane could point at a local marketplace, and flagged that it was worth confirming production actually needs it. It does not:

- The fake-runtime harness reaches the success path hermetically, which is what the issue requires.
- Adding an override for *where plugin code is fetched from* widens the supply-chain surface for a test-only benefit. This repo already refuses repo-local reviewer binaries on the same reasoning, so a new redirect seam would cut against that.

Real-`codex` enrollment remains covered by `TB1.R1`/`R2`, which drive the actual binary against a local marketplace.

## Verification

Falsified in both directions before trusting the proof:

- Treating a managed file as user-owned fails the byte-identity assertion, proving the preservation check has teeth.
- Blocking plugin installation fails the successful-handoff scenario, so it cannot pass vacuously.
- Focused migration Vitest: **77/77 passed**.
- Successful-handoff acceptance: **1 scenario, 44/44 steps passed**.
- Full CLI Vitest: **6,749 passed, 5 skipped**; relay: **167 passed, 1 skipped**.
- Build, ESLint, TypeScript, diff hygiene, dependency audit, and `bun audit`: clean.
- The full acceptance run reached **1,077 passed / 3 skipped** and reproduced one base-branch failure: the public-command determinism scenario compares two generated `codex status` `recorded_at` timestamps. #1991 does not change that scenario or implementation; the focused rerun reproduces it independently of this PR's migration path.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UPpTdcLAj7JkXzSpigWk1v)_

### DIFF
--- a/features/test-codex-plugin-migration.feature
+++ b/features/test-codex-plugin-migration.feature
@@ -97,7 +97,7 @@ Feature: Test Codex plugin migration
       And each command invokes a packaged Safe Word command entrypoint
 
   @test-codex-plugin-migration.TB1.R4 @surface.openai-codex
-  Rule: test-codex-plugin-migration.TB1.R4 — Upgrading an old project-local Codex install leaves user-owned project data and working fallback assets intact until an explicit proven handoff
+  Rule: test-codex-plugin-migration.TB1.R4 — Upgrading an old project-local Codex install leaves user-owned project data and fallback assets intact until an explicit proven handoff
 
     @rejection
     Scenario: Unavailable profile enrollment preserves an old project's authored data and fallback
--- a/features/test-codex-plugin-migration.feature
+++ b/features/test-codex-plugin-migration.feature
@@ -109,6 +109,17 @@ Feature: Test Codex plugin migration
       And Safe Word keeps repo-local `.agents/skills` available during the compatibility window
       And legacy Safe Word Codex hook runtime files remain until explicit handoff cleanup
 
+    Scenario: Successful handoff preserves authored project data while removing managed fallback
+      Given a repo installed with today's project-local Codex assets
+      And the repo contains user-owned tickets and learnings under the namespace root
+      And the repo contains a user-authored `.agents/skills/company-workflow/SKILL.md`
+      When the plugin migration handoff succeeds under an isolated Codex profile
+      Then the handoff installs the Safe Word plugin through the fake Codex boundary
+      And the user-owned tickets and learnings remain byte-identical
+      And the user-authored skill remains byte-identical
+      And Safe Word-owned Codex skill files are removed after finalization
+      And legacy Safe Word Codex hook runtime files are removed after finalization
+
     Scenario: User-authored Codex skills survive the migration
       Given an old project-local Codex install with a user-authored `.agents/skills/company-workflow/SKILL.md`
       When the plugin migration upgrade runs without profile enrollment available
--- a/features/test-codex-plugin-migration.feature
+++ b/features/test-codex-plugin-migration.feature
@@ -113,8 +113,8 @@ Feature: Test Codex plugin migration
       Given a repo installed with today's project-local Codex assets
       And the repo contains user-owned tickets and learnings under the namespace root
       And the repo contains a user-authored `.agents/skills/company-workflow/SKILL.md`
-      When the plugin migration handoff succeeds under an isolated Codex profile
-      Then the handoff installs the Safe Word plugin through the fake Codex boundary
+      When the plugin migration upgrade runs with profile enrollment available
+      Then Safe Word is enrolled in the Codex profile
       And the user-owned tickets and learnings remain byte-identical
       And the user-authored skill remains byte-identical
       And Safe Word-owned Codex skill files are removed after finalization
--- a/packages/cli/tests/commands/migrate-codex-plugin.test.ts
+++ b/packages/cli/tests/commands/migrate-codex-plugin.test.ts
@@ -88,12 +88,11 @@ describe('migrate codex-plugin command', () => {
     const configPath = nodePath.join(directory, '.codex/config.toml');
     writeFileSync(configPath, config);
 
-    const runtime = installFakeCodexRuntime(
-      directory,
+    const runtime = installFakeCodexRuntime(directory, {
       pluginEnabled,
       pluginInitiallyInstalled,
       pluginVersion,
-    );
+    });
     return {
       directory,
       configPath,
--- a/packages/cli/tests/commands/migrate-codex-plugin.test.ts
+++ b/packages/cli/tests/commands/migrate-codex-plugin.test.ts
@@ -346,7 +346,7 @@ describe('migrate codex-plugin command', () => {
     const fixture = createMigrationFixture(LEGACY_HOOK_CONFIG, true, false);
     const userOwned = {
       '.project/tickets/ABC123/spec.md': '# Authored spec\n\nOwned by the user, not Safeword.\n',
-      '.project/learnings/INDEX.md': '# Learnings\n\n- Do not clobber authored notes.\n',
+      '.project/learnings/handoff-notes.md': '# Handoff notes\n\nDo not clobber authored notes.\n',
       '.agents/skills/company-workflow/SKILL.md': 'User-authored company workflow skill\n',
     };
     for (const [relativePath, contents] of Object.entries(userOwned)) {
--- a/packages/cli/tests/commands/migrate-codex-plugin.test.ts
+++ b/packages/cli/tests/commands/migrate-codex-plugin.test.ts
@@ -1,7 +1,6 @@
 /* eslint-disable unicorn/no-null -- migration JSON uses null for unavailable profile facts */
 
 import {
-  chmodSync,
   existsSync,
   lstatSync,
   mkdirSync,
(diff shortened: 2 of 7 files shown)


TheMostlyGreat (commented):
Quality + refactor follow-up complete on `bdbe73aa4`.

Changes pushed:
- `0d4284d94` makes the new successful-handoff regression prove the fake Codex process actually received `plugin add safeword@safeword --json`; cleanup alone can no longer satisfy the central success claim.
- `bdbe73aa4` fixes the fake Codex logger to delimit invocations with real newlines, removing the latent ambiguity identified during independent review.

Review result:
- No blocking correctness, security, architecture, or documentation findings.
- The final independent cross-agent quality review approved the updated file.
- Refactor scout is clean: the existing environment-helper extraction is the useful simplification; further extraction would reduce readability without meaningful duplication savings.
- The preview/apply environment asymmetry in the local test helpers is intentional and load-bearing for the drift tests, so I left it unchanged.

Verification:
- `migrate-codex-plugin.test.ts`: 77/77 passing after each change
- ESLint + Prettier: passing on the changed file and in pre-commit
- Full TypeScript check, dependency-boundary audit, learning/domain-doc checks: passing

CI is refreshing against the pushed head; I will watch it through completion.

TheMostlyGreat (commented):
Reviewer: ran `/review-spec`, `/quality-review`, and `/refactor` over this branch. **Verdict: the change is sound and both new tests have teeth.** I pushed three commits — two refactors and one fixture fix — and left two findings inline for you.

## What I verified, not just read

- **`bun run test:bdd features/test-codex-plugin-migration.feature`** — 23 scenarios, 925 steps, green.
- **`bun run test tests/commands/migrate-codex-plugin.test.ts`** — 77 passed.
- **`bun run lint`** (eslint + lint-gherkin + `tsc --noEmit`) — clean.
- **Falsified the new acceptance scenario independently.** Injected `SAFEWORD_FAIL_PLUGIN_INSTALL=1` into the successful-handoff step and re-ran: it fails with `Codex native plugin handoff could not complete… marketplace unavailable` / `2 !== 0`. So the scenario cannot pass without a handoff that actually succeeded. Reverted after.
- **Checked the removal assertions aren't vacuous.** `createProjectLocalCodexInstallFixture` really writes all four files the scenario later asserts gone (`.agents/skills/{bdd,verify}/SKILL.md`, `.safeword/hooks/codex/{pre-tool-quality,stop}.ts`), so `existsSync(...) === false` is a real transition, not an assertion about a file that never existed.
- **Checked the new cross-package import resolves under both cucumber entry paths.** `steps/` now imports `packages/cli/tests/helpers/fake-codex-runtime.js`, and `test:bdd:live` loads step files through `tsx/esm` rather than `--import tsx`. Resolved fine under both — no live-profile breakage.

I agree with the call to drop the `marketplaceSource` override. The fake-runtime harness reaches the success path hermetically, so the override would have bought a test-only seam in the one place — where plugin code is fetched from — that least deserves a new redirect.

## Changes I pushed

| Commit                                                             | What                                                                                                                                                                                                                                                                                                                                                        |
| ------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `refactor: source Codex fixture paths from the fake runtime`       | `installFakeCodexRuntime` already returns `codexHome` and `logPath`, but `createMigrationFixture` discarded both and ~40 call sites re-derived the same paths from the `'profile'` and `'codex.log'` string literals. Extracting the helper was supposed to end that duplication; now it does. Net −11 lines.                                               |
| `test: emit a real newline in the fake Codex config mutation`      | The concurrent-config-update branch printed `'# concurrent config update\\n'` — under `String.raw` that's a literal backslash-n, so the simulated mutation appended an unterminated line to the TOML. This is the same bug your `test: delimit fake Codex command log entries` commit fixed for the log line; the sibling branch three lines above kept it. |
| `refactor(bdd): describe the successful handoff in behavior terms` | See the inline comment on the feature file.                                                                                                                                                                                                                                                                                                                 |

## Findings left for you

Both are inline. In short: the positional-boolean signature of `createMigrationFixture` has one argument that's dead at the new call site, and the acceptance rule drives `safeword upgrade`, which the CLI itself reports as deprecated.

## One thing worth naming, not fixing

The new behavior is now asserted twice — the vitest test calls `automaticallyMigrateLegacyCodex` directly, the cucumber scenario drives the real CLI as a subprocess. Repo guidance says not to duplicate a vitest integration test as a cucumber step just to feed the lane. I think this case is defensible: the two cross different boundaries, and the subprocess one is what proves the handoff works through the actual entrypoint. Flagging it so the shape doesn't get copied where the second lane would add nothing.

---

_Reviewed with [Claude Code](https://claude.com/claude-code)_

packages/cli/tests/commands/migrate-codex-plugin.test.ts
```diff
@@ -339,13 +247,46 @@ describe('migrate codex-plugin command', () => {
     ).toBe(true);
   });
 
+  // The declining handoff is covered by the acceptance lane's TB1.R4 rule. The
+  // succeeding one is where files actually move and get rewritten, so it is the
+  // path where survival is least obvious and most worth pinning.
+  it('preserves user-owned project data and authored skills through a successful handoff', () => {
+    const fixture = createMigrationFixture(LEGACY_HOOK_CONFIG, true, false);
```
TheMostlyGreat: Reviewer: `createMigrationFixture(LEGACY_HOOK_CONFIG, true, false)` — three positional booleans plus an optional version string, across ~60 call sites in this file. At this call site `pluginEnabled: true` is dead: the fake runtime only consults it when `pluginInitiallyInstalled` is true, which it isn't here. So the literal reads as a meaningful choice and isn't one.

An options object (`{ config, pluginEnabled, pluginInitiallyInstalled, pluginVersion }`) would make every call site self-describing. Not applied here — it's ~60 mechanical edits and doesn't belong in a review pass — but worth a follow-up.

features/test-codex-plugin-migration.feature
```diff
@@ -109,6 +109,17 @@ Feature: Test Codex plugin migration
       And Safe Word keeps repo-local `.agents/skills` available during the compatibility window
       And legacy Safe Word Codex hook runtime files remain until explicit handoff cleanup
 
+    Scenario: Successful handoff preserves authored project data while removing managed fallback
+      Given a repo installed with today's project-local Codex assets
+      And the repo contains user-owned tickets and learnings under the namespace root
+      And the repo contains a user-authored `.agents/skills/company-workflow/SKILL.md`
+      When the plugin migration upgrade runs with profile enrollment available
```
TheMostlyGreat, line 116: Reviewer: two things about this step, one fixed and one left for you.

**Fixed (pushed):** it read `the plugin migration handoff succeeds under an isolated Codex profile`, and the Then read `the handoff installs the Safe Word plugin through the fake Codex boundary`. Both named the harness rather than the behavior — "fake" and "isolated Codex profile" are facts about the test, not about what a user gets. The When now mirrors the rejection scenarios' own phrasing and the Then states the outcome. Assertions unchanged.

**Left for you:** the step drives `safeword upgrade`, and the CLI answers `\`upgrade\` is deprecated; use \`setup\``. That's pre-existing across the whole TB1.R4 rule rather than new here, but this scenario extends the dependence on a command the product is retiring. Worth a follow-up to move the rule onto `setup`.

packages/cli/tests/commands/migrate-codex-plugin.test.ts
TheMostlyGreat: Addressed in `acb8e3f21`. `createMigrationFixture` now takes a named options object with defaults, all non-default call sites are self-describing, and the semantically dead `pluginEnabled: true` values were removed when no plugin is installed. Revalidated on the latest-main merge: migration Vitest 77/77.

features/test-codex-plugin-migration.feature
TheMostlyGreat: Confirmed and intentionally left out of the behavior-preserving refactor: changing this rule from `upgrade` to `setup` changes the acceptance contract across all TB1.R4 scenarios. The handoff is already ticketed as #2062, **Move Codex migration acceptance coverage off the deprecated `upgrade` alias**: https://github.com/ArcadeAI/safeword/issues/2062. It is open and currently unassigned.

PR #2016, opened 2026-08-06
Migrate legacy Claude projects automatically

TheMostlyGreat:
## Summary

- automatically contract exact Safeword-owned Claude assets from supported historical releases after the native plugin proves a successful project-bound `UserPromptSubmit`
- preserve modified, unknown, shared, Cursor, Codex, and project-authored state; report one loud but non-blocking repair message when attention remains
- make cleanup crash-safe, race-safe, deadline-aware, and idempotent across prompts, sessions, developers, and worktrees
- support both project- and user-scoped Claude plugin enrollment while keeping project declarations available for the next teammate
- keep Codex teammate bootstrap continuity and harden generated plugin inventory/runtime integrity
- add BDD, unit/integration coverage, release-contract generation, maintainer runbooks, and user documentation

## Why

Projects created before native plugin delivery can still contain Safeword-owned Claude commands, skills, hooks, and settings from 0.68, 0.69, or 0.72. Native plugin installation alone does not prove those legacy bytes are safe to remove, and the previous current-template classifier could misclassify historical assets. That left duplicate behavior and support-heavy manual cleanup.

This change makes removal proof-gated and catalogue-driven: Safeword only removes exact known historical content after the exact native plugin has executed in that repository. Anything changed or unrecognized stays untouched and is named in a plain-language advisory.

## Verification

- lint, Gherkin lint, typecheck, formatting, build, historical-catalogue generation/check, plugin release contract, dependency audit, and repository audit completed
- push schema-drift gate: 753/753 passed
- Retro Relay unit suite: 167 passed, 1 skipped
- focused rechecks after rebase: 12/12 passed
- affected Retro Relay and Codex migration features in isolation: 72/72 scenarios and 2,872/2,872 steps passed

### Current-main verification caveat

The all-in-one rebased verification lane reached 6,795 passing tests with 5 skips, but initially reported two failures: a real Codex dependency-pin merge issue (fixed and rechecked) and a runner-lock partial-read race that passed its focused recheck. Its BDD phase reached 1,092 passing scenarios with 3 skips and 12 Retro Relay failures caused by overloaded run/port-lock contention; those affected features subsequently passed cleanly in isolation. This PR does not claim that the combined overloaded lane was fully green.

## Before release

Run the authenticated disposable-profile Claude RC lifecycle, mid-session reload and host-kill observations, plus the five-run cold historical-fixture benchmark. No release or publish action is included here.


steps/automatic-claude-migration.steps.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (match === null) continue;
+  const [, keyword, text] = match;
+  if (keyword === undefined || text === undefined || registered.has(text)) continue;
+  registered.add(text);
+  if (keyword === 'When') {
+    When(expression(text), function (this: MigrationWorld) {
+      this.automaticClaudeMigrationVerification = verifyAutomaticClaudeMigration();
+    });
+  } else if (keyword === 'Then') {
+    Then(expression(text), function (this: MigrationWorld) {
```
TheMostlyGreat: Reviewer (must-fix, finding #1): every `Then` in this feature — all 24 scenarios — resolves to this one assertion: the cached exit code of a nine-file vitest run. "the symlink and external file remain byte-for-byte unchanged", "exactly one transaction is created", and "the completed transaction is removed" are the same assertion here.

Because the steps are generated from the feature text at import time, no scenario in this file can ever be undefined or red on its own — rewording one auto-registers a new passing step. `test:bdd` is a blocking CI step, so this makes the gate a rubber stamp for the riskiest one-way deletion path in the product.

Every other steps file in the repo builds real fixtures; `steps/native-claude-plugin.steps.ts` is the model. This is also the anti-pattern 4495e6a9e (#1973) fixed on main just before this branch.

See the summary comment for the two options.

packages/cli/tests/claude-plugin/automatic-migration.test.ts
```diff
@@ -0,0 +1,159 @@
...
+const originalProjectDirectory = process.env.CLAUDE_PROJECT_DIR;
+
+function fixture(version = '0.72.0'): { root: string; installedPath: string } {
+  const root = mkdtempSync(nodePath.join(tmpdir(), 'safeword-auto-claude-'));
+  roots.push(root);
+  const release =
+    CLAUDE_HISTORICAL_CATALOGUE.releases[
+      version as keyof typeof CLAUDE_HISTORICAL_CATALOGUE.releases
+    ];
+  const installedPath = Object.keys(release.files)[0];
```
TheMostlyGreat, line 28: Reviewer (context for finding #2): the fixture installs exactly one file — `Object.keys(release.files)[0]`. That's why no test caught the empty-directory husks: every catalogued path is nested at least two deep, but a one-file fixture never produces a directory worth pruning.

Worth considering a fixture that installs the whole release tree for at least one scenario. Running the real 0.72.0 tree through migration is what surfaced the 22 stranded directories.

packages/cli/src/claude-plugin/runtime/dispatch.ts
```diff
@@ -398,6 +450,259 @@ function mergeHookOutput(event: string, target: HookResponse, output: string): v
...
+    attention?.plugin_version !== identity.plugin_version ||
+    attention.catalogue_sha256 !== catalogueSha256 ||
+    attention.watched_settings_sha256 !== claudeWatchedSettingsDigest(projectRoot)
+  ) {
+    return undefined;
+  }
+  return attention;
+}
+
+function automaticMigrationProjectRoot(event: string): string | undefined {
```
TheMostlyGreat: Reviewer (should-strengthen, finding #5): when `CLAUDE_PROJECT_DIR` is unset or empty this returns `undefined`, `automaticMigrationUnsafe` returns immediately, and the whole feature is silently off — no migration, no marker, no advisory, nothing in `safeword claude status`.

`canonicalClaudeProjectRoot()` in `project-root.ts` already implements the `git rev-parse --show-toplevel` fallback for exactly this case, but it is only ever reached via `safeword claude recover` because dispatch hands it an already-resolved root. Reusing `canonicalClaudeProjectRoot(process.cwd())` here would make the two paths agree.

No scenario covers a missing `CLAUDE_PROJECT_DIR`, so a regression here would also go unnoticed.

packages/cli/src/claude-plugin/migration-state.ts
```diff
@@ -0,0 +1,213 @@
...
+      : '';
+  } catch {
+    return '';
+  }
+}
+
+/** Atomically claims one bounded automatic-migration launch for a Claude session. */
+export function claimClaudeMigrationAttempt(cwd: string, sessionId: string | undefined): boolean {
+  const sessionDigest = digest(sessionId?.trim() || 'unknown-session');
+  const limit = initialSessionDigest(cwd, sessionDigest) === sessionDigest ? 3 : 1;
```
TheMostlyGreat: Reviewer (should-strengthen, finding #6): the budget is 3 launches for the session that wins `initial-session-v1.json` (the first session ever to run migration in the repo) and exactly 1 for every session after.

So in any later session, an attempt that defers with a durable transaction already on disk cannot be finished by a subsequent prompt in that same session — every one is refused with "run `safeword claude recover`". feature:117-121 promises "the next successful prompt completes the recorded transaction and enters plugin mode" without qualifying the session, so the scenario and the code disagree.

Completing an existing transaction plans nothing new — it only replays recorded images — so it could reasonably be exempt from the launch budget. Otherwise the scenario should name the budget.

packages/cli/src/claude-plugin/legacy-classifier.ts
```diff
@@ -0,0 +1,116 @@
...
+function observeFiles(
+  cwd: string,
+): Pick<ClaudeLegacyObservation, 'recognizedFiles' | 'conflictingFiles'> {
+  const recognizedFiles: string[] = [];
+  const conflictingFiles: string[] = [];
+  for (const relativePath of cataloguedClaudeLegacyPaths()) {
+    const path = nodePath.join(cwd, relativePath);
+    if (!existsSync(path)) continue;
+    try {
+      const regular = lstatSync(path).isFile();
```
TheMostlyGreat: Reviewer (should-strengthen, finding #7): this correctly rejects a symlinked *leaf* — the case the `@rejection` scenario at feature:51 covers.

A symlinked *ancestor* takes a different path. If `.claude/` itself is a symlink, `existsSync`/`lstat` on the leaf still report a regular file, so the path is classified `recognized` and enters `mutations`. `assertSafeTarget` then throws on the `.claude` segment and the entire migration falls into the generic catch in `migrateClaudeLegacyAutomatically` — nothing gets contracted at all, and the advisory carries an exception string instead of naming the unsafe path and its repair action, which is what feature:56 promises.

Classifying a symlinked ancestor as conflicting here would degrade one path instead of the whole migration.

TheMostlyGreat (commented):
> [!NOTE]
> **Status since this review:** all four must-fix findings are now fixed and pushed, including #1 (the BDD binding), which this review left open for your call — see [the follow-up](https://github.com/ArcadeAI/safeword/pull/2016#issuecomment-5200215629). Findings #5–#8 remain open by choice.

---

Reviewer: ran `/review-spec`, `/quality-review`, and `/refactor` over this branch. I pushed fixes for the four findings that were mechanically provable; the rest need your call. This review is `COMMENT` only — `gh` authenticates as the PR author, so it carries no merge block.

**Findings:** 4 must-fix (3 fixed & pushed, 1 needs your decision), 4 should-strengthen, 1 refactor applied.

---

## Must fix

### 1. The BDD lane for this feature proves nothing — 24 scenarios assert one boolean — NOT FIXED, needs your call

`steps/automatic-claude-migration.steps.ts` doesn't define steps. It regex-scans the `.feature` file at import time and auto-registers whatever it finds:

- every `Given` gets an **empty body**;
- every `Then`, regardless of what it says, runs `assert.equal(result.status, 0)` against a **single module-cached** `bun run test` of nine unit files;
- every `When` just triggers that same cached run.

Consequences:

- Scenario prose is decorative. "the symlink and external file remain byte-for-byte unchanged" (feature:54), "exactly one transaction is created" (feature:96), and "the completed transaction is removed" (feature:108) all assert the identical boolean. Delete any one scenario's underlying behavior and the lane still passes as long as the nine unit files do.
- **No scenario in this feature can ever be undefined or red on its own.** Rewording a scenario auto-registers a new passing step, so the feature file can drift arbitrarily from reality without CI noticing.
- `test:bdd` is a blocking CI step, so this converts a gate into a rubber stamp for this feature.
- It's the only steps file in the repo that does this — `steps/native-claude-plugin.steps.ts` builds real fixtures, real settings files, real symlinks, and asserts real bytes. That's the bar the other 36 step files meet.

This is the same anti-pattern `4495e6a9e` ("stop a migration-safety scenario proving nothing", #1973) fixed on main a few commits before this branch.

I did not attempt the rewrite — binding 24 scenarios to real fixtures is a substantial piece of work and the shape is yours to choose. Two options:

1. Write real step definitions modelled on `native-claude-plugin.steps.ts`. Highest confidence, largest effort.
2. Narrow the feature file to the scenarios you're willing to bind for real, and drop the rest into `spec.md` prose. Smaller feature, honest gate.

What isn't viable is leaving it as-is, because the feature file currently reads as coverage for the riskiest one-way file-deletion path in the product.

### 2. Contraction leaves 22 empty directory husks — FIXED

Reproduced by installing the full 0.72.0 legacy tree (29 files) into a temp repo and running `migrateClaudeLegacyAutomatically`:

```
INSTALLED: 29 STATE: complete unresolved: 0
EMPTY DIRS LEFT (22):
.claude/agents
.claude/skills/bdd
.claude/skills/audit
.claude/skills/figure-it-out
... (18 more)
.claude still present: true
```

All 29 catalogued paths are nested at least two deep (`.claude/skills/<skill>/SKILL.md`), and `writeImage` only did `rmSync` on the file. So the user is left with a husk tree to clean up by hand — which contradicts feature:11 ("every accepted legacy Claude asset and settings entry is removed") and the Rave Moment in `spec.md`.

**Why the unit tests missed it:** `fixture()` in `automatic-migration.test.ts` installs only `Object.keys(release.files)[0]` — exactly one file — so no test ever sees the real tree shape.

Fixed by pruning ancestors that Safeword just emptied, stopping at the first non-empty directory and never leaving the project root. Added a regression test covering both directions: a directory with a user-authored sibling survives; a directory Safeword fully emptied is removed.

### 3. Every new piece of per-session state is committed to the user's repo — FIXED

This PR adds four hook-written paths under `.safeword/claude-plugin/` — `plugin-mode-v2.json`, `attention-v1.json`, `attempts-v1/**`, and `cleanup-transaction-v1.json` — and none of them is in `SAFEWORD_TRANSIENT_PATHS`. `attempts-v1/` grows **one file per Claude session** (plus one per distinct advisory state per session), unbounded.

That's precisely the "recurring repository churn" `spec.md` says this feature eliminates, and the schema comment on that list spells out the failure mode: untracked files block the auto-upgrade gate.

Fixed by adding `.safeword/claude-plugin/` to `SAFEWORD_TRANSIENT_PATHS`, plus the dogfood `.gitignore`.

### 4. Two callers disagree on where the Claude user config lives — FIXED

- `claudeWatchedSettingsDigest` (migration-state.ts): `process.env.CLAUDE_CONFIG_DIR ?? join(homedir(), '.claude')`
- `incompatibleScopeOverlap` (dispatch.ts): trims, and falls back to the default on empty

With `CLAUDE_CONFIG_DIR=""` or whitespace, the first resolves the user settings path to a **cwd-relative `settings.json`** while the second reads `~/.claude/settings.json`. The attention record then watches a file that isn't the one the scope-overlap decision was made from, so a real user-scope change never re-triggers evaluation — the advisory stays suppressed for the life of the session.

Fixed by extracting one exported `claudeConfigDirectory()` and using it in both places.

---

## Should strengthen

### 5. Automatic migration silently no-ops when `CLAUDE_PROJECT_DIR` is unset

`automaticMigrationProjectRoot()` returns `undefined` when the env var is missing or empty, and `automaticMigrationUnsafe` then returns immediately — no migration, no marker, no advisory, nothing in `safeword claude status`. The feature is simply off, invisibly.

Meanwhile `canonicalClaudeProjectRoot()` in `project-root.ts` already implements the `git rev-parse --show-toplevel` fallback for exactly this — but it's only ever reached through `safeword claude recover`, never through the automatic path, because dispatch passes an already-resolved root.

Worth reusing `canonicalClaudeProjectRoot(process.cwd())` in dispatch so the two paths agree. No scenario covers a missing `CLAUDE_PROJECT_DIR`, so nothing would catch a regression here either.

### 6. "The next successful prompt completes the recorded transaction" only holds in the repo's first-ever session

`claimClaudeMigrationAttempt` grants **3** launches to the session whose digest happens to win `initial-session-v1.json` (the first session ever to run migration in that repo), and exactly **1** to every session after that.

So in any later session: attempt 1 defers with a durable transaction on disk, and every subsequent prompt in that session is refused with "run `safeword claude recover`" rather than finishing the transaction that's already planned and safe to replay. feature:117-121 states the completion promise without qualifying the session, so the scenario and the code disagree.

Suggest either allowing a completion-only attempt when a transaction already exists (it plans nothing new — it only replays recorded images), or qualifying the scenario to name the session budget.

### 7. A symlinked _ancestor_ aborts the whole migration instead of skipping one path

The `@rejection` scenario at feature:51 covers a symlinked **leaf**, which `observeFiles` handles cleanly (`lstatSync(path).isFile()` is false → conflicting → preserved + named).

A symlinked **ancestor** behaves differently. If `.claude/` itself is a symlink, `existsSync`/`lstat` on the leaf still report a regular file, so the path is classified `recognized` and enters `mutations`. `assertSafeTarget` then throws on the `.claude` segment, and the entire migration falls into the generic `catch` in `migrateClaudeLegacyAutomatically` — **nothing** gets contracted, and the advisory carries an exception string rather than naming the unsafe path and its repair action.

Suggest classifying a symlinked ancestor as conflicting during observation, so one bad path degrades one path instead of the whole migration.

### 8. Invariant with no scenario that would falsify it

`spec.md` states: _"every target requires exact historical ownership proof plus a durable, idempotent transaction before automatic contraction."_

The nearest scenario (feature:59, uncatalogued release preserved) asserts the bytes survive and the digest is recorded, but says nothing about the transaction. Nothing in the feature would fail if contraction ever ran without first writing one. Given #1 above, this is currently unbound twice over.

---

## Refactor

**Applied:** `historical-ownership.ts` wrote out `[CLAUDE_HISTORICAL_CATALOGUE.current, ...Object.values(CLAUDE_HISTORICAL_CATALOGUE.releases)]` **four times**, each with its own inline `as Record<...>` cast. Extracted `acceptedReleases()` and `acceptedHookFingerprints()` behind one typed `CatalogueRelease` interface — 50 lines to 40, and the casts now live in one place. Behavior-identical; `acceptedHistoricalHookEntries` keeps its dedupe.

**Recommended, not applied** (both are yours to take or leave — they touch the generated plugin bundle):

- `dispatch.ts` repeats `try { return {...execution, stdout: appendMigrationAdvisory(...)} } catch { return execution }` in four places (`automaticMigration`, `executionProofFailure`, `functionalExecutionFailure`, `startupFailure`, and a variant in `verifiedIdentity`). One `safeAdvisory(event, execution, advisory)` helper removes about 30 lines and makes the fail-open guarantee auditable in one spot.
- The plugin-mode marker literal (`state: unresolved.length === 0 ? 'clean' : 'unresolved'` plus five fields) is constructed in three places — `cleanup.ts:315`, `cleanup.ts:380`, `dispatch.ts:580`. A single `pluginModeMarker(...)` in `migration-state.ts` would keep the three in step as the schema grows.

---

## Looks good — specifically

- **`writeDurableFileExclusive` via `linkSync`.** This is the correct primitive for a single-winner claim and it's genuinely race-safe, unlike the `existsSync`-then-write that this kind of code usually ships with. The temp-file cleanup in `finally` is right too.
- **Asymmetric failure policy.** Fail-open on `UserPromptSubmit`, fail-closed (exit 2) on every other event, with the reasoning written down at `functionalExecutionFailure`. Guessing at authorization output would have been the easy wrong call.
- **`assertSafeTarget` walks every path segment** for symlinks rather than only checking the leaf and the containment prefix. That's stronger than the usual check — finding #7 is about classification, not about this function.
- **The inventory closure check** (rejecting any on-disk plugin file absent from the signed inventory) closes a real hole: integrity checks that only verify _listed_ files let an attacker _add_ files freely.
- **Recovery accepting either recorded image and refusing a third** is the right idempotency contract, and `pendingRecoveryEntries` implements it exactly as specified.

---

## Verification of the pushed changes

- `packages/cli` claude-plugin suite: **108 passed** (was 107; +1 new regression test)
- schema / owned-paths / git / reconcile-namespace-root / claude-plugin-release suites: passed
- `bun run lint` (eslint + `tsc --noEmit`): clean
- `bun run generate:claude-plugin` re-run; `plugin/runtime/*` and `plugin/identity.json` regenerated from source

I did not run the full suite or the BDD lane locally — per repo convention, CI's isolated run is authoritative there.

---

**Next:** decide on finding #1. Everything else is either landed or a small follow-up.

steps/automatic-claude-migration.steps.ts
TheMostlyGreat: Reviewer: **resolved in `eb5d7bad8`** — this comment is outdated because the generated step definitions it describes no longer exist.

The file now defines real steps against a driver module (`steps/support/claude-migration-fixtures.ts`). Most scenarios drive the real generated plugin through its own `hooks.json` command over the exact bytes a release installed; deadline and recovery scenarios use the injected clock; the concurrency scenario runs two real hook processes.

Proved by mutation rather than by passing: removing the husk pruning kills exactly the three scenarios claiming complete removal, and treating every managed path as accepted kills exactly the two claiming preservation — 24 and 25 respectively stay green. Under the old bindings either mutation would have killed all 24 or none.

packages/cli/tests/claude-plugin/automatic-migration.test.ts
TheMostlyGreat: Reviewer: partly addressed, deliberately left as-is here.

The acceptance lane now installs the **whole** release tree (`createLegacyProject({ release })`, all 29 catalogued assets), which is what surfaced the 22 stranded directories. This unit fixture still installs one file — fine for the focused cases it serves, and no longer the only coverage of tree shape. Flagging rather than changing it, since the gap it caused is now covered a level up.

TheMostlyGreat: Resolved with the coverage split kept intentional. The focused unit fixture remains one-file because its assertions are local; the acceptance driver installs every catalogued file from each real release and proves complete tree contraction for 0.68.0, 0.69.0, and 0.72.0. The exact final feature lane passes 29/29 scenarios.

steps/automatic-claude-migration.steps.ts
TheMostlyGreat: Resolved on the updated head. The generated catch-all bindings were replaced by real scenario-specific drivers in eb5d7bad8; this follow-up adds the two missing rejection cases and the exact final lane passes 29/29 scenarios and 1,166/1,166 steps. The transaction-before-mutation case also caught and drove a real recovery-message fix, which is evidence that the bindings now falsify behavior rather than only checking an exit code.

packages/cli/src/claude-plugin/runtime/dispatch.ts
TheMostlyGreat: Resolved in 02c38e73d. The dispatcher now consumes the documented hook-input cwd and routes it through canonicalClaudeProjectRoot; CLAUDE_PROJECT_DIR still has precedence when present. Execution proof and automatic migration use the same resolver. A generated-runtime regression omits CLAUDE_PROJECT_DIR and proves contraction plus the durable plugin-mode marker; it passes.

packages/cli/src/claude-plugin/migration-state.ts
TheMostlyGreat: Resolved in 02c38e73d with a bounded recovery-only budget. The initial session still has three total launches shared by migration and recovery. Each later session has one normal launch plus one separate launch only when a durable transaction already exists. This lets the next prompt finish recorded work without allowing a third-image conflict to retry forever. Unit and real-hook acceptance coverage pass.

packages/cli/src/claude-plugin/legacy-classifier.ts
TheMostlyGreat: Resolved in 02c38e73d. Ancestor containment and symlink validation now live in one shared cleanup-target module used during both classification and mutation. A path below a symlinked ancestor is classified as conflicting before planning, so it is preserved and reported instead of aborting unrelated contraction. The regression is included in the 145/145 focused test run.

packages/cli/src/claude-plugin/runtime/dispatch.ts
TheMostlyGreat: Reviewer: **resolved.** `automaticMigrationProjectRoot` now calls `canonicalClaudeProjectRoot(hookCwd ?? process.cwd())`, so the git-toplevel fallback is reachable from the automatic path and migration no longer disables itself when the env var is missing.

One note carried into the second-pass comment: `recordExecutionProof` took the same change, which means the fallback path now derives the project root from the hook's stdin `cwd`. Worth validating that against the resolved git root, or dropping it for `process.cwd()` — the hook already runs in the project.

packages/cli/src/claude-plugin/migration-state.ts
TheMostlyGreat: Reviewer: **resolved.** `claimClaudeMigrationAttempt` now takes a `kind`, and a later session gets a dedicated `recovery` launch alongside its one migration launch, so a deferred transaction is finished by the next prompt instead of waiting for a new session. The initial session's recovery still draws on its three-launch budget, and both directions are pinned by tests.

The scenario at feature:130 ("A later session can recover after spending its normal launch") now states the promise the code actually makes.

packages/cli/src/claude-plugin/legacy-classifier.ts
TheMostlyGreat: Reviewer: **resolved.** `observeFiles` runs `assertSafeClaudeCleanupTarget` and classifies anything it rejects as conflicting, so a symlinked *ancestor* now degrades that one path instead of aborting the whole migration into the generic catch. The check moved to `cleanup-target.ts`, shared with `cleanup.ts` without a cycle.

TheMostlyGreat:
Reviewer: your "shouldn't our local linters be addressing this?" question was fair, and the honest answer is two-part.

**The immediate failure genuinely wasn't a lint gap.** Both test jobs failed identically on `tests/cli-protocol/documentation.test.ts` — a unit test, not part of `bun run lint` or `bun run test:bdd`. It asserts the committed `packages/cli/architecture.generated.md` has zero `⚠ stale` / `⚠ orphaned` markers. All 30 module sections carried the same reconciled stamp, one commit behind the frontmatter fingerprint — main's own commit history since the last human reconciliation had moved the structure hash without anyone bumping the stamps.

I checked every section's one-line prose against the module it describes before touching anything (spot-checked against code I'd been reading all through this review — `claude-plugin`, `review`, `pr-review`). None were wrong; the fingerprint changed, the descriptions didn't need to. That's the ordinary case this design exists for. No CLI verb bumps the stamp automatically — the code comment says so explicitly ("surfaced, never silently healed"), so it's a manual/agent call. Bumped all 30 to the current fingerprint; `architecture --check` now reports Healthy.

**The gap you're really asking about is real, and it's mine.** I'd been running `tests/claude-plugin/` — the directly-relevant suite — on every round, for speed. That's why three straight rounds of green claude-plugin tests never caught a failure sitting one directory over in `cli-protocol/`. Widened this round to `cli-protocol/` + `claude-plugin/` + the release-contract suite (302/302) and I'll keep that wider net for the rest of this review rather than narrowing back down.

## Also in this push

Merged twice more: an unrelated release-workflow commit, then **v0.74.2 itself**. The version bump regenerated cleanly with no manual fixes needed — confirming the `SAFEWORD_SCHEMA.version` fix from two merges ago (assertions that used to hardcode `'0.73.0'`) actually holds through a real version bump, not just the one I originally broke on.

## Verification

- cli-protocol + claude-plugin + release-contract: **302/302**
- migration, Codex migration, predictable-CLI lanes: **133/133**
- `prettier --check .`, `architecture --check`: clean

## Still yours

Unchanged: the `setup` decline-vs-advisory call.

**Next:** watch CI on `26dc33351`.

PR #2019, opened 2026-08-06
Make post-upgrade setup and restart guidance converge

TheMostlyGreat:
## Summary

- make dependency readiness converge after upgrades without trusting stale markers, masked install failures, escaping workspace paths, or an unloaded PostToolUse hook
- keep legacy Codex protection in place until a restarted app proves native plugin hooks and an exact finalization plan is explicitly confirmed
- preserve Claude project enrollment while choosing install versus `/reload-plugins` from evidence the setup run can actually observe
- make Codex marketplace replacement, partial-failure reporting, and recovery explicit and shell-safe
- regenerate the Claude and Codex plugin bundles and keep dogfood/runtime mirrors aligned

## Root causes and fixes

The original upgrade path assumed the running host had loaded newly installed hooks. Setup could therefore remove legacy Codex protection before native-hook proof existed, while stale dependency recovery could depend on a PostToolUse hook unavailable in the current task.

This branch keeps legacy protection through generic setup; only explicit finalization after current hook proof removes it. Dependency readiness treats a present fingerprint marker as authoritative, rejects workspace traversal and symlink escapes, refuses compound installs whose failure can be masked, and gives an old host a self-contained recovery that clears stale evidence only after a successful install.

Codex marketplace handling now preserves newer pins, fails closed for ambiguous non-Git sources, shell-quotes recovery argv, and reports structured partial mutation when install or restoration fails. Reviewer overflow/timeout cleanup terminates descendant process trees and allows the requested ten-minute headless budget.

## Spec, quality, and refactor review

`/review-spec`, `/quality-review`, and `/refactor` were rerun iteratively on the updated PR. The final spec and quality coordinators approved the result. All actionable findings were resolved, including:

- replacing synthetic/tautological BDD assertions with authoritative handler and public CLI execution
- proving recovery overwrites a genuinely changed profile and writes fresh hook proof
- aligning typed finalization preview/replay with explicit confirmation
- asserting human-action versus executable-command shapes at the acceptance boundary
- isolating mutable host state in public-command fixtures
- replacing mirror tests and duplicate assertions with behavior-specific checks
- preserving structured failure codes and partial-mutation state
- regenerating plugin integrity inventories after the final `main` merge and improving failed-test diagnostics

Large cross-cutting ideas were intentionally not folded into this safety PR: redesigning public migration results around a new partial-mutation type, centralizing all POSIX quoting across unrelated modules, replacing the lightweight Claude settings observation with a subprocess-based version observer, and introducing a new shared dependency-workspace/template module. Each changes public behavior or broad architecture and is not required to close #1996/#2014.

## Dependency and tooling evidence

- verified [openai/codex#18115](https://github.com/openai/codex/issues/18115) remains open at the current [Codex 0.147.0 release](https://github.com/openai/codex/releases/tag/rust-v0.147.0), so the project-scope tripwire remains required
- npm currency checked: Cucumber 13.2.1, dependency-cruiser 18.1.1, and Knip 6.32.0
- `plugin/**` is generated and validated by catalogue/inventory/parity tests; dependency-cruiser retains one existing `no-orphans` warning for the shell-dispatched Codex hook entrypoint

## Final local verification

Exact merged head `f36c19b20` includes `origin/main` through `a717778d3`.

- relay Vitest: 167 passed / 1 skipped across 8 files
- CLI Vitest: 6,860 passed / 5 skipped across 441 files
- Cucumber: 1,123 passed / 3 skipped scenarios; 44,831 passed / 4 skipped steps
- ESLint, Gherkin lint, TypeScript, Prettier, build, Knip, dependency-cruiser, and `git diff --check`: clean (dependency-cruiser has the one documented shell-entrypoint warning)
- all 148 ticket test definitions complete

A pre-merge full run exposed one load-sensitive lock-runner failure; the exact case and its full 11-test file passed in isolation, and subsequent clean full runs passed. After merging current `main`, full verification correctly caught stale Claude plugin integrity metadata; the bundle was regenerated, its five affected tests passed together, and the complete unit plus acceptance suite then passed.

Fixes #1996
Fixes #2014

## CI and readiness

Exact-head GitHub CI [run 31153738664](https://github.com/ArcadeAI/safeword/actions/runs/31153738664) is fully green for `f36c19b20` on Node 22.23.2 and 24.18.1. The preceding heads exposed one lockfile-only normalization line after the final `main` merge and a Node 22 wall-clock test that completed in 1,012 ms against a 1,000 ms ceiling. The generated lockfile separator changes no dependency version; the test-only ceiling is now 1.5 seconds while the production relay deadline and aggregate budget remain unchanged.


packages/cli/src/codex-plugin/migration.ts
```diff
@@ -47,8 +55,15 @@ export interface CodexMigrationFacts {
...
-  'This Codex app may keep its loaded Safe Word catalogue. Restart Codex, start a new task, then review the installed hooks with /hooks.';
+const CODEX_RESTART_INSTRUCTION =
+  'Restart Codex, start a new task, then review the installed hooks with /hooks.';
+
+/**
+ * Why the app may still be stale. The instruction that resolves it travels as a
+ * typed human next action, so renderers must not repeat it here — a message and
+ * a `Next:` line carrying the same sentence reads as a stutter, not as emphasis.
+ */
+export const CODEX_RESTART_CONTEXT = 'This Codex app may keep its loaded Safe Word catalogue.';
```
TheMostlyGreat, line 66: Reviewer: **finding 1 (fixed)** — `dcfad5ac` folded `CODEX_RESTART_INSTRUCTION` into `CODEX_RESTART_GUIDANCE`, but `renderCodexMigrationHuman` pushes that constant and *then* appends a `Next:` line carrying the same sentence. The restart states printed the instruction back to back:

```
Codex migration: plugin_installed_app_restart_required
Protection: unprotected
This Codex app may keep its loaded Safe Word catalogue. Restart Codex, start a new task, then review the installed hooks with /hooks.
Next: Restart Codex, start a new task, then review the installed hooks with /hooks.
```

Same shape in the `CliResult` path — `operations.ts` used the guidance as the finding message, and `renderHumanStreams` appends `Next:` after it. Before this PR the `Next` line was `safeword codex status`, so the stutter is new.

Split into `CODEX_RESTART_CONTEXT` (why the app may be stale) and `CODEX_RESTART_INSTRUCTION` (what to do), with the instruction living only in the typed human action — which is what "centralize the restart instruction" was reaching for. Test added asserting the sentence appears exactly once.

TheMostlyGreat (commented):
Reviewer: inline anchors for the `/review-spec` + `/quality-review` + `/refactor` findings. The narrative version with reasoning, currency checks, and the commit table is in https://github.com/ArcadeAI/safeword/pull/2019#issuecomment-5206471320 — these comments pin each finding to its line.

Four defects fixed in `f03c07d` (comments 1–4 below show what changed and why). Two left as your call (5–6). CI is green on `bff92da` and the PR is mergeable.

No merge block from this review — `gh` authenticates as you, so `APPROVE`/`REQUEST_CHANGES` 422s and this posts as `COMMENT`.

packages/cli/src/codex-plugin/migration.ts
```diff
@@ -182,10 +198,12 @@ export function renderCodexMigrationHuman(result: CodexMigrationResultV2): strin
     result.state === 'plugin_installed_app_restart_required' ||
     result.state === 'plugin_enabled_hook_unproven'
   ) {
-    lines.push(CODEX_RESTART_GUIDANCE);
+    lines.push(CODEX_RESTART_CONTEXT);
   }
   const next = result.next_actions[0];
-  if (next !== undefined) lines.push(`Next: ${next.command}`);
+  if (next !== undefined) {
+    lines.push(`Next: ${'command' in next ? next.command : next.instruction}`);
```
TheMostlyGreat, line 205: Reviewer: this `Next:` line is now the sole home of the restart instruction — the message above it carries only the context sentence. Please keep it that way: re-embedding the instruction in `CODEX_RESTART_CONTEXT` silently reintroduces the duplication, since every renderer appends this line.

packages/cli/src/commands/converge-setup.ts
```diff
@@ -599,32 +602,81 @@ function setupResult(input: SetupResultInput): CliResult {
...
+      requiresHuman: true,
+    };
+  }
+  if (input.claudePluginConverged) {
+    return { command: '/reload-plugins', mutates: false, requiresHuman: true };
+  }
+  return { command: 'safeword claude install', mutates: true, requiresHuman: true };
+}
+
+function projectClaudePluginEnabled(cwd: string): boolean {
```
TheMostlyGreat: Reviewer: **finding 2 (fixed)** — this is the load-bearing weakness. `enabledPlugins` records *enrollment*, not *version*, so this returns `true` identically for a project enrolled at 0.72 and one at 0.73.

As written on `dcfad5ac`, setup preferred `/reload-plugins` whenever this was true. After a CLI upgrade where setup rewrites delivered files, the user got `Next: /reload-plugins` — which re-reads the build Claude already holds. For a project-scope install from the git marketplace, `safeword claude install` is what actually converges a stale enrollment. That's a non-converging next action: the same class of defect this PR closes for Codex, reintroduced on the Claude side.

Fixed by gating the reload on `changed === false` (see the guard above), which is the one condition under which enrollment-without-version really does prove convergence.

**Worth a follow-up:** the durable fix is to ask `observeApplicableClaudePlugins()` — which knows the installed version — instead of hand-reading the settings file. This is a second, weaker source of truth for a question the codebase already answers. I didn't do it because it shells out to `claude`, which setup may not want. Your call.

packages/cli/src/commands/converge-setup.ts
```diff
@@ -599,32 +602,81 @@ function setupResult(input: SetupResultInput): CliResult {
...
+    ...(actionRequired
+      ? []
+      : [
+          {
+            code: 'SETUP_CODEX_PLUGIN_HANDOFF',
+            message:
+              'Codex bootstrap is enrolled for this project; each developer profile is checked automatically at task start.',
+            severity: 'info' as const,
+          },
+        ]),
```
TheMostlyGreat, line 619: Reviewer: **refactor cleanups folded into `f03c07d`**, both in this function:

- `resultFindings.push(...)` mutated the `findings` array *in place* whenever `actionRequired` was true — the ternary aliased it rather than copying. Harmless today, a trap for whoever reads `findings` after the push. Rebuilt as a single expression.
- `mutates: !command.startsWith('/')` inferred a semantic property from a string prefix. Replaced with `setupNextAction()`, which states `mutates` explicitly per branch.

packages/cli/templates/hooks/lib/dependency-readiness.ts
```diff
@@ -488,30 +488,34 @@ export function toDependencyReadinessState(
...
+  const { installCommand, plan, status } = readiness;
+  if (installCommand === undefined) return 'install dependencies';
+
+  // A version-bump pull changes the input fingerprint without changing resolved
+  // dependencies, so the install reports "no changes" and does not refresh the
+  // marker — which would otherwise leave this stale check looping. No package
+  // manager offers a cheap "lockfile already satisfied" probe (pnpm#4861), so
+  // chain the artifact touch that clears the check in that no-op case.
+  if (status !== 'stale' || plan === undefined) return installCommand;
+  return `${installCommand} && touch ${plan.installArtifact}`;
```
TheMostlyGreat: Reviewer: **finding 4 (fixed)** — three problems in the original one-line ternary:

1. `node_modules` was hardcoded, but the path is `plan.installArtifact`. They agree today; the moment a plan uses a different artifact the recovery points at the wrong directory.
2. `installCommand` falls back to the prose `'install dependencies'`, which the new template rendered as `install dependencies && touch node_modules` — prose chained into a shell operator.
3. The comment explaining *why* the escape exists (version-bump no-op; no cheap "lockfile already satisfied" probe, pnpm#4861) was deleted with the old line. That rationale is the non-obvious part — without it the `&& touch` reads as superstition.

All three fixed here. I also changed "from the project folder" to "from the project **root** folder" below, because the `touch` path is relative — I hit this live during review: the gate fired at me while I was in `packages/cli`, where following the instruction literally touches the wrong artifact.

Byte-parity held across `templates/`, `.safeword/hooks/`, and `plugin/runtime/hooks/`.

packages/cli/features/keep-codex-protection-continuous.feature
```diff
@@ -302,17 +302,29 @@ Feature: Keep Codex protection continuous during profile-plugin migration
...
-        | unproven with legacy           | plugin_enabled_hook_unproven       | protected   | 1            | safeword codex status              | 2         |
+        | unproven without legacy        | plugin_enabled_hook_unproven       | unprotected | 1            | Restart Codex, start a new task, then review the installed hooks with /hooks. | 2         |
+        | unproven with legacy           | plugin_enabled_hook_unproven       | protected   | 1            | Restart Codex, start a new task, then review the installed hooks with /hooks. | 2         |
         | current proof with legacy      | compatibility                      | protected   | 1            | safeword codex migrate --finalize  | 2         |
         | current proof without legacy   | plugin                             | protected   | 0            | none                               | 0         |
         | no configuration               | not_configured                     | unprotected | 1            | safeword codex migrate             | 2         |
 
+    Scenario Outline: Next-action shape distinguishes a runnable command from a human step
```
TheMostlyGreat, line 325: Reviewer: **finding 3 (fixed)** — this outline is new. Both outlines you changed assert next-action **text**, and the steps normalize the two shapes through `nextActionText()`. So an implementation that regressed to `{command: 'Restart Codex, start a new task, …'}` — an English sentence sitting in the field automated callers execute — passes every row.

Run the constant-implementation lens and it's worse: emitting a human action for *every* state also passes all 15 rows of the JSON outline. The typing invariant was pinned only in `migration.test.ts`; the feature file is where this PR's headline claim lives, and there it was untested.

This outline asserts the action *shape* across two human states and two command states, so the distinction fails loudly in either direction.

package.json
```diff
@@ -34,21 +34,21 @@
     "test:bdd:live": "cucumber-js --profile live"
   },
   "devDependencies": {
-    "@cucumber/cucumber": "^13.2.0",
-    "@openai/codex": "0.145.0",
+    "@cucumber/cucumber": "^13.2.1",
+    "@openai/codex": "0.146.1",
     "@types/node": "^26.1.2",
-    "dependency-cruiser": "^18.1.0",
+    "dependency-cruiser": "^18.1.1",
```
TheMostlyGreat, line 40: Reviewer: **finding 5 — your call.** This bump isn't mentioned in the PR description, and it's not cosmetic: it flips `PINNED_VERSION` in `codex-project-scope-tripwire.test.ts`, whose header exists precisely to make someone read Codex's changelog and confirm openai/codex#18115 is still unshipped.

I checked — **still open** (verified via the GitHub API during review; last updated 2026-07-31), so the bump is correct. But that verification belongs in the PR description, not in a reviewer's head; a future reader sees a tripwire constant edited with no stated evidence.

Suggest either splitting the bumps into their own PR, or adding one line: "verified openai/codex#18115 still open at 0.146.1."

Currency on the rest (checked against npm during review): cucumber 13.2.1, dependency-cruiser 18.1.1, lint-staged 17.3.0 are all latest; `tsx ^4.23.8` covers the latest 4.23.9; `knip` is pinned 6.31.0 against 6.32.0 latest — one minor behind, fine.

packages/cli/tests/codex-project-scope-tripwire.test.ts
```diff
@@ -50,7 +50,7 @@ import { describe, expect, test } from 'vitest';
 const repoRoot = nodePath.resolve(import.meta.dirname, '../../..');
 
 /** Newest @openai/codex known to still lack project-scoped plugin activation. */
-const PINNED_VERSION = '0.145.0';
+const PINNED_VERSION = '0.146.1';
```
TheMostlyGreat, line 53: Reviewer: this is the constant the `@openai/codex` bump flips. The file header asks the person changing it to check openai/codex#18115 and record which branch they took. I verified it's still open, so the re-pin is the correct branch — please capture that evidence in the PR description so the next reader doesn't have to re-derive it.

knip.json
```diff
@@ -1,7 +1,9 @@
 {
   "ignore": [
     ".safeword/**",
+    "plugin/**",
```
TheMostlyGreat, line 4: Reviewer: **finding 6 — your call.** This disables dead-code analysis for the *entire* generated tree, not one file.

It's defensible: the directory is generated, and `runtime/dispatch.ts` really is invoked by shell (`bun "${CLAUDE_PLUGIN_ROOT}"/runtime/dispatch.ts`), so no static analyzer can see it. Dropping `test-plan` from the binaries list is straightforwardly correct — no such script exists.

Flagging only because it's unmentioned in the description and it *subtracts* coverage rather than adding it. Worth a sentence saying why, so a future audit doesn't read the blind spot as intentional-but-forgotten.

.dependency-cruiser.cjs
```diff
@@ -9,7 +9,22 @@ const generated = require('./.safeword/depcruise-config.cjs');
 
 module.exports = {
   forbidden: [
-    ...generated.forbidden,
+    ...generated.forbidden.map(rule =>
```
TheMostlyGreat, line 12: Reviewer: same category as the knip suppression — a `no-orphans` exception for generated hook entrypoints. The stated reason (dispatched by `event-groups.json`, not imported) checks out, and the inline comment carries it, which is more than the knip change does. No change requested.

packages/cli/tests/codex-plugin/migration.test.ts
```diff
@@ -219,9 +219,7 @@ describe('Codex migration result', () => {
       state === 'plugin_installed_app_restart_required' ||
       state === 'plugin_enabled_hook_unproven'
     ) {
-      lines.push(
-        'This Codex app may keep its loaded Safe Word catalogue. Restart Codex, start a new task, then review the installed hooks with /hooks.',
-      );
+      lines.push('This Codex app may keep its loaded Safe Word catalogue.');
```
TheMostlyGreat: Reviewer: **noted, not fixed.** This test rebuilds the renderer's output line by line and then asserts equality with the renderer — it re-implements the code it checks. That means it can only detect that output *changed*, never that it's *wrong*: it happily encoded the duplicated restart sentence from finding 1 as the expected value.

I updated this line only as far as the fix required. Worth replacing with assertions on the parts that carry meaning (state, protection, one `Next:` line) rather than a full-string mirror.

packages/cli/features/steps/predictable-safeword-cli.steps.ts
```diff
@@ -126,7 +126,11 @@ function runPublicFixture(world: PredictableCliWorld, definition: CommandDefinit
     {
       cwd,
       encoding: 'utf8',
-      env: { ...childEnvironment(), ...definition.fixture.environment },
+      env: {
+        ...childEnvironment(),
+        CODEX_HOME: join(cwd, '.codex'),
```
TheMostlyGreat: Reviewer: confirming this line fixes the failure I reported — and that it holds under adversarial conditions, which is the hard part to check without a live Codex app.

Without it, the child inherited the developer's real `~/.codex`, and `codex status` embeds `recorded_at` from `$CODEX_HOME/safeword/hook-proof-v2/*.json`. Those markers are rewritten on every Codex tool call, so the two invocations this scenario compares disagreed whenever Codex did anything in between.

I verified by rewriting the real marker every 200ms for the length of the run:

| Tree | Result |
| --- | --- |
| without this line | fails at `:657` |
| with this line | passes; full lane 282/282 |

Nesting under `cwd` is the right call — `cwd` is already `rmSync`'d and recreated at the top of this function, so each invocation gets a pristine profile and a mutating command can't leak state into the second run. I had written a sibling directory with its own wipe; yours gets the same property for free, so I dropped mine.

Placing it before `...definition.fixture.environment` also keeps the per-fixture override working, which the headless-activation fixture needs.

TheMostlyGreat (commented):
Reviewer: investigated the "flake" I flagged earlier. **It was never a flake — the scenario was non-hermetic**, and you'd already fixed it upstream while I was digging. Your fix is correct; I verified it empirically (details in the inline comment). Recording the diagnosis so it doesn't have to be re-derived, and correcting my own earlier characterisation.

## Root cause

`Every public command accepts the machine contract` runs each public command twice and `deepEqual`s the JSON. `safeword codex status` embeds `recorded_at` read from `$CODEX_HOME/safeword/hook-proof-v2/*.json` — and `runPublicFixture` inherited the developer's real `~/.codex` via `{...process.env}`.

Those markers are rewritten by **every live Codex tool call**. On my machine, mid-review:

```
~/.codex/safeword/hook-proof-v2/post-tool-use.json   recorded_at 2 seconds old
~/.codex/safeword/hook-proof-v2/pre-tool-use.json    recorded_at 4 seconds old
```

This repo runs Codex constantly — the retro filer and the independent-review coordinator both shell out to it. So any Codex activity landing between the two invocations desynchronised them. CI stayed green because runners have no Codex app rewriting those files, which is exactly what made it look like load-related flakiness rather than a missing environment boundary.

## Proof

Isolated `CODEX_HOME` to a copy, ran `codex status` twice, and bumped one proof marker in between — nothing else changed:

```
simulated one Codex tool call: 20:03:51.333Z -> 20:04:03.315Z
deepEqual: DIFFERENT -> the scenario fails
```

Then the controlled A/B, rewriting the real marker every 200ms for the duration of the run:

| Tree | Result |
| --- | --- |
| unpatched | **fails** at `predictable-safeword-cli.steps.ts:657` |
| with `CODEX_HOME` isolated | **passes** |

The hammer is what a busy Codex session does naturally — it just removes the timing luck.

## Your fix, verified

I'd written my own isolation before pulling; you'd already landed `CODEX_HOME: join(cwd, '.codex')`. Yours is better — `cwd` is already `rmSync`'d and recreated per invocation, so it gets per-call freshness without the extra bookkeeping mine needed. I reverted mine (`f04c38ad`) and merged yours rather than leaving two versions of the same fix.

Re-ran the A/B against **your** implementation, since the hammer is the part that's awkward to reproduce without a live Codex:

- targeted scenario under the hammer: **passes**
- full BDD lane: **282/282** — first fully green local run of this branch for me

## Correcting myself

My earlier review called this "a pre-existing BDD flake ... only shows up under local full-suite load." The first half was right (pre-existing, not from this PR); the second half framed it as a timing problem when it was a missing environment boundary. Full-suite load only widened the window. Worth the distinction: "flake" invites a retry, and a retry would never have fixed this.

One thing your fix doesn't cover, and I'd leave alone unless it bites: `childEnvironment()` still spreads the full `process.env`, so `HOME` and anything else machine-global still reaches these fixtures. `CODEX_HOME` was the one with a live writer behind it. If another host profile grows a mutable marker, this class returns — an allowlisted child environment would close it permanently.

PR #2204, opened 2026-08-07
Show review progress during long checks

TheMostlyGreat:
## What changed

- Announce reviewer dispatch and fallback transitions on stderr.
- Announce review-packet preparation when preparation is slow enough to matter — the dispatch announcement supersedes it below the 100ms announce delay.
- Emit a truthful 30-second waiting heartbeat for long independent reviews.
- Keep JSON and quiet output unchanged.

## Why

Long review routes previously emitted no output during their 120-second timeout window, leaving users unable to distinguish a running reviewer from a stalled route.

## Review follow-ups (0091b979d)

Applied after `/review-spec`, `/quality-review`, and `/refactor` — see [the review](https://github.com/ArcadeAI/safeword/pull/2204#pullrequestreview-4880880458) for the reasoning.

- **The 30s heartbeat had no end-to-end proof.** `heartbeat` is optional on `ProgressReporter`, so deleting the coordinator's call typechecked and left the suite green. Added a clamped internal `SAFEWORD_PROGRESS_HEARTBEAT_MS` override (1ms..30s, otherwise ignored) and a scenario asserting more than one heartbeat line reaches stderr.
- **"Keep JSON and quiet output unchanged" was unbound.** Added a scenario asserting `--json` emits neither the dispatch line nor a heartbeat.
- **`stop()` was half-asserted.** This PR made it cancel both handles; only the heartbeat handle was covered. Added coverage for the pending announcement.
- **Deduped the progress contract.** `ReviewProgress` was a hand-copied structural twin of `ProgressReporter`; it is now `Pick<ProgressReporter, 'start' | 'heartbeat'>`, with `stop` excluded on purpose — the command runner owns `stop()` in its `finally`.
- **Made the two routes symmetric.** The fallback inlined its own announce block; extracted `prepareFallbackReview` to match the primary.

Known trade-off: on the fallback path the primary's heartbeat stays armed across the fallback's packet preparation, so a heartbeat can briefly name the reviewer that already failed. Bounded by prep duration and re-armed immediately after — cosmetic, left as-is.

## Validation

- `bun run lint` (eslint + gherkin + `tsc --noEmit`)
- `bun run test tests/cli-protocol/policy.test.ts tests/cli-protocol/review-wiring.test.ts` — 43 passed (was 39)

Closes #2115.

### DIFF
--- a/packages/cli/src/cli-protocol/policy.ts
+++ b/packages/cli/src/cli-protocol/policy.ts
@@ -1,4 +1,5 @@
 import type { CommandDefinition } from './catalog.js';
+import type { ProgressReporter } from './handler.js';
 import type { CliResult, Effects } from './result.js';
 
 function firstNonEmptyEffect(effects: Effects): keyof Effects | undefined {
@@ -54,24 +55,37 @@ interface ProgressAdapters {
 
 /** Operations finishing faster than this are not worth announcing. */
 const PROGRESS_ANNOUNCE_DELAY_MS = 100;
+/** A long wait needs proof that the coordinator is still responsive. */
+const PROGRESS_HEARTBEAT_INTERVAL_MS = 30_000;
 
-export function createProgressReporter(adapters: ProgressAdapters): {
-  start: (message: string) => void;
-  stop: () => void;
-} {
-  let scheduledHandle: unknown;
+export function createProgressReporter(adapters: ProgressAdapters): ProgressReporter {
+  let announcementHandle: unknown;
+  let heartbeatHandle: unknown;
+
+  function scheduleHeartbeat(message: string): void {
+    heartbeatHandle = adapters.schedule(() => {
+      adapters.emit(message);
+      scheduleHeartbeat(message);
+    }, PROGRESS_HEARTBEAT_INTERVAL_MS);
+  }
+
   return {
     start(message: string): void {
-      if (scheduledHandle !== undefined) adapters.cancel(scheduledHandle);
-      scheduledHandle = adapters.schedule(() => {
+      if (announcementHandle !== undefined) adapters.cancel(announcementHandle);
+      announcementHandle = adapters.schedule(() => {
         adapters.emit(message);
-        scheduledHandle = undefined;
+        announcementHandle = undefined;
       }, PROGRESS_ANNOUNCE_DELAY_MS);
     },
+    heartbeat(message: string): void {
+      if (heartbeatHandle !== undefined) adapters.cancel(heartbeatHandle);
+      scheduleHeartbeat(message);
+    },
     stop(): void {
-      if (scheduledHandle === undefined) return;
-      adapters.cancel(scheduledHandle);
-      scheduledHandle = undefined;
+      if (announcementHandle !== undefined) adapters.cancel(announcementHandle);
+      if (heartbeatHandle !== undefined) adapters.cancel(heartbeatHandle);
+      announcementHandle = undefined;
+      heartbeatHandle = undefined;
     },
   };
 }
--- a/packages/cli/src/cli-protocol/policy.ts
+++ b/packages/cli/src/cli-protocol/policy.ts
@@ -58,15 +58,32 @@ const PROGRESS_ANNOUNCE_DELAY_MS = 100;
 /** A long wait needs proof that the coordinator is still responsive. */
 const PROGRESS_HEARTBEAT_INTERVAL_MS = 30_000;
 
+/**
+ * Shorten the heartbeat so a test can observe a real one without waiting 30
+ * seconds. Internal: a value outside 1ms..30s is ignored.
+ */
+export function resolveHeartbeatIntervalMs(environment: NodeJS.ProcessEnv = process.env): number {
+  const override = Number(environment.SAFEWORD_PROGRESS_HEARTBEAT_MS);
+  if (
+    !Number.isSafeInteger(override) ||
+    override < 1 ||
+    override > PROGRESS_HEARTBEAT_INTERVAL_MS
+  ) {
+    return PROGRESS_HEARTBEAT_INTERVAL_MS;
+  }
+  return override;
+}
+
 export function createProgressReporter(adapters: ProgressAdapters): ProgressReporter {
   let announcementHandle: unknown;
   let heartbeatHandle: unknown;
+  const heartbeatIntervalMs = resolveHeartbeatIntervalMs();
 
   function scheduleHeartbeat(message: string): void {
     heartbeatHandle = adapters.schedule(() => {
       adapters.emit(message);
       scheduleHeartbeat(message);
-    }, PROGRESS_HEARTBEAT_INTERVAL_MS);
+    }, heartbeatIntervalMs);
   }
 
   return {
--- a/packages/cli/src/review/coordinator.ts
+++ b/packages/cli/src/review/coordinator.ts
@@ -231,16 +231,41 @@ function changedReviewResult(input: {
   });
 }
 
-async function runDegradedFallback(input: {
+type ReviewProgress = {
+  readonly start: (message: string) => void;
+  readonly heartbeat?: (message: string) => void;
+};
+
+type ReviewRunInput = {
   readonly cwd: string;
   readonly kind: ReviewKind;
   readonly targets: readonly string[];
-  readonly author: ReviewAgent;
-  readonly assignedReviewer: ReviewAgent;
-  readonly preferredFailure: ReviewFailure;
-  readonly policy: ReviewPolicy;
-}): Promise<CliResult> {
+  readonly progress?: ReviewProgress;
+};
+
+function preparePrimaryReview(input: ReviewRunInput, reviewer: ReviewAgent) {
+  input.progress?.start(`Preparing the review packet for ${agentName(reviewer)}…`);
   const prepared = prepareReviewPacket(input.cwd, input.kind, input.targets);
+  input.progress?.start(`Requesting an independent ${agentName(reviewer)} review…`);
+  input.progress?.heartbeat?.(`Still waiting for a response from ${agentName(reviewer)}…`);
+  return prepared;
+}
+
+async function runDegradedFallback(
+  input: ReviewRunInput & {
+    readonly author: ReviewAgent;
+    readonly assignedReviewer: ReviewAgent;
+    readonly preferredFailure: ReviewFailure;
+    readonly policy: ReviewPolicy;
+  },
+): Promise<CliResult> {
+  input.progress?.start(
+    `${agentName(input.assignedReviewer)} did not complete; trying a ${agentName(input.author)} fallback…`,
+  );
+  const prepared = prepareReviewPacket(input.cwd, input.kind, input.targets);
+  input.progress?.heartbeat?.(
+    `Still waiting for a response from the ${agentName(input.author)} fallback…`,
+  );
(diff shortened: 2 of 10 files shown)


TheMostlyGreat (commented):
## Reviewer: `/review-spec` + `/quality-review` + `/refactor` on this PR

Ran all three passes. **Verdict: APPROVE with changes applied** — I pushed the fixes to this branch as `refactor: dedupe the review progress contract and bind its untested invariants` (0091b979d), rebased onto your `Regenerate Claude plugin assets`. Nothing here is a behavior change to what you shipped; it's coverage for it plus two structural cleanups. Read the two must-fixes below — the first one means the headline feature of this PR could have been deleted without a single test going red.

---

### `/review-spec` — scenario quality

**Findings:** 2 must-fix, 1 should-strengthen, 2 looks-good. All resolved.

#### must-fix — the 30-second heartbeat had no end-to-end proof (wiring gap)

`heartbeat` is optional on `ProgressReporter`, so the coordinator calls it as `progress?.heartbeat?.(…)`. The unit test proves `createProgressReporter` schedules and re-arms; the two wiring tests only prove `start()` reaches stderr. **Delete the coordinator's `heartbeat?.()` call entirely and the code still typechecks and the whole suite stays green** — the PR's headline behavior was unbound.

It couldn't be tested end-to-end because nothing could shorten the 30s wait. Added `SAFEWORD_PROGRESS_HEARTBEAT_MS` (internal, clamped to 1ms..30s, ignored otherwise) and a scenario that runs a 1s fake reviewer at a 150ms heartbeat and asserts **more than one** heartbeat line on stderr. Non-vacuous: a constant implementation that emits nothing scores 0 and fails.

#### must-fix — "Keep JSON and quiet output unchanged" was an unbound invariant

The PR body states it; no scenario would falsify it. Progress leaking into `--json` stderr would corrupt machine consumers and ship green. Added `stays silent on stderr when the caller asked for machine output` — same delayed-reviewer setup, `--json`, asserts neither the dispatch line nor a heartbeat appears.

#### should-strengthen — `stop()` was only half-asserted

This PR changed `stop()` from early-returning when no handle was pending to cancelling *both* handles. `keeps a long-running operation visibly alive…` only asserts `cancel(2)` — the heartbeat handle. The announcement half of the change was unasserted. Added `cancels a pending announcement as well as the heartbeat when the command ends`.

Minor, left alone: that test's title says "without claiming completion progress" but nothing in it asserts anything about completion. The title over-promises relative to the body.

#### looks-good

- Both wiring scenarios survive the constant-implementation lens — a no-op reporter fails them on a concrete string, not on existence.
- `SAFEWORD_REVIEW_FAKE_DELAY_AGENT` is a clean per-agent seam and reuses the existing fake-reviewer harness rather than adding a parallel one. Determinism is fine: `/bin/sleep 1` against a 100ms announce delay is a 10x margin, and the assertion can only fail in the direction of "too slow", which the sleep prevents.

---

### `/quality-review`

**Currency:** n/a — no dependency or version claims; `setTimeout`/`clearTimeout` only.
**Sources:** ✓ — no external claims to trace.
**Correct:** ✓ (with one behavior now documented, see below)
**Elegant:** ⚠️ → fixed
**No-bloat:** ⚠️ → fixed
**Wiring:** ❌ → ✓ (the heartbeat gap above)

**Critical issues:** None remaining.

#### The packet-preparation message is conditional, and that was invisible

`preparePrimaryReview` calls `start()` twice back-to-back. The second cancels the first's pending 100ms timer, so `Preparing the review packet for Codex…` is emitted **only when packet preparation itself takes longer than 100ms**.

That's correct — it's exactly what the 100ms announce delay is for, and announcing a sub-100ms step then immediately superseding it would be noise. But nothing in the code said so, and the PR body reads as an unconditional claim ("Show review packet preparation"). Added a doc comment. **Suggest softening that bullet in the PR description** to something like "announce packet preparation when it is slow enough to matter".

#### NOTE — a bounded stale-heartbeat window on the fallback path

When the primary reviewer fails, the primary's heartbeat (`Still waiting for a response from Codex…`) stays armed across `prepareReviewPacket` inside the fallback, so a heartbeat can name a reviewer that has already failed. The window is bounded by packet-prep duration and the fallback re-arms immediately after, so this is cosmetic, not a correctness issue. Left as-is — closing it means either stopping the heartbeat before prep (a silent gap) or threading a cancel through, and neither is worth it for a sub-100ms window. Flagging so it's a known trade-off rather than an accident.

---

### `/refactor` — ledger

Leaf-first, one change per step, tests green after each.

1. **`ReviewProgress` was a hand-copied structural twin of `ProgressReporter`** (`start` + optional `heartbeat`, minus `stop`). Two definitions of one contract in two files, and structural typing means the compiler is happy while they drift. Now `Pick<ProgressReporter, 'start' | 'heartbeat'>`, with a comment on *why* `stop` is excluded — the command runner owns `stop()` in its `finally`, the coordinator must never end the reporter.
2. **The fallback route inlined its own announce block while the primary used a helper.** Asymmetric for no reason. Extracted `prepareFallbackReview` so both routes read identically.
3. **Rejected — making `heartbeat` required on `ProgressReporter`.** Tempting (it would have made the wiring gap a type error), but two existing test doubles in `tests/commands/migrate-codex-plugin.test.ts` implement `{start, stop}` only. Churning unrelated tests to buy what a scenario now proves isn't worth it.
4. **Rejected — a generic 3-message `prepareAnnouncedReview(input, {preparing, dispatching, waiting})` helper.** Tried it; the fallback only has two distinct messages, so it forced a duplicated field. Over-abstraction for three shared lines. Two explicit functions read better.

---

### Verification

- `bun run lint` — clean (eslint + gherkin + `tsc --noEmit`)
- `bun run test tests/cli-protocol/policy.test.ts tests/cli-protocol/review-wiring.test.ts` — **43 passed** (was 39; +4 new scenarios)
- Broader `tests/cli-protocol tests/review` run is queued behind another worktree's vitest on this machine (the global build lock) — CI's isolated full suite on this PR is the authoritative signal.

**Next:** re-read the two new wiring scenarios, soften the "Show review packet preparation" bullet in the PR description, and let CI confirm the full suite.

packages/cli/src/cli-protocol/policy.ts
```diff
@@ -54,24 +55,54 @@ interface ProgressAdapters {
...
 /** Operations finishing faster than this are not worth announcing. */
 const PROGRESS_ANNOUNCE_DELAY_MS = 100;
+/** A long wait needs proof that the coordinator is still responsive. */
+const PROGRESS_HEARTBEAT_INTERVAL_MS = 30_000;
+
+/**
+ * Shorten the heartbeat so a test can observe a real one without waiting 30
+ * seconds. Internal: a value outside 1ms..30s is ignored.
+ */
+export function resolveHeartbeatIntervalMs(environment: NodeJS.ProcessEnv = process.env): number {
```
TheMostlyGreat, line 65: Reviewer: added so the 30s heartbeat is observable in a test without a 30s wait. Clamped to 1ms..30s and silently ignored otherwise, so a bad or hostile value can only ever fall back to the production default — it can never lengthen the interval or disable the heartbeat.

This exists because the heartbeat had no end-to-end coverage at all: `heartbeat` is optional on `ProgressReporter`, so removing the coordinator's `heartbeat?.()` call typechecked and left every test green.

packages/cli/src/review/coordinator.ts
```diff
@@ -231,16 +232,62 @@ function changedReviewResult(input: {
   });
 }
 
-async function runDegradedFallback(input: {
+/**
+ * The coordinator announces and keeps a wait visible, but never ends the
+ * reporter — the command runner owns `stop()` in its `finally`.
+ */
+type ReviewProgress = Pick<ProgressReporter, 'start' | 'heartbeat'>;
```
TheMostlyGreat, line 239: Reviewer: this was a hand-copied structural twin of `ProgressReporter` — same two members, declared separately in a second file. Structural typing means the two can drift apart without the compiler noticing. `Pick<>` binds them.

`stop` stays excluded on purpose, and the comment says why: the command runner owns `stop()` in its `finally`, so the coordinator must not be able to end the reporter.

packages/cli/src/review/coordinator.ts
```diff
@@ -231,16 +232,62 @@ function changedReviewResult(input: {
...
+};
+
+/**
+ * Prepare the primary reviewer's packet while narrating the wait.
+ *
+ * The dispatch announcement supersedes the preparation one: the reporter emits
+ * only an announcement still pending after 100ms, so packet preparation is
+ * announced when it is slow enough to matter and stays silent otherwise.
+ */
+function preparePrimaryReview(
```
TheMostlyGreat, line 255: Reviewer: documenting a real but invisible behavior — the second `start()` cancels the first's pending 100ms timer, so `Preparing the review packet…` only ever surfaces when preparation itself exceeds 100ms.

That's the right behavior (it's what the announce delay is for), but the PR description reads as an unconditional claim. Worth softening that bullet to "announce packet preparation when it is slow enough to matter".

packages/cli/tests/cli-protocol/review-wiring.test.ts
```diff
@@ -774,6 +776,119 @@ describe('cross-agent review public-command wiring', () => {
...
+          SAFEWORD_NO_UPDATE_CHECK: '1',
+        },
+      },
+    );
+
+    expect(result.exitCode, result.stdout).toBe(0);
+    expect(result.stderr).toContain('Codex did not complete; trying a Claude fallback…');
+  });
+
+  it('repeats a waiting heartbeat while the independent reviewer has not answered', async () => {
```
TheMostlyGreat, line 829: Reviewer: the missing proof for this PR's headline behavior. Asserts **more than one** heartbeat line, so it catches both "never emitted" and "emitted once and never re-armed" — the re-arm is the part `scheduleHeartbeat`'s recursion is responsible for.

packages/cli/tests/cli-protocol/review-wiring.test.ts
```diff
@@ -774,6 +776,119 @@ describe('cross-agent review public-command wiring', () => {
...
+    );
+
+    expect(result.exitCode, result.stdout).toBe(0);
+    const heartbeats = result.stderr
+      .split('\n')
+      .filter(line => line.includes('Still waiting for a response from Codex…'));
+    expect(heartbeats.length).toBeGreaterThan(1);
+  });
+
+  it('stays silent on stderr when the caller asked for machine output', async () => {
```
TheMostlyGreat, line 857: Reviewer: binds the "Keep JSON and quiet output unchanged" claim from the PR description. It was stated but not falsifiable by any scenario — progress leaking into `--json` stderr would corrupt machine consumers and still ship green.

packages/cli/tests/cli-protocol/policy.test.ts
```diff
@@ -76,4 +80,61 @@ describe('CLI execution policy', () => {
...
+    expect(delays.get(1)).toBe(30_000);
+    scheduled.get(1)?.();
+    expect(emit).toHaveBeenCalledWith('Still waiting for a response from Codex…');
+    expect(delays.get(2)).toBe(30_000);
+
+    progress.stop();
+    expect(cancel).toHaveBeenCalledWith(2);
+  });
+
+  it('cancels a pending announcement as well as the heartbeat when the command ends', () => {
```
TheMostlyGreat, line 111: Reviewer: this PR changed `stop()` from early-returning when nothing was pending to cancelling both handles, but the existing test only asserted the heartbeat handle was cancelled. This covers the announcement half — including that a pending announcement is cancelled rather than emitted after the command finishes.

PR #2497, opened 2026-08-11
Clarify profile-only legacy install aliases

### LINKED ISSUE #2496: Clarify profile-only legacy install commands for CLI users
## User impact

Top-level help says `claude install -> install --agents=claude` and the Codex equivalent. That implies the commands are interchangeable.

They are not: legacy `claude install` / `codex install` only mutate profile plugin state, while canonical `install --agents=…` also reconciles project configuration.

## Expected

Keep the aliases, but describe them as profile-only legacy routes and direct users to canonical install when they want project setup. Cover the distinction in help and documentation contracts.

TheMostlyGreat:
Fixes #2496.

Clarifies that legacy `claude install` and `codex install` are profile-only routes, while canonical `install --agents=…` also reconciles project configuration. Keeps every alias and its behavior intact.

Validation:
- focused CLI contract tests (33 CLI + 167 relay)
- CLI contract check
- Claude plugin release contract check

### DIFF
--- a/packages/cli/src/cli-protocol/catalog.ts
+++ b/packages/cli/src/cli-protocol/catalog.ts
@@ -110,22 +110,27 @@ function alias(name: string, aliasFor: string): CommandDefinition {
 function scopedInstallAlias(name: string, agent: 'claude' | 'codex'): CommandDefinition {
   const canonical = canonicalDefinition('install');
   return {
-    ...command(name, `Deprecated alias for install --agents=${agent}`, canonical.effectClass, {
-      promptPolicy: canonical.promptPolicy,
-      networkPolicy: canonical.networkPolicy,
-      commandOptions: agent === 'claude' ? [claudeScopeOption()] : [],
-      // The retained spelling keeps its shipped safety guarantee: profile-only
-      // installation that leaves the repository untouched (main's Rule
-      // codex-plugin-install.TBU1.R2). `install --agents=<agent>` is the
-      // canonical route that also reconciles project configuration.
-      handler: publicHandler(name),
-    }),
+    ...command(
+      name,
+      `Deprecated profile-only install; install --agents=${agent} also reconciles the project`,
+      canonical.effectClass,
+      {
+        promptPolicy: canonical.promptPolicy,
+        networkPolicy: canonical.networkPolicy,
+        commandOptions: agent === 'claude' ? [claudeScopeOption()] : [],
+        // The retained spelling keeps its shipped safety guarantee: profile-only
+        // installation that leaves the repository untouched (main's Rule
+        // codex-plugin-install.TBU1.R2). `install --agents=<agent>` is the
+        // canonical route that also reconciles project configuration.
+        handler: publicHandler(name),
+      },
+    ),
     aliasFor: 'install',
     classification: 'retained-alias',
     visibility: 'hidden',
     compatibility: {
       ...RETAINED_ALIAS,
-      replacement: `install --agents=${agent}`,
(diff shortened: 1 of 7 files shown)


packages/cli/src/cli-protocol/catalog.ts
```diff
@@ -110,22 +110,27 @@ function alias(name: string, aliasFor: string): CommandDefinition {
...
+        handler: publicHandler(name),
+      },
+    ),
     aliasFor: 'install',
     classification: 'retained-alias',
     visibility: 'hidden',
     compatibility: {
       ...RETAINED_ALIAS,
-      replacement: `install --agents=${agent}`,
+      replacement: `install --agents=${agent} (also reconciles the project)`,
```
TheMostlyGreat, line 133: Reviewer: This is the blocking line. `replacement` feeds `target:` for the argv-rewrite machine contract (`catalog.ts:867`), `alias_for` in capability output (`catalog.ts:889`), and the deprecation guidance in `register.ts:146`/`:174` — all of which expect a runnable route.

The BDD lane fails on exactly this: `No retained compatibility route maps claude install to install --agents=claude`.

Suggest reverting to the pure route and leaving the clarification in the `description` above, which already says it:

```suggestion
      replacement: `install --agents=${agent}`,
```

TheMostlyGreat (commented):
Reviewer: The goal is right and the docs prose is a real improvement, but the change puts prose into a machine-readable field, and that is what is failing CI.

## Blocking: `replacement` is a command, not a description

`compatibility.replacement` is consumed as an executable route in at least three places, not just printed:

- `catalog.ts:867` — `target: route.replacement` for every `argv-rewrite` entry in the machine contract.
- `catalog.ts:889` — `alias_for` in the machine-readable capability output.
- `register.ts:146` / `register.ts:174` — the deprecation guidance handed to callers.

After this change those all carry `install --agents=claude (also reconciles the project)`, which is not a runnable command. The help line reads `claude install -> install --agents=claude (also reconciles the project)`, where the parenthetical looks like part of the command to type.

This is not hypothetical — it is why both test jobs are red. The BDD lane asserts the semantic contract rather than the string:

```
AssertionError: No retained compatibility route maps claude install to install --agents=claude
  packages/cli/features/unified-first-time-install.feature:296
AssertionError: No retained compatibility route maps codex install to install --agents=codex
  packages/cli/features/unified-first-time-install.feature:297
```

The two unit tests in this PR were updated to match the new string, so they pass; the acceptance lane was not, and it is the one encoding the actual rule. Updating it to expect the prose would be the wrong fix — it would retire the guarantee that `replacement` names a route you can run.

## Suggested shape

Keep `replacement` as the pure canonical route and let the clarification live where prose belongs — which this PR already does well in two of three places:

- **Keep** the reworded `description` (`catalog.ts:115`). It carries the distinction and is displayed to humans.
- **Keep** both `cli.mdx` prose edits. The expanded paragraph on the legacy route is clearly better than what is there today.
- **Revert** `replacement` to `install --agents=${agent}`, and revert the two test expectations with it. The generated compatibility table then regenerates unchanged, which also drops the column-width churn.

That keeps every behavioral guarantee the PR set out to clarify, and the acceptance lane goes green without being edited.

## Staleness

This branch is 921 commits behind `main` and was authored at v0.75.0; `main` is now v0.79.0. Worth rebasing before another CI run — `catalog.ts` has changed underneath it, and the last CI result here is from Aug 11.

No other issues: the change is string-only, every alias and its behavior is preserved as the description claims, and `visibility`/`classification`/`retention` are all untouched.

PR #2903, opened 2026-08-14
fix(review): do not count zombies as running reviewer processes

TheMostlyGreat:
## What

Reviewer cleanup verified a process group with `kill(-pgid, 0)`, which cannot tell a zombie from a live process. A group holding only zombies still answered "yes", so cleanup reported failure for a tree it had already stopped and `stopReviewerOrThrow` replaced the real outcome with `process_failed`.

Liveness now comes from `/proc`, counting only members whose state is not `Z`. Where `/proc` is unavailable the kill probe stands, so macOS and Windows behaviour is unchanged.

## Why it matters

Wherever PID 1 reaps promptly the window is too small to observe — which is why CI never saw this. It is reproducible wherever PID 1 is the application or a lazy init, i.e. the common `docker run` without `--init`: a reviewer's orphaned grandchildren are reparented to PID 1 and linger as zombies. Measured in one such container at **1–3s against a 500ms cleanup budget**, so every timeout in that window was returned as `process_failed` with "processes could not be stopped".

`kill(2)` says so directly: *"Note that an existing process might be a zombie, a process that has terminated execution, but has not yet been `wait(2)`ed for."*

## Evidence

- Four review timeout tests that failed in a slow-reaping container now pass.
- `parseProcessStat` is exported and unit-tested — the integration case cannot be pinned portably, because an environment that reaps promptly never produces the zombie. Mutating the predicate to count zombies as running fails the timeout test in a slow-reaping container and passes in CI; that asymmetry is stated in the commit message rather than hidden.
- The abandonment contract (a genuinely uncleanable group is still reported, never silently abandoned) is unchanged. Its **assertion is byte-identical**; only its simulation changed, since it faked the liveness probe and the probe no longer decides. It now swallows the group signals so the tree really survives, and kills it for real afterwards.

## Also included

`test(review): prove a successful review round-trips through a real process` — every other real-process test here ends in a timeout, a rejected probe, or an uncleanable tree, so nothing covered the success path that `stopReviewerOnce` also runs on. This does **not** catch a zombie regression, and the commit message says so.

## Risk

Cleanup is safety-relevant: a wrong change leaks reviewer processes. The change errs toward reporting failure — non-`Z` states such as `X`/`x` count as running. PID/PGID reuse could cause a spurious `process_failed`, but `kill(-pgid, 0)` had identical exposure, so that is not a regression.

---


TheMostlyGreat:
## Independent Codex review

Found one actionable issue:

- **[P1] Confirm a negative `/proc` scan before declaring cleanup complete** — `packages/cli/src/review/runtime.ts:637`

  If a live group member forks after `/proc` is listed, or remains alive after its stat entry was already inspected, this non-atomic scan can return `false` even though the process group still exists. `groupIsRunning()` then skips `SIGKILL` and reports successful cleanup while reviewer processes remain alive; a reviewer that handles `SIGTERM` can deliberately trigger this by continuously forking. Treat a negative scan as inconclusive while `kill(-pgid, 0)` still succeeds, or otherwise require a stable scan that distinguishes a zombie-only group without trusting one snapshot.

TheMostlyGreat:
Confirmed and fixed in `d34b6b9`.

The sharp part of this finding is that the scan gates **escalation**, not just the verdict — I had noted the non-atomic scan as a race in the PR description but assessed it as "a slight widening of an existing exposure, since `kill(-pgid, 0)` was also point-in-time." That was wrong in one specific way: the kill probe it replaced **always** escalated at that point, so gating `SIGKILL` on the scan was a regression this PR introduced, not a pre-existing race.

Fix, along the lines you suggested:

- **Escalate on any remaining member, zombie or not.** `SIGKILL` against a zombie-only group is a no-op, so this restores the previous escalation behaviour exactly and leaves only the verdict distinguishing a zombie from a live process. A false negative can no longer cause the kill to be skipped.
- **Confirm a negative with a second scan** before calling the group stopped, so one unlucky snapshot cannot report success over a live tree.

I kept the negative-scan verdict rather than treating it as inconclusive whenever `kill(-pgid, 0)` still succeeds — that condition is true for a zombie-only group, which is precisely the case this PR exists to fix, so it would reinstate the original bug. The scan is the only thing that can tell the two apart; the hardening is unconditional escalation plus confirmation instead.

The zombie behaviour is unchanged: a group holding only zombies is still reported stopped rather than as `process_failed`.

Verified: 231 passed across the review suites, including the abandonment contract (a genuinely uncleanable group is still reported) and the four timeout tests. The one failure in my container is an unrelated uid-0 `chmod` test that is fixed in #2905 and passes in CI.

One limitation worth stating: the escalation change has no direct unit test, because there is no seam to force a false-negative scan. It is a strict superset of the previous escalation — `SIGKILL` is now sent in every case the old code sent it, and in some where the new code would have skipped it.

---

TheMostlyGreat:
## Safeword Quality Review

**Currency:** ✓ Linux `/proc/<pid>/stat` semantics checked against the current Linux manual
**Sources:** ✓ Current Linux process-state documentation and the PR diff
**Correct:** ⚠️ Zombie detection is reasonable on Linux, but the implementation treats any readable `/proc` as Linux-format
**Elegant:** ⚠️ Repeated full `/proc` scans every 5 ms make teardown unnecessarily expensive
**No-bloat:** ✓ The parser and liveness helper are focused
**Wiring:** ❌ Tests cover only the pure stat parser, not a zombie-only group or the non-Linux fallback

**Verdict:** REQUEST CHANGES

**Critical issues:**
- Guard the `/proc/<pid>/stat` parser with `process.platform === "linux"`. A readable, non-Linux `/proc` can otherwise parse as no live members and falsely report cleanup success.
- Prove the motivating behavior through the liveness helper/cleanup boundary; parser-only tests do not show that a zombie-only group is treated as stopped while a live group remains running.

The independent reviewer also found an error-severity coercion hole in `hasValidReviewerOutputBody`, but that code is unchanged from `origin/main` and is not attributed to this PR.

**Suggested improvements:** Avoid scanning every process every 5 ms; use a less aggressive poll or cache the platform/availability decision.

**Provenance:**
- (verified: https://man7.org/linux/man-pages/man5/proc_pid_stat.5.html) — current Linux stat-field and process-state documentation, fetched this session
- (verified: PR diff against `origin/main`) — `/proc` scanning and its parser tests are introduced here

**Next:** Add the Linux guard and real liveness-boundary tests, then rerun Safeword `quality-review`.

_Coverage: Safeword `quality-review`; independent Claude Opus reviewer in a separate headless process; generated `plugin/runtime/cli.js` exceeded the 256 KiB packet limit and remains covered by build/parity checks; required CI checks pass._

TheMostlyGreat:
Both critical issues addressed in `a9938c1`.

**Linux guard** — agreed, and this was the one real correctness bug in the set. `procGroupHasRunningMember` now returns `undefined` unless `process.platform === 'linux'`, so a readable-but-differently-shaped `/proc` (Solaris, some BSDs) falls back to the kill probe instead of parsing every entry into "no live members" and calling a running group stopped.

**Liveness-boundary tests** — also agreed; the parser-only coverage was the weak point. Added three cases against a real process group: a live detached group reads as running, a reaped group reads as not running, and the off-Linux contract (skipped on Linux, where the first two already prove a definite answer is returned). These have teeth — a helper hardcoded to `return false` passes every parser case and fails the live-group case.

**Efficiency** — taken, and it turned out free rather than a trade-off. `groupIsRunning` now asks the kill probe first: one syscall versus a walk of every process, and decisive on its own when the answer is no, since an empty group holds nothing zombie or otherwise. The common case — a reviewer that exits promptly — no longer scans `/proc` at all. Semantics are unchanged in all four cases (gone / live / zombie-only / `/proc` unavailable).

One limitation still stands, unchanged from the PR description: **the zombie-only case is not pinned portably.** An environment that reaps promptly never produces the zombie, so that case is proven by the timeout tests only where reaping is slow. I would rather leave that stated than add a test that passes everywhere for the wrong reason.

The `hasValidReviewerOutputBody` coercion hole you flagged is unchanged from `origin/main` — agreed it is not attributable here, and I have not touched it.

---

TheMostlyGreat:
## Quality Review

**Verdict:** REQUEST CHANGES

**Critical issue:** `hasValidReviewerOutputBody` coerces severity with `String(...)`. A value such as `["error"]` validates, then evades the strict error-finding invariant. Require an actual string and add the non-string regression case.

**Provenance:** Current head `1bc8019ba`; independent review `a6a108d9-bb64-4ac3-b6ba-b93593fd8bc2`, fetched this session.

**Next:** Tighten severity validation, add the regression test, and rerun quality review.

TheMostlyGreat:
Confirmed and fixed in `9290b12ec` (rebased onto the main catch-up at `1bc8019ba`).

The hole is worse than a lax type check, because the same field was read two different ways:

- `hasValidReviewerOutputBody` coerced with `String(finding.severity)` before testing the allowed union, so `severity: ["error"]` validated.
- `reviewerVerdictMatchesFindings` compares the **raw** value against `'error'`, and `["error"] !== 'error'`.

So a reviewer could return `verdict: "approve"` while carrying an error finding — precisely the invariant that check exists to enforce. Severity is required to be a string now, and both readings agree.

Two regression cases, both RED against the previous validator:

```
× rejects structurally invalid output: non-string severity
× rejects structurally invalid output: approval with an error finding wrapped in an array
```

`tests/review/runtime.test.ts`: 32 passed, 1 skipped. Lint and typecheck clean; plugin runtime regenerated with the pinned Bun 1.3.14.

Scope note: the earlier review pass flagged this as pre-existing on `main` and did not attribute it to this PR, which is accurate — it is not a regression from the zombie-cleanup change. I fixed it here anyway because it is a two-line correctness fix in a file this PR already touches, and leaving a known trust-boundary hole open to file separately seemed worse than carrying it. Happy to split it out if you would rather keep this PR to the cleanup change.

---

TheMostlyGreat:
Withdrawing the offer to split the severity fix out — it stays on this PR. No action needed on that question.

---

TheMostlyGreat:
CI is red on `4d6af23d8` — one BDD scenario. It is not from my commits, and it is not a flake. Bisected to a single commit and a single line.

**Failing:** `Presentation rejects inconsistent policy and status as completed coverage`, example row `invalid-top-state` (`clarify-review-coverage.feature:83`). 1459/1460 scenarios pass.

**Culprit:** `3fd7a2f4b` *fix(review): clarify pending and trusted cleanup*.

```
ddcb69445 (my last head)            24 scenarios (24 passed)
4fff406a6 (parent of the suspect)   24 scenarios (24 passed)
3fd7a2f4b                           24 scenarios (23 passed, 1 failed)   ← introduced here
origin/main                         24 scenarios (24 passed)
```

**Line:** `packages/cli/src/cli-protocol/review-presentation.ts`

```ts
if (result.state === 'failed' || !isRecord(result.data) || result.data.command !== 'review run')
  return undefined;
```

The `invalid-top-state` factory is `makeReviewResult({ status: 'approved', state: 'failed' })` — an *inconsistent* tuple, and the scenario exists to assert it presents as `Review incomplete` rather than as completed coverage. The new early return suppresses review lines entirely whenever `state === 'failed'`, so nothing is rendered and the rejection disappears.

Two contracts are in conflict, and which one wins is a judgment call for whoever wrote `3fd7a2f4b`:

1. A genuinely failed result should render no review lines (the commit's intent).
2. An `approved` status carried with a `failed` state is inconsistent and must be *called* inconsistent (`SWM1.R2`).

They can coexist — gate the early return on the result being *consistently* failed, rather than on `state === 'failed'` alone — but picking that resolution is not mine to make, and this branch has several people pushing to it right now, so I am not going to push over in-flight work.

My own commits on this branch (the zombie cleanup and the severity fix) are unaffected: `tests/review/runtime.test.ts` is 32 passed / 1 skipped at the current head.

Repro without waiting on CI — the `BeforeAll` build exceeds cucumber's 5s default locally, so raise it first:

```
cd packages/cli
printf 'import { setDefaultTimeout } from "@cucumber/cucumber";\nsetDefaultTimeout(120000);\n' > features/steps/zz-local-timeout.ts
NODE_OPTIONS='--import tsx' npx cucumber-js --tags '@clarify-review-coverage.SWM1.R2'
rm features/steps/zz-local-timeout.ts
```

Happy to push the fix if you want me to pick option 1 or 2 — say which.

---

PR #2904, opened 2026-08-14
fix(relay): stop a failed startup leaving its maintenance interval running

TheMostlyGreat:
## What

`startRelayServer` creates the maintenance interval before it listens, and left cleanup to the server's `'close'` handler. **A failed `listen` never emits `'close'`**, so the listen-error path released the process lock and returned while the interval kept running.

That interval calls `service.maintain()` against the real store and GitHub client. It went on sweeping on behalf of a server the caller had been told did not start, and every retried `startRelayServer` added another. `unref()` keeps it from holding the process open; it does not stop it firing.

## Fix

One idempotent release that clears the interval and frees the lock, wired to `'close'` and called from both post-listen failure paths, rather than releasing the lock piecemeal and hoping `'close'` arrives.

This also removes the double release on the unbound-address path, where `server.close()` already triggers the same handler — previously harmless only because `ProcessLock.release()` happens to be idempotent, and now not dependent on that.

The `recoverInFlight` failure above still releases only the lock: it runs before the interval exists, so there is nothing else to clean.

## Evidence

Written test-first. With a port already taken, startup rejects and the maintenance sweep count stops moving. Against the previous code the same test records **eight further sweeps after the rejection**.

Verified separately that `'close'` does not fire after a failed listen:

```
listen error: EADDRINUSE
'close' fired after failed listen? false
```

Relay suite: 171 passed / 1 skipped.

## Provenance

Found by an independent review of an unrelated refactor branch, which read the whole file rather than the diff. The reviewer also flagged a possible double release *"if `ProcessLock.release()` isn't idempotent"* — it is (`if (this.#released) return;`), so that half was not a live problem and is not claimed here.

---

### DIFF
--- a/packages/retro-relay/src/http-server.ts
+++ b/packages/retro-relay/src/http-server.ts
@@ -219,10 +219,24 @@ export async function startRelayServer(input: RelayServerOptions): Promise<{
           void maintain();
         }, input.maintenanceIntervalMs ?? DEFAULT_MAINTENANCE_INTERVAL_MS);
   maintenanceTimer?.unref();
-  server.once('close', () => {
+  /**
+   * Releases everything startup acquired, once.
+   *
+   * Startup can fail after the maintenance interval exists, and a failed
+   * `listen` never emits 'close' — so leaving cleanup to the close handler
+   * alone leaves that interval sweeping the real store and GitHub client on
+   * behalf of a server the caller was told did not start, with another added
+   * per retry. Every failure path calls this rather than releasing the lock
+   * piecemeal and hoping 'close' arrives.
+   */
+  let released = false;
+  const releaseResources = (): void => {
+    if (released) return;
+    released = true;
     if (maintenanceTimer !== undefined) clearInterval(maintenanceTimer);
     if (ownsProcessLock) processLock?.release();
-  });
+  };
+  server.once('close', releaseResources);
 
   const recordReconciliationOutcome = (entry: {
     receiptId: string;
@@ -386,13 +400,13 @@ export async function startRelayServer(input: RelayServerOptions): Promise<{
       server.listen(input.port ?? 0, input.host ?? '127.0.0.1', resolve);
     });
   } catch (error) {
-    if (ownsProcessLock) processLock?.release();
+    releaseResources();
     throw error;
   }
   const address = server.address();
   if (address === null || typeof address === 'string') {
     server.close();
-    if (ownsProcessLock) processLock?.release();
+    releaseResources();
     throw new Error('relay did not bind a TCP address');
   }
   return {
--- a/packages/retro-relay/src/http-server.ts
+++ b/packages/retro-relay/src/http-server.ts
@@ -408,6 +408,9 @@ export async function startRelayServer(input: RelayServerOptions): Promise<{
       server.once('listening', onListening);
       server.listen(input.port ?? 0, input.host ?? '127.0.0.1');
     });
+    server.on('error', error => {
+      observability.logs.push({ event: 'retro_server_error', message: error.message });
+    });
   } catch (error) {
     releaseResources();
     throw error;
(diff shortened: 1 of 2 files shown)


TheMostlyGreat:
## Independent Codex review

The change consistently clears the maintenance interval and releases the owned process lock on startup failure while preserving normal close cleanup through an idempotent release function. No actionable regressions were identified.

_Verification note: the reviewer attempted the targeted test, but this isolated worktree did not contain installed dependencies (`tsup` was unavailable). CI is green on the PR._

TheMostlyGreat:
## Safeword Quality Review

**Currency:** ✓ Node server lifecycle checked against current Node documentation
**Sources:** ✓ Current Node HTTP documentation and the PR diff
**Correct:** ❌ The cleanup implementation is directionally sound, but its new regression test does not prove the claimed behavior
**Elegant:** ✓ The idempotent `releaseResources` helper is a clear cleanup shape
**No-bloat:** ✓ The implementation is focused
**Wiring:** ❌ The test bypasses the owned process lock and lacks a positive control for its interval probe

**Verdict:** REQUEST CHANGES

**Critical issues:**
- `startup-cleanup.test.ts` counts `pendingAlerts()` only after maintenance succeeds, while its GitHub token supplier always rejects. With no positive control, the counter can remain zero whether or not the leaked interval is cleared.
- The test uses `allowUnlockedForTests: true`, making the owned-lock release path—the other behavior this PR claims to fix—dead code under test.

**Suggested improvements:** Add a live-server positive control proving the sweep counter advances, and a `lockPath` case proving a failed bind releases the lock so it can be reacquired.

**Provenance:**
- (verified: https://nodejs.org/api/http.html) — current Node `Server` close/lifecycle documentation, fetched this session
- (verified: PR diff against `origin/main`) — both gaps are in the newly added test

**Next:** Strengthen the two regression proofs, then rerun Safeword `quality-review`.

TheMostlyGreat:
Both review points were right; fixed in `60e24b4ce`.

**The sweep counter was unpinned.** I checked whether `pendingAlerts()` is even reachable in the failed-bind fixture — it is: `service.maintain()` swallows per-record errors and `claimDueRetries()` returns `[]` on an empty store, so nothing throws before the alert loop and the rejecting `installationToken` is never reached. So the original RED measurement was real. But you're right that nothing in the test held it there. Added a positive control that starts a live server and requires the counter to advance.

Verified by mutation rather than assertion. Replacing the alert loop's iterable with an empty array leaves the interval-leak test **green** and fails the control:

```
FAIL  relay maintenance interval > advances the sweep counter while the server is up
AssertionError: expected 0 to be greater than 0
Tests  1 failed | 2 passed (3)
```

That is exactly the silent-vacuity mode you named.

**The lock-release path was dead.** `allowUnlockedForTests: true` means `ownsProcessLock` is false, so `processLock?.release()` never ran under test. Added a `lockPath` case that reacquires the same path after a rejected bind. Dropping the release from the failure path fails it:

```
FAIL  relay startup failure > releases the process lock it acquired when the port is taken
Error: retro relay database is already locked
  at ProcessLock.acquire (src/process-lock.ts:14:48)
```

Full `packages/retro-relay` suite: 173 passed, 1 skipped. No production code changed.

---

TheMostlyGreat:
## Quality Review

**Verdict:** REQUEST CHANGES

**Critical issue:** Reacquiring the same lock after failed startup proves release only if same-process reacquisition fails while the first lock is held. Add that negative control or assert the lock state directly.

**Provenance:** Current head `60e24b4ce`; independent review `04a24e37-c601-471a-8c70-09d6ff8533a9`, fetched this session.

**Next:** Add the lock-semantics control, rerun the focused suite, and rerun quality review.

TheMostlyGreat:
Fair — added in `9ebc05d74`.

The reacquire proves a release only if acquiring while held would have failed, and nothing in the test said so. The control now acquires the lock, shows a second acquire on the same path throws `already locked`, releases, and shows the path is usable again — so the failure is the lock being *held*, not the path being unusable in the first place.

`packages/retro-relay`: 174 passed, 1 skipped. Typecheck and lint clean.

That is the same shape as the sweep-counter control from the last round: an assertion that something happened is worth what its negative case is worth.

---

PR #2905, opened 2026-08-14
test: make environment-dependent tests hold outside CI, and guard the rule

TheMostlyGreat:
## What

A class of test in this suite passes in CI and fails everywhere agents actually run. Three independent instances, plus the guards to stop them recurring.

### 1. uid — `chmod` simulates nothing under root

Root holds `CAP_DAC_OVERRIDE` and bypasses every permission bit, so a `chmod`-based failure simulation silently does not happen as uid 0: the code takes the success path and the test fails asserting on that output. GitHub Actions runs as `runner` (uid 1001), so these are green in CI and red in every root container — agent sessions, devcontainers, plain `docker run`.

`tests/helpers/io-failure.ts` induces failure through filesystem *structure* instead, which no uid can override:

| Helper | Effect | errno |
|---|---|---|
| `sinkWrites` | writes land, reads come back empty | — (`/dev/null`) |
| `blockWrites` | writing this exact path fails | `EISDIR` |
| `blockChildren` | creating anything under this dir fails | `ENOTDIR` |
| `blockScan` | scanning into this dir fails | `ELOOP` |

Eight sites converted. The `architecture-stage` git clean-filter needed care: the failure must happen *mid-`git add`*, after git has read the content, so it swaps the document for a directory at that moment rather than chmod-ing the parent.

### 2. Clone shape — a documented prerequisite that never said its name

Four Claude-migration suites read real released bytes with `git show v<version>:<path>`, and the sealed-review suite walks a path's ancestry. Both need history a shallow or tagless clone lacks. **CI already pays for this** — three `ci.yml` jobs set `fetch-depth: 0` with comments naming these suites — so the dependency is deliberate, and reading from a tag is the point: a tag is an immutable record of what shipped, which a vendored fixture could not prove.

What was missing was any statement of it. A shallow clone produced 32 bare `fatal: invalid object name 'v0.68.0'` failures across five files, which reads as a broken suite. `tests/helpers/git-history.ts` now fails once per suite with the remedy.

### 3. Ambient env — fixtures presenting two host identities

Three closeout host-adapter tests build a Codex host by spreading `process.env`, then adding `CODEX_HOME`/`CODEX_THREAD_ID`. Run identity resolves Claude **before** Codex on purpose (`run-identity.ts`: *"Keep this after Claude detection so an explicit Claude environment never adopts a Codex runtime accidentally"*), so a suite running inside a Claude Code session also handed the child its own `CLAUDE_CODE_SESSION_ID`. The production precedence is correct; the fixtures were presenting two identities and expecting the second to win.

Verified 25/25 both inside a Claude Code session and with the agent variables unset, so neither environment is now required.

## The guard

`uid-dependent-permission-tripwire.test.ts` fails on a permission-removing simulation. Review found four ways past the first version — all reproduced, all now pinned as cases:

- a line holding a URL lost its call to naive `//` stripping
- a Prettier-wrapped call with a trailing comma did not match
- a path argument with its own comma defeated the `[^,]+` match
- only `0o___` counted, so `chmodSync(p, 0)` and `'000'` passed straight through

Detection moved to `helpers/permission-simulation.ts` and stops pattern-matching source. Note the obvious repair — mask strings first — **is also wrong**: a regex literal containing a quote starts a "string" running to the next quote anywhere in the file, and that module's own source contains one. It scans left to right instead.

**The rule is also corrected.** Turning the stronger detector on the suite surfaced two calls in `setup-hooks` and `setup-linting` already handled the established way — `it.skipIf(process.getuid?.() === 0)` with a comment saying why. Those are not violations, so the guard waives a call a root guard covers. The rule is *"never let a chmod stand in for a failure root will not produce"*, not *"never chmod"*.

## Evidence

- `helpers/io-failure.test.ts` asserts each helper raises the errno it documents against the real filesystem, plus the already-exists case. Reverting the `blockChildren` fix turns the matching case red.
- Planting an unguarded `chmodSync(p, 0)` trips the tripwire; the two waived tests still skip as root rather than being flagged (167 passed / 2 skipped).
- Full CLI suite went from 43 failures to 4, and the remaining 4 were an unrelated product bug now fixed in #2903.

---

### DIFF
--- a/.project/tickets/93C14D-close-completed-sessions-safely/manual-review-request.json
+++ b/.project/tickets/93C14D-close-completed-sessions-safely/manual-review-request.json
@@ -92,24 +92,12 @@
       "sha256": "b2b6ed4428c784e2086fc2f7e9e6f95dbc2890c881f198a3ebf122026fed3962"
     },
     {
-      "path": "plugin/runtime/cli.js",
-      "sha256": "199147462ab7a94473497ae3bb38daecf5ca8b9a1cf5212bb27d133e539ff2f2"
-    },
-    {
       "path": "plugin/runtime/hooks/lib/closeout-binding.ts",
       "sha256": "4d89818ea62ee89d894a56a1947e05d1ee9edbf8eff45882a2c3adc2a589164e"
     },
     {
       "path": "plugin/runtime/hooks/lib/retro-extract.ts",
       "sha256": "390086ac3d80bbce4787aae788f6fb788adbbd383f2134964bc5bb13b877d45e"
-    },
-    {
-      "path": "plugin/identity.json",
-      "sha256": "7676ee1f028e7af127a8a4a4333ca9694acd27f7147d7a34e19827bd2735c6f8"
-    },
-    {
-      "path": "plugin/inventory.json",
-      "sha256": "77bcc3f808b9caf85f1661f6ae3f07ee0e31c541c0e17dd5b75008b076ed3f12"
     },
     {
       "path": "packages/cli/tests/closeout-skill.test.ts",
--- a/.project/tickets/93C14D-close-completed-sessions-safely/manual-review.md
+++ b/.project/tickets/93C14D-close-completed-sessions-safely/manual-review.md
@@ -1,7 +1,7 @@
 # Manual Review
 
 Fresh review is bound to the post-merge manifest. The author-side orchestrator
-recomputed 39/39 input hashes and 58/58 ordered scenario titles; the independent
+recomputed 36/36 input hashes and 58/58 ordered scenario titles; the independent
 reviewer read every bound input and evaluated every scenario without relying on
 the prior review verdict.
 
@@ -18,7 +18,7 @@ check and rejects any mismatch.
     "identity": "claude-headless:ec6efe37-35d9-4dcb-9774-2118ef47d1e7",
     "model": "claude-sonnet-5 (headless Claude Code 2.1.220)"
   },
-  "manifest_sha256": "e945fcb6eca6f238d154da6f96199517aa578e89b15a619ee6ef8ff533443d61",
+  "manifest_sha256": "fc3e1743b894a0647c7c78e1055e9a79cbce73b21303b9587457dae2b9a10c2f",
   "verdicts": [
     { "id": "01", "verdict": "pass" },
     { "id": "02", "verdict": "pass" },
--- a/packages/cli/tests/claude-plugin/automatic-migration.test.ts
+++ b/packages/cli/tests/claude-plugin/automatic-migration.test.ts
@@ -12,7 +12,7 @@ import {
 import { tmpdir } from 'node:os';
 import nodePath from 'node:path';
 
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeAll, describe, expect, it } from 'vitest';
 
 import { migrateClaudeLegacyAutomatically } from '../../src/claude-plugin/cleanup.js';
 import { CLAUDE_HISTORICAL_CATALOGUE } from '../../src/claude-plugin/historical-catalogue.generated.js';
@@ -21,6 +21,7 @@ import {
   historicalHookEntry,
 } from '../../src/claude-plugin/historical-ownership.js';
 import { readClaudePluginMode } from '../../src/claude-plugin/migration-state.js';
+import { requireHistoricalReleaseTags } from '../helpers/git-history.js';
 
 const repoRoot = new URL('../../../..', import.meta.url).pathname;
 const roots: string[] = [];
@@ -74,8 +75,15 @@ afterEach(() => {
   roots.length = 0;
 });
 
+/** Releases this suite reads real bytes from; shared with the history preflight. */
+const FIXTURE_VERSIONS = ['0.68.0', '0.69.0', '0.72.0'];
+
 describe('automatic Claude migration', () => {
-  it.each(['0.68.0', '0.69.0', '0.72.0'])(
+  beforeAll(() => {
+    requireHistoricalReleaseTags(FIXTURE_VERSIONS);
+  });
+
+  it.each(FIXTURE_VERSIONS)(
     'contracts exact %s released bytes and writes clean plugin mode silently',
     version => {
       const { root, installedPath } = fixture(version);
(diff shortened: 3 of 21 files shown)


TheMostlyGreat:
## Independent Codex review

The changes consistently replace UID-dependent permission simulations with structural filesystem failures and add appropriate history preflights. No actionable correctness regressions were identified.

_Verification note: the reviewer attempted targeted tests, but this isolated worktree did not contain installed dependencies (`tsup` was unavailable). CI is green on the PR._

TheMostlyGreat:
## Safeword Quality Review

**Currency:** ✓ Node filesystem behavior checked against current Node documentation
**Sources:** ✓ Current Node filesystem documentation and the PR diff
**Correct:** ⚠️ Structural I/O-failure helpers are sound, but the new permission-simulation guard has material blind spots
**Elegant:** ⚠️ The hand-written source scanner is complex relative to its incomplete coverage
**No-bloat:** ✓ The reusable structural-failure helpers remove repeated permission tricks
**Wiring:** ✓ The changed suites exercise real filesystem/git collaborators; required CI checks pass

**Verdict:** REQUEST CHANGES

**Critical issues:**
- `packages/cli/tests/helpers/permission-simulation.ts` only detects literal `chmodSync(...)` and a narrow shell form. Async/aliased calls, `execFileSync("chmod", ...)`, and common `chmod -R`/symbolic-mode forms evade a guard whose purpose is to prevent this entire regression class.
- Its regex-literal scanner can desynchronize after common contexts such as `return /.../` or `=> /.../`, potentially hiding a later violation.

The reviewer also found a vacuous assertion and a risky cleanup elsewhere in changed files, but their problematic lines predate this PR, so they are not classified as introduced defects here.

**Suggested improvements:** Expand the detector and add explicit evasion tests, or narrow its stated guarantee so a green tripwire is not over-trusted.

**Provenance:**
- (verified: https://nodejs.org/api/fs.html) — current Node filesystem API behavior, fetched this session
- (verified: PR diff against `origin/main`) — detector implementation is newly introduced here

**Next:** Cover the reachable detector evasions and regex contexts, then rerun Safeword `quality-review`.

TheMostlyGreat:
First point confirmed and fixed in `cbd704794`; second point does not reproduce, and there are now cases pinning why.

## The blind spots were real — all four

None of them is an evasion attempt. They are what someone writes when the situation calls for it, which is exactly the class this guard exists to catch.

| Form | Why it got past |
|---|---|
| `chmod -R a-w dir` | the pattern expected the mode immediately after the word |
| `chmod u=r file` | symbolic assignment was not handled at all |
| `execFileSync('chmod', ['000', p])` | the mode never sits beside the word, and no fs binding is touched |
| `fs.promises.chmod` / `import { chmodSync as lockDown }` | only the literal name `chmodSync` was known |

Each is a RED-first case; all five failed before the fix. Shell modes are now classified by the bits they leave rather than by a numeric pattern, so `chmod 0000` and `chmod u=r` land as the same finding, and the `=` form is judged by what it **omits** — `u=r` never mentions write and takes it away regardless.

End-to-end through the suite scan, not just the unit entry point — a planted file reports:

```
+   "zz-detector-probe.ts: lockDown(…, 0o000)",
+   "zz-detector-probe.ts: chmod -R a-w",
+   "zz-detector-probe.ts: chmod u=r",
```

Four matching allow-cases keep the widening from flagging grants (`chmod -R u+w`, `u=rw`, `execFileSync('chmod', ['755', …])`, `fs.promises.chmod(p, 0o755)`). An over-eager tripwire gets waived, and then it guards nothing.

## The desync claim does not hold

You are right that the scanner mis-reads both contexts — neither `n` (from `return`) nor `>` (from `=>`) opens a regex context, so the `/` reads as division and a quote inside the literal starts a "string" that runs on. But it cannot hide a violation, because **only comment spans are ever removed**. Every other byte is emitted verbatim, so a mis-tracked literal leaves the later call sitting in the output for the scan to find. The failure direction is a spurious report, never a false clean.

Three cases now pin that, including the sharpest form (a regex holding both quote characters, followed by a comment, followed by a violation) — all green.

## Stated guarantee narrowed

Per your suggested improvement, the module now says what a green result covers and what it does not: a mode held in a variable is not evaluable from source and is skipped rather than guessed, and a callee named at runtime is not resolvable. Both are reachable by someone working around the guard; neither is a shape anyone writes by accident.

## Suite

`tests/uid-dependent-permission-tripwire.test.ts`: 33 passed. Full CLI suite: 7891 passed, 5 failed — all five reproduce with my changes stashed at `81e89c9b5`. Four are the review-timeout product bug fixed in #2903; the fifth (`runs the real packaged CLI from the isolated snapshot`) needs `SAFEWORD_TEST_CLI_ROOT`, which a bare `vitest run` does not set.

---

TheMostlyGreat:
## Quality Review

**Verdict:** REQUEST CHANGES

**Critical issues:**
- Regex-context desynchronization can cause a real `chmodSync(..., 0o000)` call after a URL to be stripped as a comment.
- `chmodModeArguments` checks only the final argument, missing callback-form `fs.chmod(path, 0o000, callback)`.

**Provenance:** Current head `cbd704794`; independent review `25127c68-75e2-4e11-b734-b0d1ad91ce82`, fetched this session.

**Next:** Add both false-clean regressions, fix the scanner/argument extraction, and rerun quality review.

TheMostlyGreat:
Both confirmed, both fixed in `19d9755f1`. My earlier reply on the desync was wrong and I'll say where.

## The desync does hide a violation

I argued it could not, because only comment spans are removed. That reasoning has a hole: **stripping a comment deletes code**, so a mis-tracked literal that leads the scan into a false comment is a false clean, not noise. Your version reaches it in two steps:

```js
const f = () => /(["'])/u;
const u = "https://x.com"; chmodSync(p, 0o000);
```

`>` does not open a regex context, so the `/` reads as division; the `"` inside the literal opens a "string" that runs to the **opening** quote of the URL; the scan resumes inside `https://x.com`, sees `//`, and takes the rest of the line — call included. Reported clean. My earlier cases all put a newline between the literal and the call, which is why they passed and why they proved less than I claimed.

Two fixes, because one is not enough:

- `>` and the keywords a regex may follow (`return`, `typeof`, `case`, `await`, …) now open a regex context. That covers the two forms you named.
- `)` **cannot** join them — `(a + b) / c` is division and far more common than a regex there — so `if (x) /(["'])/u.test(s)` stays mis-read. A `//` directly after a colon is now treated as a URL scheme rather than a comment, which holds however the scan arrived there.

Each is pinned by exactly one case; removing either fails exactly that one and nothing else.

## Callback form

Right — the mode was read as the *last* argument, so `chmod(path, 0o000, callback)` reported the callback as its mode and evaluated to nothing. It is read by position now: every fs chmod takes the mode second. The existing trailing-comma and comma-in-path cases still pass, since both leave the mode at index 1.

## Suite

`tests/uid-dependent-permission-tripwire.test.ts`: 36 passed. Lint and typecheck clean. Both regressions were RED first — reproduced against `cbd704794` before either fix.

---

PR #2906, opened 2026-08-14
refactor: collapse the duplication in the relay HTTP router

TheMostlyGreat:
## What

Five behaviour-preserving steps on `packages/retro-relay/src/http-server.ts`. One file, no behaviour change.

| Commit | Smell | Refactoring |
|---|---|---|
| `2c21706` | Duplicated observability writes (4 sites) | Extract `recordReconciliationOutcome` |
| `6b22afb` | `/reconcile` and `/recover` bodies structurally identical | Extract `respondWithReconciliation` |
| `0c865db` | 37-line submit route inline in the router | Extract `respondWithSubmission` |
| `a3fe0d0` | Regex + exec + method check + decode repeated 3× | Extract `matchReceiptRoute`, name the patterns |
| `aecf138` | Unexplained numeric defaults | Name 5 constants |

`handle` goes from 165 lines to 73. Net diff is **+16 lines** — the extracted helpers cost signature boilerplate, which is why that is stated rather than claimed as a shrink.

## Behaviour preservation

Each step was committed only after its own green run of the relay suite (170 passed / 1 skipped throughout).

Two things worth a reviewer's attention:

- `matchReceiptRoute` now skips the regex `exec` when the method does not match, which the original always ran. Safe **only** because none of the three patterns carries the `g` flag — no `lastIndex` state to disturb. All three are `/…/u`.
- Malformed-path cases were traced through both versions — `GET /v1/retro-filings/%/reconcile`, `POST /v1/retro-filings/%` — with identical outcomes, including which paths decode and which 404 first.

## Wiring

`relay.integration.test.ts` exercises submit / reconcile / recover / status / health against a **real SQLite `RelayStore`**, mocking only the GitHub REST boundary.

## Known, deliberately unaddressed

The `complexity` / `cognitive-complexity` suppression on `handle` stays. I attempted removal twice: complexity fell 27 → 21 (cognitive 22 → 19) against a ceiling of 10. Closing that gap needs a route-table redesign, which contradicts the suppression's own stated rationale — *"a single composition-root router keeps the public contract visible."* That is a design call, not a refactor, so both attempts were reverted rather than forced.

An independent review of this file also surfaced a **pre-existing** startup-cleanup bug that this branch neither introduced nor touches. It is fixed separately in #2904 rather than folded in here, since a behaviour change does not belong in a behaviour-preserving branch.

---

### DIFF
--- a/packages/retro-relay/src/http-server.ts
+++ b/packages/retro-relay/src/http-server.ts
@@ -197,6 +197,25 @@ export async function startRelayServer(input: RelayServerOptions): Promise<{
     if (ownsProcessLock) processLock?.release();
   });
 
+  const recordReconciliationOutcome = (entry: {
+    receiptId: string;
+    subject: string;
+    disposition: string;
+    alert?: boolean;
+  }): void => {
+    observability.logs.push({
+      event: 'retro_reconciliation',
+      receiptId: entry.receiptId,
+      subject: entry.subject,
+      disposition: entry.disposition,
+      ...(entry.alert === true && { alert: true }),
+    });
+    observability.metrics.push({
+      metric: 'retro_reconciliation_outcome',
+      disposition: entry.disposition,
+    });
+  };
+
   // eslint-disable-next-line complexity, sonarjs/cognitive-complexity -- A single composition-root router keeps the public contract visible.
   async function handle(request: IncomingMessage, response: ServerResponse): Promise<void> {
     try {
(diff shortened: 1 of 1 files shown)


TheMostlyGreat:
## Independent Codex review

The refactoring preserves the existing routing, response, observability, and error-handling behavior. No actionable defects were identified.

_Verification note: the reviewer attempted the targeted tests, but this isolated worktree did not contain installed dependencies (`tsup` was unavailable). CI is green on the PR._

TheMostlyGreat:
## Safeword Quality Review

**Currency:** ✓ Node HTTP/timer APIs checked against current Node documentation
**Sources:** ✓ Current Node HTTP documentation and the PR diff
**Correct:** ✓ The refactor preserves the existing route, response, and cleanup behavior
**Elegant:** ✓ Helpers reduce router duplication without changing its contract
**No-bloat:** ✓ Constants and two focused helpers are proportionate
**Wiring:** ✓ Existing required CI checks pass; this PR adds no entry point

**Verdict:** APPROVE

**Critical issues:** None introduced by this PR. The independent whole-file pass found broader relay concerns, including unbounded observability buffers and an existing double-release path, but `origin/main` contains those same behaviors; this refactor neither introduces nor worsens them.

**Suggested improvements:** Track the broader relay findings separately rather than charging them to this refactor.

**Provenance:**
- (verified: https://nodejs.org/api/http.html) — current Node server lifecycle documentation, fetched this session
- (verified: PR diff against `origin/main`) — every blocking whole-file observation was checked for introduction by this PR

**Next:** Proceed with review/merge; file the pre-existing relay findings separately if they are not already tracked.

PR #3052, opened 2026-08-16
fix(review): allow configuring Claude effort

### LINKED ISSUE #3051: Claude review coordinator does not expose reviewer effort
**Category:** gap
**Safeword surface:** `hooks/run-review.ts`

**What happened**

Safeword launched independent Claude reviews without an effort limit. Both the default primary model and an explicitly selected alternate model exhausted a 240-second review timeout on the same bounded packet, while the alternate model completed the identical structured review in 45 seconds when invoked directly with low effort.

**Why it is friction**

The official independent-review route timed out and degraded to a non-independent fallback, so Safeword could not produce authoritative cross-agent provenance or advance the gated workflow even though the reviewer service, authentication, schema mode, and packet size were healthy.

**Repro**

Run `safeword review run plan-implementation <accepted-targets> --agent-handoff --json` with `SAFEWORD_REVIEW_TIMEOUT_MS=240000`; observe the Claude route time out. Repeat with `SAFEWORD_REVIEW_PRIMARY_MODEL_CLAUDE=sonnet` and observe the same timeout. Invoke the equivalent bounded Claude review directly with `--effort low`; it completes successfully, but Safeword provides no corresponding reviewer-effort option.

<sub>Auto-filed by `safeword retro run` from a session transcript; sanitized at egress (no customer data).</sub>

TheMostlyGreat:
Closes #3051.

Adds a validated `SAFEWORD_REVIEW_EFFORT_CLAUDE` setting and passes it to Claude review processes only. This lets large review packets use a bounded reasoning effort without changing the default behavior.

Verification:
- `bun run test tests/review/runtime.test.ts --reporter=dot`
- `bun run --cwd packages/cli typecheck`
- targeted ESLint


packages/cli/src/review/runtime.ts
```diff
@@ -54,6 +54,12 @@ const REVIEW_OUTPUT_SCHEMA_SHAPE = {
 } as const;
 
 const REVIEW_OUTPUT_SCHEMA = JSON.stringify(REVIEW_OUTPUT_SCHEMA_SHAPE);
+const CLAUDE_EFFORT_LEVELS = new Set(['low', 'medium', 'high', 'xhigh', 'max']);
+
+function configuredClaudeEffort(environment: NodeJS.ProcessEnv): string | undefined {
```
TheMostlyGreat: Reviewer: Minor consistency nit, not blocking. This file already has an established convention for env parameters — `reviewRunCeiling` (line 213) and three others use `Readonly<Record<string, string | undefined>>`. This introduces a second spelling, `NodeJS.ProcessEnv`, for the same concept at lines 59 and 105.

Both typecheck, so this is purely about the file reading as one voice. Matching the existing type would also make the parameter structurally read-only, which is what you want for something only ever inspected.

TheMostlyGreat (commented):
Reviewer: Reviewed against current main, including a trial merge. Verdict: sound change, worth merging after a mechanical conflict fix. Two small notes, neither blocking. (`gh` authenticates as the author, so this is event COMMENT and carries no merge block.)

**Verified the load-bearing external claim.** `--effort` is a real flag on the installed Claude CLI (2.1.226): `--effort <level>  Effort level for the current session`. I probed every value in the allow-list — `low`, `medium`, `high`, `xhigh`, `max` all exit 0, and `bogus` produces `Warning: Unknown --effort value ... ignoring it`. The allow-list matches the CLI exactly, so this is correct rather than plausible.

**The conflict is mechanical, not semantic.** The branch is 89 commits behind and shows CONFLICTING, but I trial-merged `origin/main`: only `plugin/identity.json` and `plugin/inventory.json` conflict, and both are generated content-address manifests. `src/review/runtime.ts` and its test **auto-merge cleanly** with #3194's reviewer-staging work that landed tonight — I confirmed both changes coexist in the merged file. Resolve by regenerating rather than picking a side:

```
git merge origin/main
git checkout --ours plugin/identity.json plugin/inventory.json && git add plugin/identity.json plugin/inventory.json
cd packages/cli && bun run generate:claude-plugin && bun run check:claude-plugin
```

Picking a side ships a manifest that does not match the regenerated `cli.js` — that bit #3178 tonight. After regenerating: release contract aligned at 0.78.6, `tests/review/` green at 184 passed / 2 skipped.

**Design I liked:** injecting `environment` as a defaulted parameter keeps the test free of global env mutation, and the allow-list means a hostile or fat-fingered value can never reach argv as a bare token. Scoping to `claude` with an explicit test that Codex never receives `--effort` is the right shape.

**Note 1 — invalid values fail silently.** `configuredClaudeEffort` returns `undefined` for anything off the list, so `SAFEWORD_REVIEW_EFFORT_CLAUDE=hi` yields no effort and no message. The Claude CLI itself warns in this case; Safeword swallows it. Since the whole point is tuning behavior for large packets, a user who typos gets the default and no signal that their setting did nothing. A one-line warning would close that gap — or accept it deliberately, but it is worth being a decision rather than a default.

**Note 2 — no docs, but consistent with convention.** The variable is undocumented; I checked before raising it, and there are ~50 `SAFEWORD_REVIEW_*` variables in `src/`, essentially none documented. So this matches the repo. Flagging only because unlike the `FAKE_*` test hooks this one is a genuine user-facing knob, so it is the sort of thing that earns a line in the README whenever that reference gets written. Not a blocker.

TheMostlyGreat:
Pushed `c0eed0ff1` — resolves the conflict with main and implements the warn behavior from my review note.

**Conflict:** merged `origin/main` (now includes #3194's reviewer staging and #3200's audit gate). Only the two generated manifests conflicted; `runtime.ts` auto-merged cleanly. Resolved by regenerating rather than picking a side — release contract aligned at 0.78.6.

**Behavior change.** A non-empty invalid `SAFEWORD_REVIEW_EFFORT_CLAUDE` now warns instead of vanishing. The reasoning that changed my mind: because we strip the value before spawning, Claude's *own* `Warning: Unknown --effort value` never fires either — so a typo left the user with no signal from either layer. Empty/whitespace still means "unset" and stays quiet, matching how `reviewTimeoutMilliseconds` and `configuredCourtesyWait` treat blank input in this same file.

This follows the config convention npm and Bitcoin both converged on: ignore unknown *keys* quietly, but don't be silent about a known key with a broken *value*. Hard-failing was considered and rejected — that's for required config, and here the fallback is the documented default.

Also applied the type nit: `Readonly<Record<string, string | undefined>>` to match the file's other four env parameters.

**Verification:** 312 passed / 2 skipped across `tests/review/` + `tests/claude-plugin/`; typecheck, eslint, prettier clean.

I mutation-tested the new assertions rather than trusting green. Removing the `warn` call fails exactly 4 tests; removing the empty-guard fails 3. That exercise caught a bug in my own tests — `mockRestore()` sat after the assertion, so a failure threw before restoring and leaked the mock into an unrelated test. Restoration now lives in the shared `afterEach` via `vi.restoreAllMocks()`.

**One hygiene caveat:** the fix is folded into the merge commit rather than sitting in its own. Splitting needed a `git reset --hard` that my permission layer blocked, and I didn't route around it. Full diff is still reviewable; only the commit boundary is muddier than ideal.

PR #3085, opened 2026-08-17
Keep reviews focused on authored changes

TheMostlyGreat:
## Summary

- preserve the CKWE2D behavior specification, scenarios, decision record, and implementation plan
- define an explicit `linguist-generated=true` exclusion rule for independently reviewed artifacts
- require visible exclusion accounting in review packets

## Validation

- `bun packages/cli/src/cli.ts lint-gherkin`
- Markdown formatting and diff checks

## Intentionally draft

This is a planning/BDD package only; production implementation and proof mappings are intentionally not included. Safeword also reports missing early-phase provenance and an unchecked ledger item, which should be resolved before this moves forward.


TheMostlyGreat:
Superseded by #3079, which already merged the CKWE2D artifact set. Closing this duplicate draft without advancing the ticket phase.

PR #2813, opened 2026-08-13
Keep Safeword's repository dependency tree free of the Nano ID vulnerability

TheMostlyGreat:
Closes #2808

## Summary
- move the repository Nano ID override and lockfile to patched `3.3.18`
- add a standalone high-severity dependency audit that gates automatic relay deployment
- harden relay input detection and require all CI gates before automatic deployment
- add real workflow/lockfile contract tests, including incompatible-major detection

## Verification
- exhaustive repository verification before this isolated fix: 7,789 CLI tests passed (5 skipped), 170 relay tests passed (1 skipped), 1,471 Gherkin scenarios passed (3 skipped), all builds/typechecks passed
- focused post-fix contracts: 6/6 passed
- `bun run lint`
- `bun run format:check`
- `bun run typecheck`
- `bun audit --audit-level high` (no vulnerabilities)
- full diff-scoped Safeword audit (healthy)
- independent quality review (approved; suggested durability hardening applied)

This is intentionally separate from bookkeeping PR #2779.

PR #3200, opened 2026-08-19
Block high-severity dependency advisories before they reach production

### LINKED ISSUE #2808: Keep Safeword's repository dependency tree free of the Nano ID vulnerability
## Problem

The canonical repository dependency audit fails on `GHSA-2v37-7h3g-55p8`. The root override pins transitive `nanoid` to `3.3.17`, while the first patched 3.x release is `3.3.18`. Nano ID is not a runtime dependency of the published CLI; this affects Safeword’s development/build dependency tree.

## Expected outcome

- Update only the narrow transitive override and lockfile resolution needed to reach `nanoid@3.3.18`.
- `bun audit` reports no vulnerabilities.
- Add an independent high-severity CI audit that gates automatic production relay deployment.
- Full repository verification remains green.
- Keep this change separate from unrelated closeout/bookkeeping PR #2779.

## Evidence

GitHub advisory: https://github.com/advisories/GHSA-2v37-7h3g-55p8

Discovered by the canonical full verification on 2026-08-13.

TheMostlyGreat:
Replaces #2813, which was overtaken by main. Closes #2808.

## Why

#2813 set a `nanoid` override to patched `3.3.18` to clear a transitive advisory. That fix **already landed on main** on Aug 15 via #2262 — main carries the identical override and `bun audit --audit-level high` reports no vulnerabilities. #2813 is now 617 commits behind and would regress the CLI version (0.78.6 → 0.78.0) and the BDD lane scripts if merged.

What #2813 had that main still lacks is the *durable* part: a CI gate so the next advisory is caught automatically. This PR salvages just that.

## What this does

- Adds a `dependency-audit` job running `bun audit --audit-level high`.
- Gates `deploy-retro-relay` on it, so a high or critical advisory can't ride out to production. The environment-protected deploy workflow remains the manual break-glass path.

Scoped to high/critical deliberately: an unpatchable development-only advisory shouldn't halt delivery, and lower severities stay visible to local audits.

## What I deliberately left out

#2813's `nanoid` tripwire test. The latest `3.x` is `3.3.18`, so postcss's `^3.3.16` resolves to the patched release unaided — the override is now a harmless floor rather than the fix. A test asserting the override equals `3.3.18` would fail the next legitimate patch bump to `3.3.19`. The audit gate covers the same regression without pinning brittleness.

Also left out: #2813's relay-input hardening and workflow-helper refactors, which are a separate concern from the audit gate.

## Verification

- `bun audit --audit-level high` on this branch: **No vulnerabilities found**
- `ci.yml` parses; `deploy-retro-relay` needs resolve to `[dogfood-parity, dependency-audit, test, lint, relay-inputs]`
- Workflow contract tests pass (3 files, 5 tests) — includes the updated `needs` assertion in `retro-relay-deploy-workflow.test.ts`
- `prettier --check` clean on both changed files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,25 @@ jobs:
       - name: Check public CLI contract
         run: bun run check:cli-contract
 
+  dependency-audit:
+    name: Dependency audit
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version-file: package.json
+
+      - name: Block high and critical advisories
+        # Lower-severity advisories stay visible to local and repository audits
+        # without letting an unpatchable development-only advisory halt delivery.
+        run: bun audit --audit-level high
+
   test:
     name: test (node ${{ matrix.node-version }})
     runs-on: ubuntu-latest
@@ -319,7 +338,10 @@ jobs:
 
   deploy-retro-relay:
     name: Deploy Retro Relay
-    needs: [dogfood-parity, test, lint, relay-inputs]
+    # High and critical advisories block automatic production delivery. The
+    # environment-protected deploy-retro-relay workflow stays the manual
+    # break-glass path for an urgent relay security fix.
+    needs: [dogfood-parity, dependency-audit, test, lint, relay-inputs]
     if: github.ref == 'refs/heads/main' && needs.relay-inputs.outputs.deploy == 'true'
     runs-on: ubuntu-latest
     environment: retro-relay-production
--- a/packages/cli/tests/retro-relay-deploy-workflow.test.ts
+++ b/packages/cli/tests/retro-relay-deploy-workflow.test.ts
@@ -53,7 +53,13 @@ describe('Retro Relay deployment workflow', () => {
     const deployment = workflow.jobs['deploy-retro-relay'];
     expect(deployment).toBeDefined();
     if (deployment === undefined) throw new Error('missing deploy-retro-relay job');
-    expect(deployment.needs).toEqual(['dogfood-parity', 'test', 'lint', 'relay-inputs']);
+    expect(deployment.needs).toEqual([
+      'dogfood-parity',
+      'dependency-audit',
+      'test',
+      'lint',
+      'relay-inputs',
+    ]);
     expect(deployment.environment).toBe('retro-relay-production');
     expect(deployment.if).toContain("github.ref == 'refs/heads/main'");
     expect(source).toContain('git diff --name-only "$BEFORE" "$SHA"');


.github/workflows/ci.yml
```diff
@@ -62,6 +62,25 @@ jobs:
...
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version-file: package.json
+
+      - name: Block high and critical advisories
+        # Lower-severity advisories stay visible to local and repository audits
+        # without letting an unpatchable development-only advisory halt delivery.
+        run: bun audit --audit-level high
```
TheMostlyGreat, line 82: Reviewer: Proven armed, not assumed. Against a synthetic tree pinning `nanoid@3.3.11` this exact command exits 1 and reports GHSA-2v37-7h3g-55p8 — the advisory from #2808.

Also intentional: there is no `bun install` above this step. `bun audit` resolves from `bun.lock` alone (verified with `node_modules` removed), which is what keeps the job at ~6s. Worth leaving a note here if anyone later "fixes" the missing install.

TheMostlyGreat (commented):
Reviewer: Self-review of my own PR, so treat it as a checklist rather than independent sign-off — and note `gh` authenticates as the author, so this carries no merge block.

I verified the gate is actually armed rather than assuming it. Two things worth your attention, one of them a genuine limitation.

**Verified working (empirically, not by inspection):**

- Built a synthetic tree pinning `nanoid@3.3.11` and ran the exact job command. `bun audit --audit-level high` exits **1** and names `GHSA-2v37-7h3g-55p8` — the precise advisory from #2808. The gate fails on the thing it was built to catch.
- The job has **no `bun install` step**, which I want to flag as intentional and proven, not an oversight: `bun audit` reads `bun.lock` directly. Re-ran the probe with `node_modules` deleted and it still exits 1. That is why the job lands in ~6s instead of ~60s.

**Limitation — the gate does not block merges.** `repos/ArcadeAI/safeword/rules/branches/main` lists only `Dogfood parity` and `CLI contract` as required status checks. `Dependency audit` is not among them, so a red audit shows up visibly but will not stop a merge. Its real teeth are on `deploy-retro-relay` via `needs`, which does hard-block automatic production delivery.

That may well be the posture you want, and I did not change it unilaterally — editing the ruleset is a repo-settings change and your call. Worth deciding explicitly rather than by default:

- Leave as-is: audit is advisory on PRs, blocking on deploy. A newly published advisory degrades to a visible warning instead of red-lining every open PR.
- Add to required checks: audit blocks merges too, at the cost that any new high-severity advisory in a transitive dep stops all merges until patched.

I lean toward leaving it, precisely because the failure mode of the stricter option is what makes teams delete gates.

**Accepted risks:** `bun audit` queries the registry, so an outage or rate-limit produces a red job. Bounded by the same reasoning — not merge-required, and the job is cheap to re-run.

**Note on scope:** I deliberately did not port #2813's `nanoid` tripwire. Latest 3.x is 3.3.18, so postcss's `^3.3.16` resolves patched unaided; a test asserting the exact override would fail the next legitimate bump to 3.3.19. Reasoning is in the PR body.

